PR #12, opened 2015-04-21
Remove ability to continue to login with servers identity of which can't be verified.

bsuvorov:
Also introduced new error 495 in similar fashion to NGINX error:
495 Cert Error (Nginx)
Nginx internal code used when SSL client certificate error occurred to
distinguish it from 4XX in a log and an error page redirection.

### DIFF
--- a/BoxContentSDK/BoxContentSDK/BOXContentSDKErrors.h
+++ b/BoxContentSDK/BoxContentSDK/BOXContentSDKErrors.h
@@ -35,6 +35,8 @@ typedef enum {
     BOXContentSDKAPIErrorRequestEntityTooLarge = 413,
     BOXContentSDKAPIErrorPreconditionRequired = 428,
     BOXContentSDKAPIErrorTooManyRequests = 429, // rate limit exceeded
+    BOXContentSDKAPIErrorServerCertError = 495, // indicates error when app encounters server with unverified identity.
+
     // 5xx errors
     BOXContentSDKAPIErrorInternalServerError = 500,
     BOXContentSDKAPIErrorInsufficientStorage = 507,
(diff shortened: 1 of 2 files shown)


BoxContentSDK/BoxContentSDK/BOXContentSDKErrors.h
```diff
@@ -35,6 +35,8 @@ typedef enum {
     BOXContentSDKAPIErrorRequestEntityTooLarge = 413,
     BOXContentSDKAPIErrorPreconditionRequired = 428,
     BOXContentSDKAPIErrorTooManyRequests = 429, // rate limit exceeded
+    BOXContentSDKAPIErrorServerCertError = 495, // indicates error when app encounters server with unverified identity.
```
boxtcarpel: Should we also handle '496 No Cert (Nginx)' in the same or similar way

BoxContentSDK/BoxContentSDK/OAuth2/BOXAuthorizationViewController.m
```diff
@@ -315,67 +316,23 @@ - (void)connection:(NSURLConnection *)connection willSendRequestForAuthenticatio
...
-		}
+            [self completeServerTrustAuthenticationChallenge:challenge shouldTrust:shouldTrustServer];
+        }
 		else
 		{
-			// By default, allow a certificate if its status was evaluated successfully and the result is
-			// that it should be trusted
-			BOOL shouldTrustServer = (status == errSecSuccess && (trustResult == kSecTrustResultProceed || trustResult == kSecTrustResultUnspecified));
-			[self completeServerTrustAuthenticationChallenge:challenge shouldTrust:shouldTrustServer];
+            UIAlertView *serverTrustAlertView = [[UIAlertView alloc] initWithTitle:NSLocalizedString(@"Cannot Verify Server Identity", @"Alert view title: title for SSO server trust challenge")
```
boxtcarpel: If we want to handle the failure case here directly, we should clean up the completeServerTrust method - it no longer needs a shouldTrust property, since we're only using it when the host is trusted, and we can take out the conditional in the implementation.

PR #17, opened 2015-04-22
Add random nonce and matching it with state returned with auth code response

bsuvorov:
Nonce check prevents from CSRF attack which otherwise would
technically let malicious user login in non-malicious user
into malicious user's spoofed account.

### DIFF
--- a/BoxContentSDK/BoxContentSDK/BOXContentSDKConstants.h
+++ b/BoxContentSDK/BoxContentSDK/BOXContentSDKConstants.h
@@ -61,6 +61,7 @@ extern BOXAPIHTTPHeader *const BOXAPIHTTPHeaderBoxAPI;
 
 // OAuth2 constants
 // Authorization code response
+extern NSString *const BOXOAuth2URLParameterAuthorizationStateKey;
 extern NSString *const BOXOAuth2URLParameterAuthorizationCodeKey;
 extern NSString *const BOXOAuth2URLParameterErrorCodeKey;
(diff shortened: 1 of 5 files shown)


BoxContentSDK/BoxContentSDK/BOXContentSDKConstants.h
```diff
@@ -61,6 +61,7 @@ extern BOXAPIHTTPHeader *const BOXAPIHTTPHeaderBoxAPI;
 
 // OAuth2 constants
 // Authorization code response
+extern NSString *const BOXOAuth2URLParameterAuthorizationStateKey;
```
boxtcarpel: this appears to be duplicated, we already have a state key below... (line 67)

BoxContentSDK/BoxContentSDK/OAuth2/BOXOAuth2Session.h
```diff
@@ -80,6 +80,11 @@ extern NSString *const BOXOAuth2UserIDKey;
  */
 @property (nonatomic, readwrite, weak) BOXAPIQueueManager *queueManager;
 
+/**
+ * The randomly generated nonce used to prevent spoofing attack during login
+ */
+@property (nonatomic, readwrite, strong) NSString *nonce;
```
boxtcarpel: Does this need to be public? I don't see it referred outside of the OAuth2Session class' implementation.

PR #19, opened 2015-04-22
When querying keychain, do not assume that entries are strings before checking for Box keychain entries.

RicoYao:
- If there are other entries that are not strings, then we were crashing
  when calling 'hasPrefix'.


salexkidd:
Awesome fix! :)

boxtcarpel:
+10 LGTM

:yum:

PR #27, opened 2015-05-26
Issue 26: When making a request for user object with 'enterprise' field, null value causes a crash.

RicoYao:
- We should be allowing null, and supressing it as nil.

https://github.com/box/box-ios-content-sdk/issues/26


boxcla:

RicoYao:
@boxtcarpel please review. Thanks.

boxtcarpel:
+2 LGTM

:floppy_disk:

PR #33, opened 2015-06-05
Extract common testing interface

DJBen:
Tested with

```
    pod 'box-ios-sdk-tests', :podspec => 'https://raw.githubusercontent.com/DJBen/box-ios-sdk/extracttests/box-ios-sdk-tests.podspec'
```

in the podfile. All good.

### DIFF
--- a/box-ios-sdk-tests.podspec
+++ b/box-ios-sdk-tests.podspec
@@ -1,0 +1,34 @@
+
+Pod::Spec.new do |s|
+
+  s.name         = "box-ios-sdk-tests"
+  s.version      = "1.0.5"
+  s.summary      = "A common testing interface extracted from Content SDK."
+  s.license      = { :type => "Apache 2.0", :file => "LICENSE" }
+  s.author       = "Box"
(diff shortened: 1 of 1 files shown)


boxcla:
Verified that @DJBen has signed the CLA. Thanks for the pull request!

box-ios-sdk-tests.podspec
```diff
@@ -0,0 +1,34 @@
+
+Pod::Spec.new do |s|
+
+  s.name         = "box-ios-sdk-tests"
+  s.version      = "1.0.0"
```
RicoYao: 1.0.5 to match the main podspec. We'll have to keep this updated I guess? Any way we can just make this point to the version in the main podspec?

bsuvorov:
LGTM, but please reply to Rico's comment.

box-ios-sdk-tests.podspec
DJBen: Yes. Although we can leave the :tag blank to ensure it always pull the head, it does not refer to the latest release. Also it generates a pod spec lint warning.

PR #36, opened 2015-06-12
Remove BOXFolder.items property.

RicoYao:
- Developers should be using BOXFolderItemsRequest to obtain children of a folder.
- BOXFolder.items only contains the first 100 items and there was no way for developers
  to request items beyond that in BOXFolderRequest. So it was a dangerous property to expose
  because developers might not realize they're not getting all the items in a folder.


boxcla:

crousselle:
LGTM. + 2

PR #52, opened 2015-07-15
Workaround podspec linting warnings for unused variables

jpang1:


boxcla:

PR #60, opened 2015-08-03
Fix the framework naming issue that prevents pod install in swift to …

DJBen:
…find the source.

Note: When merging, we need to add a tag with version 1.1.0.
Why not 1.0.9? Because after renaming, it's not compatible with 1.0.8 anymore, so we use semantic versioning to indicate that this is incompatible with the past version.

### DIFF
--- a/box-ios-sdk-interfaces.podspec
+++ b/box-ios-sdk-interfaces.podspec
@@ -4,7 +4,7 @@ Pod::Spec.new do |s|
 # Root specification
 
 s.name                  = "box-ios-sdk-interfaces"
-s.version               = "1.0.8"
+s.version               = "1.1.0"
 s.summary               = "iOS SDK for the Box API."
 s.homepage              = "https://github.com/box/box-ios-sdk"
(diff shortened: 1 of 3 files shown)


boxcla:
Verified that @DJBen has signed the CLA. Thanks for the pull request!

DJBen:
## Problem

In podfile, `use_frameworks!` will cause the current version `1.0.8` fail to be integrated because the dynamic framework it builds will be named `box_ios_sdk` instead of `BoxContentSDK`, causing all references within the framework to lose tracks.

## Solution

By adding a module name specification inside the pod spec, now the pods build the dynamic framework with the correct name.

## Workaround

This does not affect any pod references without specifying `use_frameworks!`. In another words, if the users' pods do not include any Swift sources then he/she will probably be fine. However if you are mixmatching this pod with some other Swift written pods, the current version will fail.

RicoYao:
Hey @DJBen should we do the same for the Browse and Share SDK?

boxtcarpel:
Continued on https://github.com/box/box-ios-sdk/pull/70

PR #63, opened 2015-08-17
Passing a generated nonce string in app to app authentication flow

sowmiya:
with box app

### DIFF
--- a/BoxContentSDK/BoxContentSDK/OAuth2/BOXOAuth2Session.h
+++ b/BoxContentSDK/BoxContentSDK/OAuth2/BOXOAuth2Session.h
@@ -156,6 +156,12 @@
  */
 - (NSString *)redirectURIString;
 
+/**
+ * Returns the randomly generated nonce used to prevent spoofing attack during login
+ * @return generated nonce
+ */
+- (NSString *)nonce;
+
 #pragma mark - Token Refresh
 /** @name Token Refresh */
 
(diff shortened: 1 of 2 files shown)


boxcla:
Hi @sowmiya, thanks for the pull request. Before we can merge it, we need you to sign our Contributor License Agreement. You can do so electronically here: http://opensource.box.com/cla

BoxContentSDK/BoxContentSDK/OAuth2/BOXOAuth2Session.h
```diff
@@ -156,6 +156,12 @@
  */
 - (NSString *)redirectURIString;
 
+/**
+ * Returns the randomly generated nonce used to prevent spoofing attack during login
+ * @return generated nonce
+ */
+- (NSString *)nonce;
```
boxtcarpel, line 163: Any reason that we should not expose this in the header? (in other words, in your thought process what were the pros and cons, or does this just make sense as-is?)

boxtcarpel:
+2 LGTM, Glad it was just the one step that was missing in the chain.

BoxContentSDK/BoxContentSDK/OAuth2/BOXOAuth2Session.h
sowmiya, line 163: As far as I can tell, we are not allowing users to modify this string by exposing it in the header. Also once generated, it remains the same so we wont run into any inconsistency in value.

PR #67, opened 2015-08-24
Ensure server nonce check succeeds only if it was previously

sowmiya:
generated by the client app


boxcla:

boxtcarpel:
+2 LGTM

:penguin:

PR #70, opened 2015-08-25
Fix the framework naming issue preventing pod install in swift to find the source

boxtcarpel:


boxcla:

boxtcarpel:
From https://github.com/box/box-ios-sdk/pull/60

boxtcarpel:
From Ben's comments:

## Problem

In podfile, `use_frameworks!` will cause the current version `1.0.8` fail to be integrated because the dynamic framework it builds will be named `box_ios_sdk` instead of `BoxContentSDK`, causing all references within the framework to lose tracks.

## Solution

By adding a module name specification inside the pod spec, now the pods build the dynamic framework with the correct name.

## Workaround

This does not affect any pod references without specifying `use_frameworks!`. In another words, if the users' pods do not include any Swift sources then he/she will probably be fine. However if you are mixmatching this pod with some other Swift written pods, the current version will fail.

PR #71, opened 2015-08-25
Add string method for detecting empty strings

boxtcarpel:
Where empty strings are strings that only contain whitespaces or new lines
or no characters at all.

### DIFF
--- a/BoxContentSDK/BoxContentSDK/Categories/NSString+BOXAdditions.m
+++ b/BoxContentSDK/BoxContentSDK/Categories/NSString+BOXAdditions.m
@@ -52,4 +52,12 @@ long long const BOX_TERABYTE = BOX_GIGABYTE * 1024;
     return result_str;
 }
 
+- (BOOL)box_isEmptyOrWhitespacesOnly
+{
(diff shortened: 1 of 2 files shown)


boxcla:

BoxContentSDK/BoxContentSDK/Categories/NSString+BOXAdditions.m
```diff
@@ -52,4 +52,12 @@ + (NSString *)box_humanReadableStringForByteSize:(NSNumber *)size
     return result_str;
 }
 
+- (BOOL)box_isEmptyOrWhitespacesOnly
+{
+    NSCharacterSet *nonWhitespaceCharacterSet = [[NSCharacterSet whitespaceAndNewlineCharacterSet] invertedSet];
+    BOOL isEmptyOrWhitespaces = ([self length] == 0 ||
+                                 [self rangeOfCharacterFromSet:nonWhitespaceCharacterSet].location != NSNotFound);
```
sowmiya: we should either not do the characterset inverting or check for == NSNotFound.

PR #73, opened 2015-08-26
Perform check for server nonce if the auth redirection was triggered

sowmiya:
by the client app

### DIFF
--- a/BoxContentSDK/BoxContentSDK/OAuth2/BOXOAuth2Session.m
+++ b/BoxContentSDK/BoxContentSDK/OAuth2/BOXOAuth2Session.m
@@ -69,14 +69,17 @@
     NSString *authorizationCode = [URLQueryParams valueForKey:BOXAuthURLParameterAuthorizationCodeKey];
     NSString *authorizationError = [URLQueryParams valueForKey:BOXAuthURLParameterErrorCodeKey];
 
-    if ((_nonce || serverNonce) && [serverNonce isEqualToString:_nonce] == NO) {
-        NSError *error = [[NSError alloc] initWithDomain:BOXContentSDKErrorDomain 
-                                                    code:BOXContentSDKAuthErrorAccessTokenNonceMismatch
-                                                userInfo:nil];
-        if (block) {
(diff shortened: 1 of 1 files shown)


boxcla:

BoxContentSDK/BoxContentSDK/OAuth2/BOXOAuth2Session.m
```diff
@@ -64,14 +64,17 @@ - (void)performAuthorizationCodeGrantWithReceivedURL:(NSURL *)URL withCompletion
     NSString *authorizationCode = [URLQueryParams valueForKey:BOXAuthURLParameterAuthorizationCodeKey];
     NSString *authorizationError = [URLQueryParams valueForKey:BOXAuthURLParameterErrorCodeKey];
 
-    if ((_nonce || serverNonce) && [serverNonce isEqualToString:_nonce] == NO) {
-        NSError *error = [[NSError alloc] initWithDomain:BOXContentSDKErrorDomain 
-                                                    code:BOXContentSDKAuthErrorAccessTokenNonceMismatch
-                                                userInfo:nil];
-        if (block) {
-            block(nil, error);
+    NSString *authenticationRedirectURIScheme = [[NSURL URLWithString:self.redirectURIString] scheme];
```
boxtcarpel, line 72: Ah, good catch.

So if this is the check to see if we SHOULD/NEED TO do a nonce-comparison, is it better (more inclusive) to check if the server response included the 'state' query parameter (not if there is a value, if the key is there)?
We can double check if that's the better check and if it is, either do it in this commit or a separate one.

boxtcarpel:
+2 LGTM

One comment

:green_book:

BoxContentSDK/BoxContentSDK/OAuth2/BOXOAuth2Session.m
sowmiya, line 72: From what I checked with the server team, we are not 100% confident that the server will always return a state param(when it is empty) so we cannot base it on that. We will go with the current fix.

PR #75, opened 2015-08-26
Set a strong reference to the NSOutputStream for BOXFileThumbnailRequest.

RicoYao:
We need to read from this outputstream in order to get the image when firing
the completion block.


boxcla:

boxtcarpel:
+2 LGTM

:part_alternation_mark:

PR #77, opened 2015-08-27
Support custom redirect URIs in the OAuth completion flow

boxtcarpel:
Check for the whole redirect URI instead of just the custom scheme,
since the scheme may be https:// for a custom defined redirect URI


boxcla:

RicoYao:
lgtm

PR #78, opened 2015-08-27
[WIP] Enable caching for request responses

h2kuo:


boxcla:

BoxContentSDK/BoxContentSDK/Requests/BOXFolderPaginatedItemsRequest.m
```diff
@@ -90,12 +116,21 @@ - (void)performRequestWithCompletion:(BOXItemArrayCompletionBlock)completionBloc
                                                                                       ancestors:pathFolders];
             }
             [BOXDispatchHelper callCompletionBlock:^{
-                completionBlock(items, totalCount, NSMakeRange(offset, limit), nil);
+                refreshBlock(items, totalCount, NSMakeRange(offset, limit), nil);
+                
+                // Don't want to update cache for a subclass recursively calling this method.
```
RicoYao, line 121: I think it would be good to update the cache in this case. It's a valid cache entry as long as the cache key contains the offset and limit somehow.

We just need to make sure these cache entries have zero impact on the caching of the subclass (BOXFolderItemsRequest) which should be doing its own caching.

BoxContentSDK/BoxContentSDK/Requests/BOXFolderPaginatedItemsRequest.m
```diff
@@ -90,12 +116,21 @@ - (void)performRequestWithCompletion:(BOXItemArrayCompletionBlock)completionBloc
...
+                    [self.requestCache updateCacheForRequest:self withResponse:JSONDictionary];
+                }
             } onMainThread:isMainThread];
         };
         folderOperation.failure = ^(NSURLRequest *request, NSHTTPURLResponse *response, NSError *error, NSDictionary *JSONDictionary) {
             [BOXDispatchHelper callCompletionBlock:^{
-                completionBlock(nil, 0, NSMakeRange(0, 0), error);
+                refreshBlock(nil, 0, NSMakeRange(0, 0), error);
+                // TODO: only if not network error
+                if ([self isMemberOfClass:[BOXFolderPaginatedItemsRequest class]]) {
```
RicoYao, line 131: Similar to previous comment, we might not need this memberOfClass check

BoxContentSDK/BoxContentSDK/Requests/BOXFolderRequest.h
```diff
@@ -18,5 +18,6 @@
 - (instancetype)initWithFolderID:(NSString *)folderID;
 - (instancetype)initWithFolderID:(NSString *)folderID isTrashed:(BOOL)isTrashed;
 - (void)performRequestWithCompletion:(BOXFolderBlock)completionBlock;
+- (void)performRequestWithCached:(BOXFolderBlock)cacheBlock refreshed:(BOXFolderBlock)refreshBlock;
```
RicoYao, line 21: @scottliubox what do you think about naming for this?

BoxContentSDK/BoxContentSDK/Cache/BOXRequestCache.m
```diff
@@ -0,0 +1,87 @@
...
+    if (self) {
+        NSFileManager *fileManager = [NSFileManager defaultManager];
+        
+        if (!cacheDirectory) {
+            cacheDirectory = [NSURL URLWithString:[NSSearchPathForDirectoriesInDomains(NSLibraryDirectory, NSUserDomainMask, YES) objectAtIndex:0]];
+        }
+        
+        _cachePath = [[cacheDirectory path] stringByAppendingPathComponent:[NSString stringWithFormat:@"%@/%@", CACHE_FOLDER, userID]];
+        
+        if (![fileManager fileExistsAtPath:_cachePath]) {
```
RicoYao: Use fileExistsAtPath:isDirectory: and ensure that it's a directory.

BoxContentSDK/BoxContentSDK/Cache/BOXRequestCache.m
```diff
@@ -0,0 +1,87 @@
...
+            [fileManager createDirectoryAtPath:_cachePath
+                   withIntermediateDirectories:YES
+                                    attributes:nil
+                                         error:nil];
```
RicoYao: check for error

BoxContentSDK/BoxContentSDK/Cache/BOXRequestCache.m
```diff
@@ -0,0 +1,87 @@
...
+        }
```
RicoYao, line 70: If we were able to establish a cache directory, log a warning and ensure that we don't try to read from/write to the cache in any other methods.

RicoYao:
- We need to wipe the cache if the user gets logged out.
- Can we ensure that the cache files cannot get syncup to iCloud or iTunes backup?
- We need to set an appropriate file protection level.

BoxContentSDK/BoxContentSDK/Cache/BOXRequestCache.h
```diff
@@ -0,0 +1,23 @@
...
+//
+
+#import <Foundation/Foundation.h>
+#import "PINCache.h"
+#import "BOXAPIJSONOperation.h"
+#import "BOXRequest.h"
+
+@interface BOXRequestCache : NSObject
+
+- (instancetype) initWithUserID:(NSString *)userID cacheDirectory:(NSURL *)cacheDirectory;
```
irvingruan: Minor: spaces.

BoxContentSDK/BoxContentSDK/Cache/BOXRequestCache.h
```diff
@@ -0,0 +1,23 @@
+//
+//  BOXRequestCache.h
+//  BoxContentSDK
+//
+//  Created on 8/25/15.
+//  Copyright (c) 2015 Box. All rights reserved.
+//
+
+#import <Foundation/Foundation.h>
+#import "PINCache.h"
```
irvingruan: Any reason why we need to add PINCache.h import in the header and not implementation?

BoxContentSDK/BoxContentSDK/Cache/BOXRequestCache.m
```diff
@@ -0,0 +1,87 @@
...
+#import "BOXContentClient.h"
+#import "BOXUser.h"
+#import "BOXRequest_Private.h"
+#import "BOXFolderItemsRequest.h"
+
+#define CACHE_FOLDER @"boxRequestCache"
+
+@interface BOXRequestCache ()
+
+@property (nonatomic, readwrite, strong) NSString *cachePath;
```
irvingruan: It's safer to use copy over strong for NSStrings: https://stackoverflow.com/questions/11249656/should-an-nsstring-property-under-arc-be-strong-or-copy

BoxContentSDK/BoxContentSDK/Cache/BOXRequestCache.m
```diff
@@ -0,0 +1,87 @@
...
+@end
+
+@implementation BOXRequestCache
+
+- (instancetype) initWithUserID:(NSString *)userID
+{
+    return [self initWithUserID:userID cacheDirectory:nil];
+}
+
+- (instancetype) initWithUserID:(NSString *)userID cacheDirectory:(NSURL *)cacheDirectory
```
irvingruan: Space between ) and init.

BoxContentSDK/BoxContentSDK/Cache/BOXRequestCache.m
```diff
@@ -0,0 +1,87 @@
...
+        _cache = [[PINCache alloc] initWithName:@"BoxRequestCache" rootPath:_cachePath];
+    }
+    
+    return self;
+}
+
+- (void)fetchCacheResponseForRequest:(BOXRequest *)request cacheBlock:(void(^)(NSDictionary *dictionary))cacheBlock
+{
+    PINCacheObjectBlock localCacheBlock = ^void(PINCache *cache, NSString *key, id __nullable object) {
+        if ([object isKindOfClass:[NSDictionary class]]) {
```
irvingruan: Need to check if cacheBlock != nil.

BoxContentSDK/BoxContentSDK/Cache/BOXRequestCache.m
```diff
@@ -0,0 +1,87 @@
...
+            cacheBlock(object);
+        }
+    };
+    [self.cache objectForKey:[self keyForRequest:request] block:localCacheBlock];
+}
+
+- (void)removeCacheResponseForRequest:(BOXRequest *)request
+{
+    [self.cache removeObjectForKey:[self keyForRequest:request]];
```
irvingruan, line 125: Is `removeObjectForKey` nil safe? If not, we should be careful about what `keyForRequest` returns.

BoxContentSDK/BoxContentSDK/Cache/BOXRequestCache.m
```diff
@@ -0,0 +1,87 @@
...
+}
+
+- (void)updateCacheForRequest:(BOXRequest *)request withResponse:(NSDictionary *)JSONDictionary
+{
+    [self.cache setObject:JSONDictionary forKey:[self keyForRequest:request] block:nil];
```
irvingruan, line 130: Should we check if `keyForRequest` is non-nil? if PINCache's `setObject:forKey:` is nil safe then we should be OK.

BoxContentSDK/BoxContentSDK/Clients/BOXContentClient.h
```diff
@@ -46,6 +47,12 @@
 @property (nonatomic, readonly, assign) BOOL appToAppBoxAuthenticationEnabled;
 
 /**
+ * This property reflects whether the client should cache API request responses. 
+ * Its default is NO.
+ */
+@property (nonatomic, readwrite, assign) BOOL requestCachingEnabled;
```
irvingruan: Suggestion: `@property (nonatomic, readwrite, assign, getter=isRequestCachingEnabled) BOOL requestCachingEnabled`. Apple has never been the best at auto-generating the proper grammar for bool checks.

BoxContentSDK/BoxContentSDK/Clients/BOXContentClient.m
```diff
@@ -126,6 +129,22 @@ + (void)setAppToAppBoxAuthenticationEnabled:(BOOL)enabled
     staticAppToAppBoxAuthenticationEnabled = enabled;
 }
 
+- (void)setRequestCachingEnabled:(BOOL)enabled
+{
+    [self setRequestCachingEnabled:enabled cacheDirectory:nil];
```
irvingruan, line 148: We might be able to save some electronics if we check if `_enabled != enabled`.

BoxContentSDK/BoxContentSDK/Requests/BOXFolderItemsRequest.m
```diff
@@ -80,58 +83,79 @@ + (NSArray *)dedupeItemsByBoxID:(NSArray *)items
...
-    // used to work around warning on retain loop on paginatedItemFetch
-    __block BOXItemArrayCompletionBlock recursiveFetch = nil;
-    
-    BOXItemsBlock localCompletionBlock = ^(NSArray *items, NSError *error){
-        [BOXDispatchHelper callCompletionBlock:^{
-            completionBlock(items, error);
-        } onMainThread:isMainThread];
-    };
+    if (cacheBlock) {
+        if (self.requestCache) {
```
irvingruan, line 94: If requestCache is nil, should we just surface up a nil object in cacheBlock? Otherwise, the caller might expect the completion block to be called.

BoxContentSDK/BoxContentSDK/Requests/BOXFolderPaginatedItemsRequest.m
```diff
@@ -60,10 +60,36 @@ - (BOXAPIOperation *)createOperation
...
+
+- (void)performRequestWithCached:(BOXItemArrayCompletionBlock)cacheBlock refreshed:(BOXItemArrayCompletionBlock)refreshBlock
+{
     BOOL isMainThread = [NSThread isMainThread];
     BOXAPIJSONOperation *folderOperation = (BOXAPIJSONOperation *)self.operation;
-
-    if (completionBlock) {
+    
+    if (cacheBlock) {
+        if (self.requestCache) {
```
irvingruan, line 72: Curious: can cacheBlock and refreshBlock exist independently of each other? e.g. refreshBlock is non-nil, cacheBlock is nil.

BoxContentSDK/BoxContentSDK/Cache/BOXRequestCache.m
h2kuo, line 125: `removeObjectForKey` is nil-safe, just returns if !key

BoxContentSDK/BoxContentSDK/Cache/BOXRequestCache.m
h2kuo, line 130: `setObject:forKey:` is nil-safe for both object and key.

BoxContentSDK/BoxContentSDK/Requests/BOXFolderItemsRequest.m
h2kuo, line 94: The idea is that if the requestCache is nil, they don't have caching enabled, so I think it's fine to not call it. We could add something in the future to log a warning if they try to pass in a cacheBlock but don't have caching enabled.

BoxContentSDK/BoxContentSDK/Requests/BOXFolderPaginatedItemsRequest.m
h2kuo, line 72: Yes, so if they want to use exclusively refreshed or cached data, they have the option to. `performRequestWithCompletion:` actually just calls this method with a nil cacheBlock.

BoxContentSDK/BoxContentSDK/Requests/BOXFolderPaginatedItemsRequest.m
h2kuo, line 121: I agree that it would be good to update the cache here, but I haven't figured out a good way to differentiate it, since it still gets passed as the subclass to the cache update method. Would be good to add in the future.

BoxContentSDK/BoxContentSDK/Requests/BOXFolderItemsRequest.m
irvingruan, line 94: Got it. Thanks for expanding on that. In that case, I think it's safe to optimize for `(cacheBlock && self.requestCache)`.

BoxContentSDK/BoxContentSDK/Requests/BOXFolderPaginatedItemsRequest.m
irvingruan, line 72: Got it.

PR #81, opened 2015-08-31
Completion block is not getting called on terminal connection errors

boxtcarpel:
The user shouldn't have to dismiss the error alert and then tap cancel.


boxcla:

RicoYao:
LGTM

PR #59, opened 2015-08-03
ISSUE-55 - Memory leaks from requests causing completionBlock to be r…

achunBOX:
…etained indefinitely

PR #86, opened 2015-09-05
Improve handling of expired access tokens, especially for upload operations.

RicoYao:
1. Upload operations (BOXAPIMultipartToJSONOperation) cannot be re-enqueued. Currently, we
   are re-enqueing them, which results in a (previously) mysterious 405 response. Added an explicit
   (BOOL)canBeReenqueued method so that it's clearer to see when it's possible to re-enqueue.
2. For BOXFileUploadRequest and BOXFileUploadNewVersionRequest, added a pre-emptive check
   on the access token expiry, and if it is close to expiration, we now manually attempt a token
   refresh. This minimizes the chance that a developer will actually get the error due to our
   inability to re-enqueue (BOXContentSDKAuthErrorAccessTokenExpiredOperationCannotBeReenqueued).
3. Un-did some of the changes made in BOXOAuth2Session in https://github.com/box/box-ios-sdk/pull/59/files
   wherein we de-referenced 'self'. This (sometimes) led to us not storing new tokens from the server,
   after doing a refresh. In this case, we really do need to hang on to 'self' until the refresh request completes.

### DIFF
--- a/BoxContentSDK/BoxContentSDK/OAuth2/BOXOAuth2Session.m
+++ b/BoxContentSDK/BoxContentSDK/OAuth2/BOXOAuth2Session.m
@@ -210,7 +210,6 @@
                                  BOXAuthTokenRequestClientSecretKey : self.clientSecret,
                                  };
     
-    __weak BOXOAuth2Session *weakSelf = self;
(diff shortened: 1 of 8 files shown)


boxcla:

crousselle:
lgtm :) +1

RicoYao:
@boxtcarpel please review post-merge

PR #87, opened 2015-09-07
Make a call to the 'revoke' endpoint on logout.

RicoYao:
- We cannot guarantee that the call will succeed (e.g. network connectivity) but we still
  make our best attempt at revoking the tokens.

### DIFF
--- a/BoxContentSDK/BoxContentSDK/OAuth2/BOXOAuth2Session.m
+++ b/BoxContentSDK/BoxContentSDK/OAuth2/BOXOAuth2Session.m
@@ -251,6 +251,36 @@
     [self.queueManager enqueueOperation:operation];
 }
 
+#pragma mark - Token revoke
+
+- (void)revokeCredentials
+{
+    // Call this before '[super revokeCredentials]' otherwise we would have already lost our accessToken.
+    [self sendRevokeRequest];
+    
+    [super revokeCredentials];
+}
+
+/**
+ *  Send API request to revoke tokens. We don't care about the response to this request.
+ *  This will be called when we're about to wipe the token from keychain+memory, so we could not do anything
+ *  about successes/failures anyway.
+ */
+- (void)sendRevokeRequest
+{
+    // We don't go through any of our regular queues/operations because those get shut down upon logout.
+    NSMutableURLRequest *request = [NSMutableURLRequest requestWithURL:[NSURL URLWithString:[NSString stringWithFormat:@"%@/oauth2/revoke", BOXAPIBaseURL]]];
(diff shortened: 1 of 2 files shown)


boxcla:

h2kuo:
+2

BoxContentSDK/BoxContentSDK/OAuth2/BOXOAuth2Session.m
```diff
@@ -251,6 +251,36 @@ - (void)performRefreshTokenGrant:(NSString *)expiredAccessToken withCompletionBl
...
+    // We don't go through any of our regular queues/operations because those get shut down upon logout.
+    NSMutableURLRequest *request = [NSMutableURLRequest requestWithURL:[NSURL URLWithString:[NSString stringWithFormat:@"%@/oauth2/revoke", BOXAPIBaseURL]]];
+    [request setHTTPMethod:@"POST"];
+    [request setValue:@"application/x-www-form-urlencoded charset=utf-8" forHTTPHeaderField:@"Content-Type"];
+    
+    NSData *postData = [[NSString stringWithFormat:@"client_id=%@&client_secret=%@&token=%@", self.clientID, self.clientSecret, self.accessToken] dataUsingEncoding:NSUTF8StringEncoding];
+    [request setValue:[NSString stringWithFormat:@"%lu", (unsigned long)postData.length] forHTTPHeaderField:@"Content-Length"];
+    [request setHTTPBody:postData];
+    
+    NSURLConnection *connection = [[NSURLConnection alloc] initWithRequest:request delegate:nil];
```
boxtcarpel, line 280: We can't leverage any of our operations for this? Like we use the json operation for /token ?

Basically, there is nothing else in this level that makes a direct network request

boxtcarpel:
Nice, great catch, I didn't realize we didn't have this before.

BoxContentSDK/BoxContentSDK/OAuth2/BOXOAuth2Session.m
RicoYao, line 280: We could do something if we wanted to set up a new queue manager and ensure that it doesn't get killed by other logout routines. Seemed unnecessarily complicated though. What do you think?

boxtcarpel, line 280: I think it's OK, just a bit of a standout, but I guess it's a standout thing and we don't need to be too aggressive on separating the layers. We can always re-evaluate if/when we need to.

Could this be a good candidate for the first usage of a URL session, or is that crazy talk?

PR #88, opened 2015-09-08
Add check for errors which should result in cached response removal.

h2kuo:


boxcla:

BoxContentSDK/BoxContentSDK/Cache/BOXRequestCache.m
```diff
@@ -135,4 +136,15 @@ - (void)clearForLogout
     }
 }
 
+- (BOOL)shouldRemoveCachedResponseForRequestError:(NSError *)error
```
RicoYao: Perhaps this belongs in BOXRequest so that it can be extended for different types of requests?

BoxContentSDK/BoxContentSDK/Cache/BOXRequestCache.m
```diff
@@ -135,4 +136,15 @@ - (void)clearForLogout
...
+{
+    if (error.code == BOXContentSDKAPIErrorUnauthorized ||
```
RicoYao: - Check for error domain.
- Safety check for nil error

BoxContentSDK/BoxContentSDK/Cache/BOXRequestCache.m
h2kuo: As discussed offline, moved to static method in BOXRequest.

BoxContentSDK/BoxContentSDK/Cache/BOXRequestCache.m
h2kuo: Done.

RicoYao:
lgtm

PR #91, opened 2015-09-09
Fix BOXAPIMultipartToJSONOperation not closing its output stream on t…

jpang1:
…he same runloop when operation is canceled.

### DIFF
--- a/BoxContentSDK/BoxContentSDK/Operations/BOXAPIMultipartToJSONOperation.m
+++ b/BoxContentSDK/BoxContentSDK/Operations/BOXAPIMultipartToJSONOperation.m
@@ -485,14 +485,6 @@ static NSString * BOXAPIMultipartContentTypeHeader(void)
     }
 }
 
-- (void)cancel
-{
-    // Close the output stream before cancelling the operation
-    [self close];
-
-    [super cancel];
-}
-
 #pragma mark - Multipart Stream methods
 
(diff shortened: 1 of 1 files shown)


boxcla:

RicoYao:
lgtm. this makes much more sense than the old way.

BoxContentSDK/BoxContentSDK/Operations/BOXAPIMultipartToJSONOperation.m
```diff
@@ -590,6 +587,11 @@ - (void)writeDataToOutputStream
 {
     while ([self.outputStream hasSpaceAvailable])
     {
+        if (self.isCancelled) {
```
boxtcarpel, line 590: Anywhere else to handle this or are these 2 the only stream methods?

How do we make sure this is handled correctly if/when another method is added?

boxtcarpel:
+1

BoxContentSDK/BoxContentSDK/Operations/BOXAPIMultipartToJSONOperation.m
RicoYao, line 590: These changes are both in one method. I believe there's only one stream delegate method for handling events in the runloop (handleEvent) so I don't think we need to worry about some other method.

PR #92, opened 2015-09-09
Do cache retrievals synchronously so we can ensure that cache blocks fire before refresh blocks.

RicoYao:
…

2 reasons to do this:
1. Developers might reasonably expect that cache blocks should always fire before refresh blocks.
2. We are vulnerable to inconsistencies within the SDK itself if we end up writing to the cache
in the refresh block before we've even read from the cache block.


boxcla:

h2kuo:
+2

PR #95, opened 2015-09-10
Converting methods used to set keychain prefix & access group to be

sowmiya:
class methods


boxcla:

sowmiya:
Since there is more than one way to create a contentClient, it is tedious to set up the keychain prefix and access groups for each instance. Also, the prefix & access group strings are static already.

boxtcarpel:
+2 LGTM

:golf:

RicoYao:
Good change. It's a breaking change, but hopefully no-one was using those methods yet. Let's go for it.

PR #98, opened 2015-09-11
Converting methods used to set keychain prefix & access group to be

sowmiya:
class methods


boxcla:

sowmiya:
cherrypicking commit to master branch.

PR #99, opened 2015-09-12
Get rid of extraneous use of canOpenURL: since it is more locked down in iOS 9

boxtcarpel:
We still have some uses of it, but they are more targeted, which is supported in
iOS 9 with the registration of schemes in the plist under LSApplicationQueriesSchemes.


boxcla:

RicoYao:
lgtm

PR #102, opened 2015-09-12
Perform check for server nonce if the auth redirection was triggered

sowmiya:
by the client app


boxcla:

sowmiya:
cherrypicking change from master.

PR #103, opened 2015-09-13
Do not try to detect expired access tokens ahead-of-time in BOXAPIAuthenticatedRequest.

RicoYao:
- It's not a good idea to do it here because there's no way to NOT re-enqueue if the
  token refresh request fails.
- Let the expired token detection at the end of the request take care of re-enqueuing.
  It's only very slightly less optimal.


boxcla:

boxtcarpel:
+2 LGTM

:hatched_chick:

PR #106, opened 2015-09-14
We were vulnerable to the following scenario when refreshing expired tokens:

RicoYao:
1. Be in airplane mode with an expired token. The first refresh request will fail (expected), but
   will also block all future attempts to refresh the same token.
2. Turn off airplane mode. We won't attempt to refresh the token because we've already blacklisted
   it.

We've unconvered some technical debt:
- 'performRefreshTokenGrant' takes in a completion block but cannot actually handle calling it with
  meaningful data, in a parallel session. We whould either remove the completion block as a parameter
  to that method, or we need to implement the observer-notifier pattern.

### DIFF
--- a/BoxContentSDK/BoxContentSDK/OAuth2/BOXParallelOAuth2Session.m
+++ b/BoxContentSDK/BoxContentSDK/OAuth2/BOXParallelOAuth2Session.m
@@ -10,6 +10,7 @@
 #import "BOXAPIOAuth2ToJSONOperation.h"
 #import "BOXLog.h"
 #import "BOXContentSDKConstants.h"
+#import "BOXContentSDKErrors.h"
 
 @interface BOXParallelOAuth2Session ()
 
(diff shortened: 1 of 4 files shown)


boxcla:

BoxContentSDK/BoxContentSDK/OAuth2/BOXParallelOAuth2Session.m
```diff
@@ -44,19 +45,44 @@ - (void)performRefreshTokenGrant:(NSString *)expiredAccessToken withCompletionBl
...
+                [self.expiredOAuth2Tokens removeObject:expiredAccessToken];
+            }
+        }];
     }
 }
 
+- (BOOL)isInvalidGrantError:(NSError *)error
+{
+    BOOL isInvalidGrantError = NO;
+    if ([error.domain isEqualToString:BOXContentSDKErrorDomain]) {
```
boxtcarpel, line 78: We don't have this already somewhere we can reuse? Can any other code piggy-back on this now?

boxtcarpel:
+2 LGTM

:dolphin:

BoxContentSDK/BoxContentSDK/OAuth2/BOXParallelOAuth2Session.m
RicoYao, line 78: There's a slightly different flavor of this called 'shouldErrorTriggerLogout' but that also checks for other types of error codes, which I don't think are appropriate for this. I think that's the one that's more likely to be re-used.

PR #109, opened 2015-09-15
BOXParallelOAuth2Session - nil safety check before removing from NSMutableSet self.expiredOAuth2Tokens.

RicoYao:


boxcla:

boxtcarpel:
+2 LGTM

So we don't want to skip the call to super at this point to minimize the impact of the change. Sounds good.

:interrobang:

RicoYao:
@boxtcarpel Correct, I think we need a more general refactoring here, but I'm not comfortable doing that right now. There could some other component that is relying on the existing behavior (failed oAuth2 POST request). If I intercept that, something else could break.

PR #118, opened 2015-10-15
Update auth view controller

boxtcarpel:
Expose some advanced configuration options

### DIFF
--- a/BoxContentSDK/BoxContentSDK/OAuth2/BOXAuthorizationViewController.h
+++ b/BoxContentSDK/BoxContentSDK/OAuth2/BOXAuthorizationViewController.h
@@ -7,8 +7,9 @@
 //
 
 #import <UIKit/UIKit.h>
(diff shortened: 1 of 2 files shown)


boxcla:

BoxContentSDK/BoxContentSDK/OAuth2/BOXAuthorizationViewController.h
```diff
@@ -30,14 +31,36 @@
  */
 @interface BOXAuthorizationViewController : UIViewController
 
+/**
+ * The authroziation URL to be used.
```
sowmiya, line 35: spelling.

sowmiya:
I dont fully understand why we need to expose a setter for authorizationURL, would love to catch up on that. Other than that it is a +2.

boxtcarpel:
It's mainly for the ability to have fall-back auth URLs, we may remove it.

PR #120, opened 2015-10-15
Remove an extra public property from the auth view controller interface

boxtcarpel:
Made the authorizeURL property private only.


boxcla:

sowmiya:
+2

PR #154, opened 2015-12-22
When querying the keychain for 'usersInKeychain', set kSecAttrService.

RicoYao:
This gets set 'automatically' by BOXKeychainItemWrapper, but we were not setting in when querying
the keychain because we don't use BOXKeychainItemWrapper for querying (we only use it for adding values).


boxcla:

boxtcarpel:
+2 We discussed offline about some refactoring that will make this abstraction cleaner and will make it harder to omit/forget parts of the configuration when using the keychain in the future.

:recycle:

PR #175, opened 2016-02-19
update .podspec, remove -ObjC and -all_load flags

aletan:
Fix conflict with other pods (example libjingle_peerconnection)

### DIFF
--- a/BoxContentSDKSampleApp/Podfile.lock
+++ b/BoxContentSDKSampleApp/Podfile.lock
@@ -14,6 +14,6 @@ EXTERNAL SOURCES:
     :path: ../
 
 SPEC CHECKSUMS:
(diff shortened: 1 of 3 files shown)


boxcla:
Hi @aletan, thanks for the pull request. Before we can merge it, we need you to sign our Contributor License Agreement. You can do so electronically here: http://opensource.box.com/cla

aletan:
CLA signed

boxcla:
Verified that @aletan has just signed the CLA. Thanks, and we look forward to your contribution.

RicoYao:
@aletan Thanks for submitting this.

I believe the reason we added all_load was to have the categories compiled in. Without it, one might run into runtime errors (crashes). We should test that category methods still work with this, and also look into whether we need to use force_load instead of all_load.

boxtcarpel:
We have moved to using Carthage. Thank you for your pull request and for using the SDK.

PR #180, opened 2016-02-25
IOS-10522 device pinning, per code review renaming property to additi…

jdizoglio:
…onalMessageParameters


boxcla:

jdizoglio:
@boxtcarpel I didn't include the property name change in my earlier CR - additionalMessageParameters

jdizoglio:
An already approved simple change, +2 to unblock app related app build issue.

PR #181, opened 2016-02-25
IOS-10592 Auth VC - Do not attempt to load UIWebView request until viewDidAppear.

RicoYao:
- If we attempt to load the request to soon we risk the VC not being in the view
  heirarchy yet, and so it makes it difficult to load alerts, respond to errors, etc.
- Use a UIAlertController for showing error alert instead of UIAlertView. Fire the
  completion handler with the error so that developers can respond.
- Update deployment target to iOS 8.0 (won't apply until we actually push to CocoaPods).


boxcla:

PR #183, opened 2016-03-01
Add constants for API error code recent_similar_comment

h2kuo:


boxcla:

BoxContentSDK/BoxContentSDK/BOXContentSDKErrors.h
```diff
@@ -14,6 +14,9 @@ extern NSString *const BOXContentSDKErrorDomain;
 // A key for the userInfo dictionary to look up the underlying JSON error response from the API
 extern NSString *const BOXJSONErrorResponseKey;
 
+// A key for the error response dictionary to look up the underlying error code from the API
+extern NSString *const BOXErrorInfoCodeKey;
```
boxtcarpel, line 18: We don't have this already? I thought we do this in other places, perhaps auth for example

BoxContentSDK/BoxContentSDK/BOXContentSDKErrors.h
```diff
@@ -84,3 +87,5 @@ extern NSString *const BOXAuthErrorMissingDeviceId;
 extern NSString *const BOXAuthErrorUnsupportedDevicePinningRuntime;
 
 extern NSString *const BOXAuthErrorAccessDenied; // User denied access to their Box account
+
+extern NSString *const BOXRequestErrorRecentSimilarComment;
```
boxtcarpel, line 91: Is this unique to comments or can the error be more generic like identicalRequest or duplicateRequest or something so that it could later apply to other things?

Or even, what is the error for adding a collaborator who is already collaborated? Is that appropriate for this type of situation? Adding a comment that was already added...

h2kuo:
Deciding to keep specific error codes out of the content SDK.

PR #184, opened 2016-03-04
Fix shared link storage for folders within a folder

h2kuo:
Changed class check for folders to BOXFolder instead of BOXFile.


boxcla:

boxtcarpel:
+2 LGTM

:walking:

h2kuo:
I realized this could be done in a simpler way - revised commit and tested all item types to make sure it still worked.

PR #186, opened 2016-03-05
Improve the handling of the auth web view loading indicator

boxtcarpel:
Do not display the loading indicator when no web content actually
has to load. This avoid a situation where the loading indicator
remains on screen until the next action / page load.


boxcla:

h2kuo:
+2

PR #188, opened 2016-03-10
Prevent race condition in NSDate+BOXAdditions where the date formatte…

jpang1:
…r could be modified across multiple threads causing incorrect values to be returned.


boxcla:

boxtcarpel:
We should @synchronize just the critical code path, as we discussed offline.

With some additional testing on the new change, LGTM +2

:family:

PR #190, opened 2016-03-11
IOS-10592 : Fix connection error handling.

crousselle:


boxcla:

BoxContentSDK/BoxContentSDK/OAuth2/BOXAuthorizationViewController.m
```diff
@@ -241,38 +250,49 @@ - (void)failAuthenticationWithConnection:(NSURLConnection *)connection
...
-        UIAlertAction *okAction = [UIAlertAction actionWithTitle:NSLocalizedString(@"OK", @"Label: Allow the user to accept the current condition, often used on buttons to dismiss alerts")
-                                                           style:UIAlertActionStyleDefault
-                                                         handler:completion];
-        
-        [alertController addAction:okAction];
-        [self presentViewController:alertController animated:YES completion:nil];
+        [self handleConnectionError];
+    } else {
+        // We still want to report the completion block so that the necessary cleanup is done up in the call tree.
+        if (self.completionBlock) {
```
crousselle, line 262: Part of the issue was that the completion block was not called when the login controller was loaded in the 'background' before the user taps the login button. That lead it to not be cleared in memory, and the same controller was used when the user actually touched the login button. Calling the completion block clears the required components and provides a correct state for when the user touches the login button.

crousselle:
@RicoYao FYI

BoxContentSDK/BoxContentSDK/OAuth2/BOXAuthorizationViewController.m
```diff
@@ -241,38 +250,52 @@ - (void)failAuthenticationWithConnection:(NSURLConnection *)connection
...
+{
+    UIAlertController *alertController = [UIAlertController alertControllerWithTitle:NSLocalizedString(@"Unable to Log In", @"Alert view title: Title for failed SSO login due to authentication issue")
+                                                                             message:self.connectionErrorMessage
+                                                                      preferredStyle:UIAlertControllerStyleAlert];
+    
+    __weak typeof(self) weakSelf = self;
+    void (^completion)(UIAlertAction *action) = ^void(UIAlertAction *action) {
+        [alertController dismissViewControllerAnimated:YES completion:nil];
+        
+        if (weakSelf.completionBlock) {
```
boxtcarpel: We could get multiple completion block calls for the same error this way...

BoxContentSDK/BoxContentSDK/OAuth2/BOXAuthorizationViewController.m
```diff
@@ -241,38 +250,52 @@ - (void)failAuthenticationWithConnection:(NSURLConnection *)connection
...
-        [alertController addAction:okAction];
-        [self presentViewController:alertController animated:YES completion:nil];
+        [self handleConnectionError];
+        // Nil out to avoid any view did appear extra call.
+        self.connectionError = nil;
+        self.connectionErrorMessage = nil;
+    } else {
+        // We still want to report the completion block so that the necessary cleanup is done up in the call tree.
+        if (self.completionBlock) {
+            self.completionBlock(self, nil, self.connectionError);
```
boxtcarpel, line 263: We should either handle the error here, or save the error information and then handle it on appearance. That will ensure it only gets handled once.

Advantage of doing this: faster response to error.
Disadvantage: less built-in error handling in the SDK, more reliance on developer's app layer code, behavior is different based on conditions.

crousselle, line 263: I don't think that would work in all cases, what happens if an error (that is not necessarily connection-related) occurs after appearance. The fact that we get this called before appearance is an edge case when in airplane mode or have no connection at all.

BoxContentSDK/BoxContentSDK/OAuth2/BOXAuthorizationViewController.m
crousselle: True. the complex thing here is that we're trying to handle Flexo's specific use of this controller (loading it without presenting it). I'll fix it.

BoxContentSDK/BoxContentSDK/OAuth2/BOXAuthorizationViewController.m
crousselle, line 263: After thinking about it, we need it here. We have several cases we need to handle, and we can't only handle the error in one place without missing one of them : 
- Error when the controller is not presented (as done in flexo, viewDidAppear will never get called, need to call the completion block)
- Error when presented but viewDidAppear was not called yet : we handle the error and then make sure it won't be handled by viewDidAppear
- Error after viewDidAppear is called : handle the error.

BoxContentSDK/BoxContentSDK/OAuth2/BOXAuthorizationViewController.m
crousselle: it would be the case if and only if (a very edge case imo) : the error comes when the view's window is nil, and then somehow the viewDidAppear method is called.

BoxContentSDK/BoxContentSDK/OAuth2/BOXAuthorizationViewController.m
```diff
@@ -241,38 +250,52 @@ - (void)failAuthenticationWithConnection:(NSURLConnection *)connection
...
+    
     [self.activityIndicator stopAnimating];
 }
 
+- (void)handleConnectionError
+{
+    UIAlertController *alertController = [UIAlertController alertControllerWithTitle:NSLocalizedString(@"Unable to Log In", @"Alert view title: Title for failed SSO login due to authentication issue")
+                                                                             message:self.connectionErrorMessage
+                                                                      preferredStyle:UIAlertControllerStyleAlert];
+
```
boxtcarpel, line 275: nil out connection message

BoxContentSDK/BoxContentSDK/OAuth2/BOXAuthorizationViewController.m
```diff
@@ -241,38 +250,52 @@ - (void)failAuthenticationWithConnection:(NSURLConnection *)connection
...
-        };
-        
-        UIAlertAction *okAction = [UIAlertAction actionWithTitle:NSLocalizedString(@"OK", @"Label: Allow the user to accept the current condition, often used on buttons to dismiss alerts")
-                                                           style:UIAlertActionStyleDefault
-                                                         handler:completion];
-        
-        [alertController addAction:okAction];
-        [self presentViewController:alertController animated:YES completion:nil];
+        [self handleConnectionError];
+        // Nil out to avoid any view did appear extra call.
```
boxtcarpel: don't need this here, we nil them out in handleConnectionError

boxtcarpel:
+2 LGTM

:leaves:

PR #191, opened 2016-03-16
IOS-11071 : Improved test coverage on device pinning errors.

crousselle:


boxcla:

BoxContentSDK/BoxContentSDK/BOXContentSDKErrors.m
```diff
@@ -16,7 +16,7 @@
 NSString *const BOXAuthTokenRequestErrorInvalidRequest = @"invalid_request";
 
 NSString *const BOXAuthErrorUnauthorizedDevice = @"unauthorized_device";
-NSString *const BOXAuthErrorExceededDeviceLimit = @"exceeded_device_limit";
+NSString *const BOXAuthErrorExceededDeviceLimit = @"exceeded_device_limits";
```
thuyBox, line 19: since we've been checking against exceeded_device_limit (without 's') in Bender, I wonder if it's a typo in Bender or in the API document. Could we confirm with @ryao, @boxtcarpel if they might know the context?

thuyBox:
Could we confirm if we should be checking for unauthorized device (400, 403), exceeded device limit (400, 403), unsupported device pinning runtime error (400, 403), unauthorized error (401) as in Bender, or keeping the current error handling to only 400 for all unauthorized device, exceeded device limit, unsupported device pinning runtime and unauthorized error? @crousselle @boxtcarpel @RicoYao

crousselle:
@thuyBox we should probably follow the documentation regarding the error code. The error code seems to have changed at some point which is why we ended up testing for both in Bender. 
After several tests of different use cases, I systematically got a 400 error and never a 403.

BoxContentSDK/BoxContentSDK/BOXContentSDKErrors.m
crousselle, line 19: I'm pretty sure this error code is deprecated as I've not been able to get it from testing, I'll reach out to the API team to confirm. However, I would follow the documentation and not what was in Bender.

crousselle, line 19: @thuyBox Just confirmed with the API team that the error code is indeed 400 and the error type "exceeded_device_limits".

thuyBox:
+2 great work!

PR #192, opened 2016-03-16
Add refreshDefaultClientFromKeychain

h2kuo:
Add a method to reset the default client based on what is currently in the keychain. This is useful, for example, in a multi-process situation where one process logs out or changes the user, and other instances of the default client need to be updated to reflect the change.

### DIFF
--- a/BoxContentSDK/BoxContentSDK/Clients/BOXContentClient.m
+++ b/BoxContentSDK/BoxContentSDK/Clients/BOXContentClient.m
@@ -72,6 +72,25 @@ static BOXContentClient *defaultInstance = nil;
     return defaultInstance;
 }
 
++ (void)refreshDefaultClientFromKeychain
+{
+    NSArray *usersFromKeychain = [BOXAbstractSession usersInKeychain];
+    
+    if (usersFromKeychain.count == 0)
+    {
+        [defaultInstance logOut];
+    }
+    else if (usersFromKeychain.count == 1)
+    {
+        BOXUserMini *storedUser = [usersFromKeychain firstObject];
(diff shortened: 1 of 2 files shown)


boxcla:

h2kuo:
FYI @RicoYao

BoxContentSDK/BoxContentSDK/Clients/BOXContentClient.m
```diff
@@ -72,6 +72,33 @@ + (BOXContentClient *)defaultClient
...
+    else {
+        [NSException raise:@"You cannot use 'defaultClient' if multiple users have established a session."
+                    format:@"Specify a user through clientForUser:"];
+    }
+    
+    if (oldDefaultInstance.accessTokenDelegate) {
+        defaultInstance.accessTokenDelegate = oldDefaultInstance.accessTokenDelegate;
+    }
+    oldDefaultInstance.session = defaultInstance.session;
+}
```
thuyBox, line 92: should we update logoutAll to also clear out in-memory user objects of each content client?

BoxContentSDK/BoxContentSDK/Clients/BOXContentClient.m
```diff
@@ -72,6 +72,33 @@ + (BOXContentClient *)defaultClient
     return defaultInstance;
 }
 
++ (void)refreshDefaultClientFromKeychain
+{
+    [[[self class] SDKClients] removeAllObjects];
```
thuyBox: would this remove every content client and not just the default content client?

boxtcarpel:
+2 LGTM

:package:

PR #193, opened 2016-03-17
Introduced a new operation queue in our queue manager to allow more thumbnails to be downloaded in parallel.

crousselle:
This is fixing the thumbnail download performance issue that we're seeing in Flexo

### DIFF
--- a/BoxContentSDK/BoxContentSDK/Operations/BOXAPIDataOperation.h
+++ b/BoxContentSDK/BoxContentSDK/Operations/BOXAPIDataOperation.h
@@ -94,6 +94,13 @@ typedef void (^BOXAPIDataProgressBlock)(long long expectedTotalBytes, unsigned l
  */
 @property (nonatomic, readwrite, strong) BOXAPIDataProgressBlock progressBlock;
 
+
+/**
+ * Describes wether or not the operation is a small download such as a thumbnail retrieval.
+ * If it is, the operation will be run on a specific queue whose max concurrent operation count will be NSOperationQueueDefaultMaxConcurrentOperationCount
+ */
+@property (nonatomic, readwrite, assign) BOOL isSmallDownloadOperation;
+
 /**
  * Call success or failure depending on whether or not an error has occurred during the request.
  * @see successBlock
--- a/BoxContentSDK/BoxContentSDK/QueueManagers/BOXParallelAPIQueueManager.h
+++ b/BoxContentSDK/BoxContentSDK/QueueManagers/BOXParallelAPIQueueManager.h
@@ -48,6 +48,13 @@
  */
 @property (nonatomic, readwrite, strong) NSOperationQueue *uploadsQueue;
 
+/**
+ * The NSOperationQueue on which all small download operations such as thumbnail retrievals
+ * are enqueued. This queue is configured
+ * with `maxConcurrentOperationCount = NSOperationQueueDefaultMaxConcurrentOperationCount`.
+ */
+@property (nonatomic, readwrite, strong) NSOperationQueue *smallDownloadsQueue;
+
 /** @name Designated initializer */
 
 /**
--- a/BoxContentSDK/BoxContentSDK/QueueManagers/BOXParallelAPIQueueManager.m
+++ b/BoxContentSDK/BoxContentSDK/QueueManagers/BOXParallelAPIQueueManager.m
@@ -25,7 +25,7 @@
 @synthesize globalQueue = _globalQueue;
 @synthesize downloadsQueue = _downloadsQueue;
 @synthesize uploadsQueue = _uploadsQueue;
-
+@synthesize smallDownloadsQueue = _smallDownloadsQueue;
 @synthesize currentAccessTokenHasExpired = _currentAccessTokenHasExpired;
 
 - (id)init
@@ -51,6 +51,10 @@
         _uploadsQueue.name = @"BOXParallelAPIQueueManager upload queue";
         _uploadsQueue.maxConcurrentOperationCount = 2;
 
+        _smallDownloadsQueue = [[NSOperationQueue alloc] init];
+        _smallDownloadsQueue.name = @"BOXParallelAPIQueueManager small downloads queue";
+        _smallDownloadsQueue.maxConcurrentOperationCount = NSOperationQueueDefaultMaxConcurrentOperationCount;
+        
         _currentAccessTokenHasExpired = NO;
     }
 
@@ -105,10 +109,17 @@
             }
         }
 
+        
         if ([operation isKindOfClass:[BOXAPIDataOperation class]])
         {
-            [self.downloadsQueue addOperation:operation];
-            BOXLog(@"enqueued %@ on download queue", operation);
+            BOXAPIDataOperation *apiDataOperation = (BOXAPIDataOperation *)operation;
+            if (apiDataOperation.isSmallDownloadOperation) {
+                [self.smallDownloadsQueue addOperation:operation];
+                BOXLog(@"enqueued %@ on thumbnails download queue", operation);
+            } else {
+                [self.downloadsQueue addOperation:operation];
+                BOXLog(@"enqueued %@ on download queue", operation);
+            }
         }
         else if ([operation isKindOfClass:[BOXAPIMultipartToJSONOperation class]])
         {
--- a/BoxContentSDK/BoxContentSDKTests/BOXFileThumbnailRequestTests.m
+++ b/BoxContentSDK/BoxContentSDKTests/BOXFileThumbnailRequestTests.m
@@ -98,6 +98,15 @@
     XCTAssertEqual(1 + numberOfIntermediate202Responses, numberOfOperationsEnqueued);
 }
 
+- (void)test_that_operation_is_marked_as_small_download
+{
+    BOXFileThumbnailRequest *request = [[BOXFileThumbnailRequest alloc] initWithFileID:@"123" size:BOXThumbnailSize64];
+    BOXAPIOperation *operation = [request operation];
+    XCTAssert([operation isKindOfClass:[BOXAPIDataOperation class]]);
+    BOXAPIDataOperation *dataOperation = (BOXAPIDataOperation *)operation;
+    XCTAssert(dataOperation.isSmallDownloadOperation);
+}
+
 #pragma mark - Private helper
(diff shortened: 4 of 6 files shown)


boxcla:

crousselle:
@thuyBox once we push this, there will be a required change on the preview sdk's internal classes : BOXFilePreviewRequest, where we will do a similar change to what we did in BoxFileThumbnailRequest.

crousselle:
@boxtcarpel @RicoYao FYI

BoxContentSDK/BoxContentSDK/Operations/BOXAPIDataOperation.h
```diff
@@ -19,10 +19,10 @@ typedef void (^BOXAPIDataProgressBlock)(long long expectedTotalBytes, unsigned l
 /**
  * BOXAPIDataOperation is a concrete subclass of BOXAPIAuthenticatedOperation.
  * This operation receives binary data from the Box API which may be in the form
- * of downloads or thumbnails.
+ * of downloads or thumbnails. Thumbnail downloads are performed via BOXAPIThumbnailDownloadDataOperation which are runned on a specific queue.
```
boxtcarpel: - of downloads.
- Thumbnail downloads are performed via a BOXAPIThumbnailDataOperation which run on a specific queue.

BoxContentSDK/BoxContentSDK/Operations/BOXAPIThumbnailDownloadDataOperation.h
```diff
@@ -0,0 +1,28 @@
+//
+//  BOXAPIThumbnailDownloadOperation.h
```
boxtcarpel: BOXAPIThumbnailOperation

Also, should we limit it to thumbnails? It's really about small data operations that you might need quickly and in bulk - you could imagine that also being metadata (which we may one day need while scrolling through items for  cell details), or audio mini snippets, or something else.

What do you think of BOXAPISmallDataOperation or something like that?

BoxContentSDK/BoxContentSDK/Operations/BOXAPIThumbnailDownloadDataOperation.h
```diff
@@ -0,0 +1,28 @@
...
+//  BoxContentSDK
+//
+//  Created by Clement Rousselle on 3/16/16.
```
boxtcarpel: We try to clear the created by info for the SDK

BoxContentSDK/BoxContentSDK/Operations/BOXAPIThumbnailDownloadDataOperation.m
```diff
@@ -0,0 +1,13 @@
+//
+//  BOXAPIThumbnailDownloadOperation.m
+//  BoxContentSDK
+//
+//  Created by Clement Rousselle on 3/16/16.
+//  Copyright © 2016 Box. All rights reserved.
+//
+
+#import "BOXAPIThumbnailDownloadDataOperation.h"
```
RicoYao: I would suggest just adding something like a "isThumbnailDownloadOperation" property to BOXAPIDataOperation, rather than creating an empty subclass.

BoxContentSDK/BoxContentSDK/QueueManagers/BOXParallelAPIQueueManager.m
```diff
@@ -51,6 +52,10 @@ - (id)initWithSession:(BOXAbstractSession *)session
         _uploadsQueue.name = @"BOXParallelAPIQueueManager upload queue";
         _uploadsQueue.maxConcurrentOperationCount = 2;
 
+        _thumbnailDownloadsQueue = [[NSOperationQueue alloc] init];
+        _thumbnailDownloadsQueue.name = @"BOXParallelAPIQueueManager thumbnails downloads queue";
+        _thumbnailDownloadsQueue.maxConcurrentOperationCount = 10;
```
RicoYao: 10 seems quite high to me. I suspect you would just get throttled at a deeper network layer actually. Did you choose this based on tests on an actual device? Did you try 3G grade connections?

BoxContentSDK/BoxContentSDK/Operations/BOXAPIThumbnailDownloadDataOperation.m
thuyBox: +1

BoxContentSDK/BoxContentSDK/Operations/BOXAPIDataOperation.h
crousselle: Done.

BoxContentSDK/BoxContentSDK/QueueManagers/BOXParallelAPIQueueManager.h
```diff
@@ -48,6 +48,13 @@
  */
 @property (nonatomic, readwrite, strong) NSOperationQueue *uploadsQueue;
 
+/**
+ * The NSOperationQueue on which all BOXAPIThumbnailDownloadDataOperations
+ * are enqueued. This queue is configured
+ * with `maxConcurrentOperationCount = 10`.
+ */
+@property (nonatomic, readwrite, strong) NSOperationQueue *thumbnailDownloadsQueue;
```
boxtcarpel: small asset download queue?

BoxContentSDK/BoxContentSDK/Operations/BOXAPIThumbnailDownloadDataOperation.m
boxtcarpel: +1

boxtcarpel: Or if we buy into the 'small operation' not-only-thumbnails stance, we should make this isSmallDownload which would be applicable for thumbnails, metadata download, etc.

boxtcarpel:
+2 LGTM

Some comments.

BoxContentSDK/BoxContentSDK/QueueManagers/BOXParallelAPIQueueManager.m
crousselle: it's very hard to say. At some point between very good performance and very good performance, it's hard to see a difference. Those are very small operations data and even on a 3G connection, or at least how the network conditioner interprets a 3G connection, performances are still solid, downgrading the number of connection does not result in a noticeable change.
I'll bring it down to 8, which is the threshold where I stopped noticing performance differences.

BoxContentSDK/BoxContentSDKTests/BOXFileThumbnailRequestTests.m
```diff
@@ -98,6 +98,15 @@ - (void)test_that_thumbnail_request_returns_expected_thumbnail_after_intermediat
     XCTAssertEqual(1 + numberOfIntermediate202Responses, numberOfOperationsEnqueued);
 }
 
+- (void)test_that_operation_is_marked_as_thumbnail_download
+{
+    BOXFileThumbnailRequest *request = [[BOXFileThumbnailRequest alloc] initWithFileID:@"123" size:BOXThumbnailSize64];
+    BOXAPIOperation *operation = [request operation];
+    XCTAssert([operation isKindOfClass:[BOXAPIDataOperation class]]);
+    BOXAPIDataOperation *dataOperation = (BOXAPIDataOperation *)dataOperation;
+    XCTAssert(dataOperation.isSmallDownloadOperation);
```
RicoYao, line 107: Could you please also add a test for regular downloads that asserts that this is false. Thanks!

RicoYao:
LGTM +2

BoxContentSDK/BoxContentSDK/QueueManagers/BOXParallelAPIQueueManager.m
```diff
@@ -51,6 +51,10 @@ - (id)initWithSession:(BOXAbstractSession *)session
         _uploadsQueue.name = @"BOXParallelAPIQueueManager upload queue";
         _uploadsQueue.maxConcurrentOperationCount = 2;
 
+        _smallDownloadsQueue = [[NSOperationQueue alloc] init];
+        _smallDownloadsQueue.name = @"BOXParallelAPIQueueManager thumbnails downloads queue";
```
thuyBox: might want to change the name to reflect small downloads

thuyBox:
+2 :+1: :+1:

BoxContentSDK/BoxContentSDKTests/BOXFileThumbnailRequestTests.m
crousselle, line 107: done.

BoxContentSDK/BoxContentSDK/QueueManagers/BOXParallelAPIQueueManager.m
```diff
@@ -51,6 +51,10 @@ - (id)initWithSession:(BOXAbstractSession *)session
         _uploadsQueue.name = @"BOXParallelAPIQueueManager upload queue";
         _uploadsQueue.maxConcurrentOperationCount = 2;
 
+        _smallDownloadsQueue = [[NSOperationQueue alloc] init];
+        _smallDownloadsQueue.name = @"BOXParallelAPIQueueManager small downloads queue";
+        _smallDownloadsQueue.maxConcurrentOperationCount = NSOperationQueueDefaultMaxConcurrentOperationCount;
```
boxtcarpel, line 56: Was deprecated?
Available in iOS 2.0 through iOS 8.4.

boxtcarpel:
+2 LGTM

:u7a7a:

BoxContentSDK/BoxContentSDK/QueueManagers/BOXParallelAPIQueueManager.m
crousselle, line 56: Yeah this is weird. We've been using this value in Bender, and I don't really notice a performance difference between our iOS 8.3 device and an iOS 9. My guess would be that this removes the limitation after iOS 8.4 and allows us to run as many request as we want in parallel. 
IMO this is fine because we still have a small average of thumbnails requests that can be requested at the same time, testing this on 3G does not show any impact in  the responsiveness of the UI.

PR #201, opened 2016-04-05
Do not release NSDateFormatter and instead reset property values if d…

irvingruan:
…ate formats change.

### DIFF
--- a/BoxContentSDK/BoxContentSDK/Categories/NSDate+BOXAdditions.m
+++ b/BoxContentSDK/BoxContentSDK/Categories/NSDate+BOXAdditions.m
@@ -40,6 +40,7 @@
 
 - (NSString *)box_ISO8601String
 {
+    static NSString *token = @"box_iso8601stringToken";
     static BOXISO8601DateFormatter *dateFormatter;
     static dispatch_once_t pred;
     dispatch_once(&pred, ^{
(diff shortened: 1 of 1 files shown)


boxcla:

boxtcarpel:
We caught up offline. I think we should synchronize the stringFromDate: call in box_ISO8601String in NSDate+BOXAdditions instead. It's the overarching, simplest change that should cover all of the cases. Assuming that our diagnosis is correct and that code is the cause for the crash.

BoxContentSDK/BoxContentSDK/Categories/NSDate+BOXAdditions.m
```diff
@@ -49,8 +49,10 @@ - (NSString *)box_ISO8601String
         dateFormatter.includeTime = YES;
         dateFormatter.defaultTimeZone = [[NSTimeZone alloc] initWithName:@"UTC"];
     });
-    
-    return [dateFormatter stringFromDate:self];
+
+    @synchronized (dateFormatter) {
```
boxtcarpel: Lets be consistent with the method above.

BoxContentSDK/BoxContentSDK/Categories/NSDate+BOXAdditions.m
```diff
@@ -49,8 +49,10 @@ - (NSString *)box_ISO8601String
...
+        return [dateFormatter stringFromDate:self];
```
boxtcarpel: lets set up the string here, inside the synchronized block, and then return it below.

That makes it easier to debug, for breakpoints etc

irvingruan: +1

boxtcarpel:
+2 LGTM

:diamond_shape_with_a_dot_inside:

PR #202, opened 2016-04-05
Set ISO8601DateFormatter to en_US_POSIX locale

adempsey:
We need to specify the locale of our date formatter to ensure we produce consistent results across a variety of device language and time settings. en_US_POSIX produces strings that satisfy RFC 3339.

See https://developer.apple.com/library/mac/qa/qa1480/_index.html


boxcla:

PR #205, opened 2016-04-06
Make sure the authorization header is set not added for requests

boxtcarpel:
There should only be one Bearer header entry for each authenticated request


boxcla:

boxtcarpel:
@adempsey FYI

RicoYao:
+2

PR #206, opened 2016-04-07
Add experimental OS X support

clindsay3:
- Exclude ALAsset-related code
- Exclude AppToApp authentication
- Add OS X target and scheme, OS X unit test target and scheme
- Updates to handle some differences in keychain behavior on OS X vs. iOS
- Replace UIImage with NSImage on OS X
- Add OS X specific user agent
- Update podspec to support OS X

To do:
- Authentication
- User-friendly names for Mac models to include in user agent

### DIFF
--- a/BoxContentSDK/BoxContentSDK/External/KeychainItemWrapper/BOXKeychainItemWrapper.m
+++ b/BoxContentSDK/BoxContentSDK/External/KeychainItemWrapper/BOXKeychainItemWrapper.m
@@ -250,7 +250,9 @@
     // Add the proper search key and class attribute.
     [returnDictionary setObject:(id)kCFBooleanTrue forKey:(id)kSecReturnData];
     [returnDictionary setObject:(id)kSecClassGenericPassword forKey:(id)kSecClass];
-    
+    [returnDictionary removeObjectForKey:kSecAttrModificationDate];
+    [returnDictionary removeObjectForKey:kSecAttrCreationDate];
+
     // Acquire the password data from the attributes.
     NSData *passwordData = NULL;
     OSStatus status = SecItemCopyMatching((CFDictionaryRef)returnDictionary, (CFTypeRef *)&passwordData);
--- a/BoxContentSDK/BoxContentSDK/External/KeychainItemWrapper/BOXKeychainItemWrapper.m
+++ b/BoxContentSDK/BoxContentSDK/External/KeychainItemWrapper/BOXKeychainItemWrapper.m
@@ -253,7 +253,8 @@
     
     // Acquire the password data from the attributes.
     NSData *passwordData = NULL;
-    if (SecItemCopyMatching((CFDictionaryRef)returnDictionary, (CFTypeRef *)&passwordData) == noErr)
+    OSStatus status = SecItemCopyMatching((CFDictionaryRef)returnDictionary, (CFTypeRef *)&passwordData);
+    if (status == noErr)
     {
         // Remove the search, class, and identifier key/value, we don't need them anymore.
         [returnDictionary removeObjectForKey:(id)kSecReturnData];
@@ -266,7 +267,7 @@
     else
     {
         // Don't do anything if nothing is found.
-        NSAssert(NO, @"Serious error, no matching item found in the keychain.\n");
+        NSAssert(NO, @"Serious error: %d, no matching item found in the keychain.\n", status);
     }
     
     [passwordData release];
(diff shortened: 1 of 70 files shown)


boxcla:
Verified that @clindsay has signed the CLA. Thanks for the pull request!

BoxContentSDK/BoxContentSDK/External/KeychainItemWrapper/BOXKeychainItemWrapper.m
```diff
@@ -250,10 +250,13 @@ - (NSMutableDictionary *)secItemFormatToDictionary:(NSDictionary *)dictionaryToC
     // Add the proper search key and class attribute.
     [returnDictionary setObject:(id)kCFBooleanTrue forKey:(id)kSecReturnData];
     [returnDictionary setObject:(id)kSecClassGenericPassword forKey:(id)kSecClass];
-    
+    [returnDictionary removeObjectForKey:kSecAttrModificationDate];
+    [returnDictionary removeObjectForKey:kSecAttrCreationDate];
```
clindsay3, line 254: On OS X a keychain search query that includes a modification date or creation date seems to never return any result. These are included in the result from SecItemCopyMatching that's passed into this method, but shouldn't be necessary to find the correct item, so I've excluded them.

BoxContentSDK/BoxContentSDK/OAuth2/BOXAbstractSession.m
```diff
@@ -227,7 +227,7 @@ + (NSArray *)usersInKeychain
         NSArray* keychainEntries = (__bridge_transfer NSArray*)keychainQueryResult;
         for (NSDictionary *dict in keychainEntries)
         {
-            NSObject *object = [dict objectForKey:((__bridge id)kSecAttrGeneric)];
+            NSObject *object = [dict objectForKey:((__bridge id)kSecAttrAccount)];
```
clindsay3, line 230: kSecAttrGeneric is supposed to have a CFData type. On iOS, it's returning a string when we pass a string into it, but I'm not sure that's behavior we should rely on. In OS X, it always returns a string. Since we put the same value into kSecAttrAccount, and that is supposed to be a string type, I switched to using that value instead.

BoxContentSDK/BoxContentSDKTests/BOXContentSDKTestCase.m
```diff
@@ -302,7 +303,8 @@ - (void)wipeAllKeychainEntries
                            (__bridge id)kSecClassIdentity];
     
     for (id secClass in secClases) {
-        NSMutableDictionary *query = [NSMutableDictionary dictionaryWithObjectsAndKeys:secClass, (__bridge id)kSecClass, nil];
+        NSDictionary *query = @{ (__bridge id)kSecClass: secClass,
+                                 (__bridge id)kSecAttrService: [BOXKeychainItemWrapper keychainServiceIdentifier] };
```
clindsay3, line 307: Since OS X isn't sandboxed, we can't just tell it to wipe the entire keychain (we get a permission denied error if we try to). So we specify instead to delete all items with the Box keychain service identifier.

BoxContentSDK/BoxContentSDKTests/BOXOauth2SessionTests.m
```diff
@@ -122,7 +122,7 @@ - (void)test_that_multiple_credentials_are_stored_to_keychain_correctly
     NSArray* keychainEntries = (__bridge_transfer NSArray*)keychainQueryResult;
     for (NSDictionary *dict in keychainEntries)
     {
-        NSString *keychainIdentifier = [dict objectForKey:((__bridge NSString *)kSecAttrGeneric)];
+        NSString *keychainIdentifier = [dict objectForKey:((__bridge NSString *)kSecAttrAccount)];
```
clindsay3, line 125: See above note about switching from kSecAttrGeneric to kSecAttrAccount.

clindsay3:
I'm not crazy about all of the `#if TARGET_OS_IPHONE`s strewn about. I think it might be cleaner to move some things into Mac-specific and iOS-specific files at some point. Or at least factor things in such a way that we don't need to use #if's to block out pieces in the middle of a method. But it does create the simplest diff to understand, so I think it's good to start from this and evolve it afterwards.

clindsay3:
@RicoYao

clindsay3:
@rknottsbox

rknottsbox:
I agree that the #if TARGET_OS_IPHONE code is a little messy, but it certainly makes it easy to figure out what code is iOS specific.

PR #212, opened 2016-04-18
Fix updated categorie's headers that somehow were changes to 'project' instead of 'public'

crousselle:


boxcla:

PR #216, opened 2016-04-19
Make sure we only clear out completion blocks when we are not re-quei…

scottliubox:
…ng the operation.


boxcla:

h2kuo:
+2

PR #218, opened 2016-04-20
Uncomment out a test case

h2kuo:
Crash only seems to be occurring locally, not when run by Travis CI. We should leave it in to make sure the test is run.


boxcla:

crousselle:
+2

PR #229, opened 2016-06-01
Fix argument types for token expiration method

h2kuo:


boxcla:

BoxContentSDK/BoxContentSDK/OAuth2/BOXParallelOAuth2Session.m
```diff
@@ -46,8 +46,8 @@ - (void)performRefreshTokenGrant:(NSString *)expiredAccessToken withCompletionBl
 }
 
 - (void)performRefreshTokenGrant:(NSString *)expiredAccessToken
-      newAccessTokenExpirationAt:(NSString *)accessTokenExpirationTimestamp
-     newRefreshTokenExpirationAt:(NSString *)refreshTokenExpirationTimestamp
+      newAccessTokenExpirationAt:(NSNumber *)accessTokenExpirationTimestamp
```
boxtcarpel, line 49: should we clean up the naming as well? An nsnumber is likely not a timestamp, probably an offset?

boxtcarpel:
+2 LGTM

Minor comment about naming.

🐻

BoxContentSDK/BoxContentSDK/OAuth2/BOXParallelOAuth2Session.m
h2kuo, line 49: Since the API documentation refers to this as a timestamp, we'll leave it this way for now to be consistent.

PR #231, opened 2016-06-01
Call completion block after BOXParallelOAuth2Session token refresh

h2kuo:
Previously, the completion block would only be called if there was already an existing refresh in progress.


boxcla:

boxtcarpel:
+2 LGTM

🍶

PR #232, opened 2016-06-01
Make BOXContentSDK extension-safe : introducing an UIApplication category that will avoid compiling forbidden extensions APIs.

crousselle:


boxcla:

BoxContentSDK/BoxContentSDK/Categories/UIApplication+ExtensionSafeCompilation.m
```diff
@@ -0,0 +1,47 @@
+//
+//  UIApplication+ExtensionSafeCompilation.m
+//  app
```
scottliubox: probably delete "app" here. Same for .h

BoxContentSDK/BoxContentSDK/Categories/UIApplication+ExtensionSafeCompilation.m
```diff
@@ -0,0 +1,47 @@
...
+    return nil;
+#else
+    // If we are compiling from the main app, return normal sharedApplication.
+    return [UIApplication sharedApplication];
+#endif
+}
+
+- (BOOL)box_canOpenURL:(NSURL *)url
+{
+#ifdef EXTENSION
```
scottliubox: maybe re-name this to something more explicit? Maybe CompileForExtension

crousselle: this should be to specify that the current target is an extension. What about TARGET_IS_EXTENSION ?

scottliubox: That sounds good, let's go with that.

BoxContentSDK/BoxContentSDK/Categories/UIApplication+ExtensionSafeCompilation.m
```diff
@@ -0,0 +1,44 @@
...
+#ifdef TARGET_IS_EXTENSION
+    // If we are compiling from an extension, return nil here since UIApplication isn't extension-safe.
+    return nil;
+#else
+    // If we are compiling from a non-extension target, use the regular sharedApplication.
+    return [UIApplication sharedApplication];
+#endif
+}
+
+- (BOOL)box_canOpenURL:(NSURL *)url
```
boxtcarpel: why do we need these? If you call box_sharedApplication you're set. Is there a reason we can't use box_SharedApplication in some cases?

BoxContentSDK/BoxContentSDK/OAuth2/BOXAuthorizationViewController.m
```diff
@@ -355,7 +356,7 @@ - (BOOL)webView:(UIWebView *)webView shouldStartLoadWithRequest:(NSURLRequest *)
 
     // Mailto URLs can be encountered if there is a hyperlink for sending an email.
     if ([[[[request URL] scheme] lowercaseString] isEqualToString:@"mailto"]) {
-        [[UIApplication sharedApplication] openURL:[request URL]];
+        [[UIApplication box_sharedApplication] box_openURL:[request URL]];
```
boxtcarpel, line 359: couldn't this just call the regular openURL method? Or will that not compile for extensions??

crousselle, line 359: @scottliubox is there any resaons behind those implementations (this is the category that was in the app)

crousselle, line 359: @boxtcarpel somehow Xcode throws an error if we do not use our implementations. Hence the category methods.

scottliubox, line 359: XCode actually throws an error for using openURL itself, it specifically mentions that method not being extension safe and doesn't let you compile.

scottliubox:
+2

PR #233, opened 2016-06-02
Fix the extension safe UIApplication category to be in the right setup.

crousselle:


boxcla:

crousselle:
@scottliubox FYI. Order matters.

PR #238, opened 2016-06-07
Resolve an issue with the default behavior for safe extension compilation

boxtcarpel:
Make the app-level behavior the default and extension behavior declarable
via pre-processor const.
This also fixes the functionality of the sample app.


boxcla:

crousselle:
cool, +2 !

PR #239, opened 2016-06-14
Fix the UIApplication category logic that was reversed.

crousselle:


boxcla:

PR #240, opened 2016-06-14
Update the UIApplication category to be truly extension safe if the box-ios-sdk is used as a dylib.

crousselle:
If compiling with `APPLICATION_EXTENSION_API_ONLY` set to `YES` and using the SDK as a pod, compilation would fail as calls to `sharedApplication` would be detected by the compiler, even with build flags checks.

### DIFF
--- a/BoxContentSDK/BoxContentSDK/Categories/UIApplication+ExtensionSafeAdditions.m
+++ b/BoxContentSDK/BoxContentSDK/Categories/UIApplication+ExtensionSafeAdditions.m
@@ -8,37 +8,65 @@
 
 @implementation UIApplication (ExtensionSafeAdditions)
 
-+ (UIApplication *)box_sharedApplication
++ (BOOL)box_isRunningExtension
 {
-#ifndef BOX_TARGET_IS_EXTENSION
-    // If we are compiling from a non-extension target, use the regular sharedApplication.
-    return [UIApplication sharedApplication];
-#else
-    // If we are compiling from an extension, return nil here since UIApplication isn't extension-safe.
-    return nil;
-#endif
+    return [[[NSBundle mainBundle] executablePath] containsString:@".appex/"];
 }
 
++ (UIApplication *)box_sharedApplication
+{
+    UIApplication *application = nil;
+    if ([UIApplication box_isRunningExtension] == NO) {
+        // If we are compiling from a non-extension target, use the regular sharedApplication.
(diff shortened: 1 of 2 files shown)


boxcla:

BoxContentSDK/BoxContentSDK/Categories/UIApplication+ExtensionSafeAdditions.m
```diff
@@ -12,18 +12,23 @@ + (UIApplication *)box_sharedApplication
 {
 #ifndef BOX_TARGET_IS_EXTENSION
```
boxtcarpel: Do we want to keep the conditional compiling? If it's a dylib from a pod, it will only get compiled once for all targets including extensions, no?

crousselle: yeah but during execution, what would happen when the extension calls those methods ? it would result in calling the forbidden APIs

boxtcarpel:
Discussed offline, @crousselle will follow-up

BoxContentSDK/BoxContentSDK/Categories/UIApplication+ExtensionSafeAdditions.m
crousselle: synced up offline. This compiles regardless of the target since it's a dylib. need to remove compile time checks for runtime checks.

BoxContentSDK/BoxContentSDK/Categories/UIApplication+ExtensionSafeAdditions.m
```diff
@@ -8,37 +8,65 @@
...
++ (BOOL)box_isRunningExtension
 {
-#ifndef BOX_TARGET_IS_EXTENSION
-    // If we are compiling from a non-extension target, use the regular sharedApplication.
-    return [UIApplication sharedApplication];
-#else
-    // If we are compiling from an extension, return nil here since UIApplication isn't extension-safe.
-    return nil;
-#endif
+    return [[[NSBundle mainBundle] executablePath] containsString:@".appex/"];
```
crousselle, line 13: we wil go this direction as it seems to be the more reliable until we can syncup with apple engineers at WWDC about a possible better approach.

scottliubox:
+2

PR #245, opened 2016-08-16
IOS-12035: Allow to download file object excluding specific file fields

thuyBox:


boxcla:
Hi @thuyBox, thanks for the pull request. Before we can merge it, we need you to sign our Contributor License Agreement. You can do so electronically here: http://opensource.box.com/cla

thuyBox:
@RicoYao

BoxContentSDK/BoxContentSDK/Requests/BOXFolderPaginatedItemsRequest.h
```diff
@@ -10,6 +10,7 @@
 @property (nonatomic, readonly, strong) NSString *folderID;
 @property (nonatomic, readonly, assign) NSUInteger limit;
 @property (nonatomic, readonly, assign) NSUInteger offset;
+@property (nonatomic, readwrite, assign) BOOL requestPerformantItemFields;
```
boxtcarpel: Should 'performantItemFields' be exposed at the same level as AllItemFields?

BoxContentSDK/BoxContentSDK/Requests/BOXRequest.m
```diff
@@ -300,6 +300,40 @@ - (NSArray *)fullFileFieldsArray
...
+                       BOXAPIObjectKeyPurgedAt,
+                       BOXAPIObjectKeyContentCreatedAt,
+                       BOXAPIObjectKeyContentModifiedAt,
+                       BOXAPIObjectKeyCreatedBy,
+                       BOXAPIObjectKeyModifiedBy,
+                       BOXAPIObjectKeyOwnedBy,
+                       BOXAPIObjectKeySharedLink,
+                       BOXAPIObjectKeyParent,
+                       BOXAPIObjectKeyItemStatus,
+                       //BOXAPIObjectKeyVersionNumber, excluded could reduce ~70% API response time
```
boxtcarpel: Alternatively, could get fullFieldsArray and remove BOXAPIObjectKeyVersionNumber before returning it.

Advantage: One place where all fields are listed - no duplication, new fields automatically get into 'performant' listing.
Disadvantage: If a new field is introduced that is also non-performant, we need to know about this and remember to add a removal step for it as well.

What do you think?

BoxContentSDK/BoxContentSDK/Requests/BOXRequest.m
```diff
@@ -336,6 +370,15 @@ - (NSString *)fullBookmarkFieldsParameterString
     return [[self fullBookmarkFieldsArray] componentsJoinedByString:@","];
 }
 
+- (NSString *)performantItemFieldsParameterString
+{
+    NSMutableOrderedSet *set = [NSMutableOrderedSet orderedSet];
+    [set addObjectsFromArray:[self fullFolderFieldsArray]];
```
boxtcarpel, line 342: With these full arrays in here we may actually accidentally reintroduce the non-performant fields, since the fields are generic and can be applied to any item, today folder and bookmark don't have 'BOXAPIObjectKeyVersionNumber' but if that was added tomorrow (may make sense for bookmark for example) and we added it to the full fields array, that would creep into this performant parameter string and we would have the same problem over again.

BoxContentSDK/BoxContentSDK/Requests/BOXFolderItemsRequest.h
```diff
@@ -12,6 +12,7 @@
 
 @property (nonatomic, readwrite, strong) NSString *folderID;
 @property (nonatomic, readwrite, assign) BOOL requestAllItemFields;
+@property (nonatomic, readwrite, assign) BOOL requestPerformantItemFields;
```
boxtcarpel: Where does it make sense to add a comment that explains this? Unfortunately 'performant item fields' is not as self explanatory or clear as 'all item fields' IMO.

thuyBox:
@boxtcarpel, all your questions are valid. I've updated the approach which addresses most of the questions. Will upload the new patch soon.

BoxContentSDK/BoxContentSDK/Requests/BOXFolderItemsRequest.h
```diff
@@ -12,6 +12,7 @@
 
 @property (nonatomic, readwrite, strong) NSString *folderID;
 @property (nonatomic, readwrite, assign) BOOL requestAllItemFields;
+@property (nonatomic, readwrite, strong) NSArray *requestAllItemFieldsAndExcludingFields;
```
boxtcarpel: Name suggestion: fieldsToExclude/WhenRequestingItemFields

BoxContentSDK/BoxContentSDK/Requests/BOXCollectionItemsRequest.m
```diff
@@ -38,6 +38,8 @@ - (BOXAPIOperation *)createOperation
     
     if (self.requestAllItemFields) {
         parameters[BOXAPIParameterKeyFields] = [self fullItemFieldsParameterString];
+    } else if (self.requestAllItemFieldsAndExcludingFields) {
```
boxtcarpel: so if we change the meaning of the array (to not also be it's own flag), then this moves out of the if block.

BoxContentSDK/BoxContentSDK/Requests/BOXFolderPaginatedItemsRequest.m
```diff
@@ -38,6 +38,8 @@ - (BOXAPIOperation *)createOperation
 
     if (self.requestAllItemFields) {
         queryParameters[BOXAPIParameterKeyFields] = [self fullItemFieldsParameterString];
+    } else if (self.requestAllItemFieldsAndExcludingFields) {
```
boxtcarpel: same

boxtcarpel:
We discussed this offline. One more slight shift to the approach.

BoxContentSDK/BoxContentSDK/Requests/BOXRequest.m
```diff
@@ -345,6 +345,19 @@ - (NSString *)fullItemFieldsParameterString
     return [[set array] componentsJoinedByString:@","];
 }
 
+- (NSString *)fullItemFieldsParameterStringExcludingFields:(NSArray *)excludedFields
+{
+    NSMutableOrderedSet *set = [NSMutableOrderedSet orderedSet];
+    [set addObjectsFromArray:[self fullFolderFieldsArray]];
+    NSArray * fileFields = [self fullFileFieldsArray];
+    if (excludedFields.count > 0) {
+        fileFields = [[NSMutableArray arrayWithArray:allFileFields] removeObjectsInArray:excludedFields];
```
boxtcarpel: right now we only apply this to file. Should we build the combined set, then remove excluded fields from the union then turn it into an array?

boxtcarpel:
1 comment, otherwise LGTM +2

💠

BoxContentSDK/BoxContentSDK/Requests/BOXRequest.m
thuyBox: done

boxtcarpel:
+2 LGTM

boxtcarpel:
+2

PR #250, opened 2016-08-24
Add representations property to BOXFile

h2kuo:


boxcla:

h2kuo:
Not yet ready to merge, but wanted to start the review process and get feedback. More info/links to the representations documentation are here: https://cloud.app.box.com/notes/74871682842

BoxContentSDK/BoxContentSDK/Models/BOXFile.m
```diff
@@ -79,6 +80,20 @@ - (instancetype)initWithJSON:(NSDictionary *)JSONResponse
         if (lock && ![lock isKindOfClass:[NSNull class]]) {
             self.lock = [[BOXFileLock alloc] initWithJSON:lock];
         }
+        
+        // Parse Representations.
+        NSDictionary *representationsJSON = [NSJSONSerialization box_ensureObjectForKey:BOXAPIObjectKeyRepresentations
+                                                                          inDictionary:JSONResponse
+                                                                       hasExpectedType:[NSDictionary class]
+                                                                           nullAllowed:NO];
```
RicoYao, line 88: for crash defensiveness, let's just make nullAllowed YES

h2kuo: `nullAllowed:YES suppressNullAsNil:YES` or just `nullAllowed:YES`?

BoxContentSDK/BoxContentSDK/Models/BOXRepresentation.m
```diff
@@ -0,0 +1,77 @@
...
+        self.status = [NSJSONSerialization box_ensureObjectForKey:BOXAPIObjectKeyStatus
+                                                     inDictionary:JSONResponse
+                                                  hasExpectedType:[NSString class]
+                                                      nullAllowed:NO];
+        
+        NSDictionary *detailsJSON = [NSJSONSerialization box_ensureObjectForKey:BOXAPIObjectKeyDetails
+                                                                   inDictionary:JSONResponse
+                                                                hasExpectedType:[NSDictionary class]
+                                                                    nullAllowed:NO];
+        if ([detailsJSON count] > 0) {
```
RicoYao: Purpose of the 'count' check?

BoxContentSDK/BoxContentSDK/Models/BOXFile.m
RicoYao: I think it should be:
nullAllowed:YES suppressNullAsNil:YES

Could probably do the same for almost all properties in BOXRepresentation

RicoYao:
Could we please add some tests for the parsing of the representation json objects

h2kuo:
Yes, will add tests before merging. Right now, the JSON data would still contain internal URLs so wanted to wait until they moved to the public endpoint.

RicoYao:
I think the tests would just contain dummy URLs anyway so you may not need to wait.

BoxContentSDK/BoxContentSDK/BOXContentSDKConstants.m
```diff
@@ -303,7 +335,7 @@
 NSString *const BOXAPIEventStreamTypeSync = @"sync";
 NSString *const BOXAPIEventStreamTypeAdminLogs = @"admin_logs";
 
-// Standart Events
+// Standard Events
```
RicoYao, line 338: :)

BoxContentSDK/BoxContentSDK/Models/BOXRepresentation.m
```diff
@@ -0,0 +1,77 @@
...
+                                                   inDictionary:JSONResponse
+                                                hasExpectedType:[NSString class]
+                                                    nullAllowed:NO];
+        
+        NSDictionary *propertiesJSON = [NSJSONSerialization box_ensureObjectForKey:BOXAPIObjectKeyProperties
+                                                                  inDictionary:JSONResponse
+                                                               hasExpectedType:[NSDictionary class]
+                                                                   nullAllowed:NO];
+        
+        self.dimensions = [propertiesJSON objectForKey:BOXAPIObjectKeyDimensions];
```
RicoYao, line 31: Should we use NSJSONSerialization box_ensureObjectForKey just to be safe?

BoxContentSDK/BoxContentSDK/Models/BOXRepresentation.m
h2kuo: If there are no details, the API still returns an empty dictionary, so thought we should check to see if it has anything in it before setting the property. Is it better for 1) the API to return null if there's nothing in it or 2) just set it as an empty dictionary?

RicoYao: Probably would be ok either way either way in real world use, but my vote would be (2), since  that reflects what the API returned.

jdizoglio:
+1 nothing to add

BoxContentSDK/BoxContentSDK/Models/BOXRepresentation.m
```diff
@@ -0,0 +1,77 @@
...
+        }
+        
+        self.contentType = [NSJSONSerialization box_ensureObjectForKey:BOXAPIObjectKeyType
+                                                          inDictionary:contentJSON
+                                                       hasExpectedType:[NSString class]
+                                                           nullAllowed:NO];
+        
+        NSDictionary *infoJSON = linksJSON[BOXAPIObjectKeyInfo];
+        
+        self.infoURL = [NSJSONSerialization box_ensureObjectForKey:BOXAPIObjectKeyURL
```
thuyBox, line 66: looks like we use the same key and dict to extract contentURLString and self.infoURL? probably some typo?

BoxContentSDK/BoxContentSDK/Models/BOXRepresentation.m
```diff
@@ -0,0 +1,77 @@
...
+                                                              hasExpectedType:[NSDictionary class]
+                                                                  nullAllowed:NO];
+        
+        NSDictionary *contentJSON = linksJSON[BOXAPIObjectKeyContent];
+        
+        NSString *contentURLString = [NSJSONSerialization box_ensureObjectForKey:BOXAPIObjectKeyURL
+                                                             inDictionary:contentJSON
+                                                          hasExpectedType:[NSString class]
+                                                              nullAllowed:NO];
+        if (contentURLString.length > 0) {
```
thuyBox, line 55: do we need this check? or will it automatically return nil if contentURLString.length == 0 or contentURLString == nil?

if we do self.contentURL = [NSURL URLWithString:contentURLString] always, it serves as initializing the property to nil accordingly as well.

BoxContentSDK/BoxContentSDK/Requests/BOXRequest.m
```diff
@@ -296,7 +296,8 @@ - (NSArray *)fullFileFieldsArray
                        BOXAPIObjectKeyExtension,
                        BOXAPIObjectKeyIsPackage,
                        BOXAPIObjectKeyAllowedSharedLinkAccessLevels,
-                       BOXAPIObjectKeyCollections];
+                       BOXAPIObjectKeyCollections,
+                       BOXAPIObjectKeyRepresentations];
```
thuyBox, line 300: do you know how performant the request for representations is? just wonder if we should exclude this field in the app, and just have it used by preview sdk.

BoxContentSDK/BoxContentSDK/Models/BOXRepresentation.h
```diff
@@ -0,0 +1,42 @@
...
+//
+//  Created on 7/15/16.
+//  Copyright © 2016 Box. All rights reserved.
+//
+
+#import <Foundation/Foundation.h>
+#import "BOXContentSDKConstants.h"
+
+@interface BOXRepresentation : NSObject
+
```
thuyBox, line 13: could we add comment for each of the field in this object?

BoxContentSDK/BoxContentSDK/Models/BOXRepresentation.m
```diff
@@ -0,0 +1,77 @@
...
+            self.contentURL = [NSURL URLWithString:contentURLString];
+        }
+        
+        self.contentType = [NSJSONSerialization box_ensureObjectForKey:BOXAPIObjectKeyType
+                                                          inDictionary:contentJSON
+                                                       hasExpectedType:[NSString class]
```
thuyBox, line 61: should expected type be BOXRepresentationContentType instead of NSString?

BoxContentSDK/BoxContentSDK/BOXContentSDKConstants.h
```diff
@@ -145,6 +145,37 @@ extern BOXCollaborationRole *const BOXCollaborationRoleViewer;
...
+extern BOXRepresentationStatus *const BOXRepresentationStatusPending;
+extern BOXRepresentationStatus *const BOXRepresentationStatusNone;
+
+// Representation content type
+typedef NSString BOXRepresentationContentType;
+extern BOXRepresentationContentType *const BOXRepresentationContentTypeAsset;
+extern BOXRepresentationContentType *const BOXRepresentationContentTypeManifest;
+
+// Representation dimensions
+typedef NSString BOXRepresentationDimensions;
```
thuyBox, line 176: should the type name be singular? i.e. BOXRepresentationDimension without 's'? since each value contains only 1 dimension?

BoxContentSDK/BoxContentSDK/Models/BOXRepresentation.h
```diff
@@ -0,0 +1,42 @@
...
+@property (nonatomic, readwrite, strong) NSURL *infoURL;
+
+@property (nonatomic, readwrite, strong) NSURL *contentURL;
+
+@property (nonatomic, readwrite, strong) BOXRepresentationContentType *contentType;
+
+@property (nonatomic, readwrite, strong) NSDictionary *details;
+
+@property (nonatomic, readwrite, strong) BOXRepresentationDimensions *dimensions;
```
thuyBox, line 26: should this be 'dimension' without 's' if it represents only 1 dimension?

BoxContentSDK/BoxContentSDK/BOXContentSDKConstants.h
h2kuo: It's actually 2 dimensions since it's width x height :) The JSON key is also "dimensions" so I think it's fine to leave it this way.

BoxContentSDK/BoxContentSDK/Requests/BOXRequest.m
h2kuo: It's currently not very performant, would definitely consider adding it to exclude fields.

BoxContentSDK/BoxContentSDK/Models/BOXRepresentation.m
h2kuo: Good catch, the key should be the same (it's just "url") but the dictionary here should be infoJSON.

BoxContentSDK/BoxContentSDK/Models/BOXRepresentation.h
h2kuo: No, we should leave it dimensions (see other comment)

BoxContentSDK/BoxContentSDK/Models/BOXRepresentation.m
h2kuo: No, I think it's fine to leave it as NSString because that's the type we expect in the JSON dictionary.

jdizoglio:
This was migrated to #268, cherry picked into Hamilton and now branch xcode8. Will close this pull request.

PR #258, opened 2016-09-01
IOS-12165: fix crash upon launching EMM because we try to extract use…

thuyBox:
…rID from a string userInfo

### DIFF
--- a/BoxContentSDK/BoxContentSDK/Clients/BOXContentClient.m
+++ b/BoxContentSDK/BoxContentSDK/Clients/BOXContentClient.m
@@ -237,10 +237,12 @@ static BOXContentClient *defaultInstance = nil;
 
 - (void)didReceiveUserWasLoggedOutNotification:(NSNotification *)notification
 {
-    NSDictionary *userInfo = (NSDictionary *)notification.object;
-    NSString *userID = [userInfo objectForKey:BOXUserIDKey];
-    if ([userID isEqualToString:self.user.modelID]) {
-        [self logOut];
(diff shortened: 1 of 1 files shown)


boxcla:

boxtcarpel:
+2 LGTM

💃

BoxContentSDK/BoxContentSDK/Clients/BOXContentClient.m
```diff
@@ -237,10 +237,12 @@ - (void)didReceiveAccessTokenRefreshedNotification:(NSNotification *)notificatio
 
 - (void)didReceiveUserWasLoggedOutNotification:(NSNotification *)notification
 {
-    NSDictionary *userInfo = (NSDictionary *)notification.object;
-    NSString *userID = [userInfo objectForKey:BOXUserIDKey];
-    if ([userID isEqualToString:self.user.modelID]) {
-        [self logOut];
+    if ([notification.object isKindOfClass:[NSDictionary class]]) {
```
RicoYao, line 240: I think this was caused by commit dda2df7eff30 (part of the App Users work).

Can we just do a refactor and stop using the 'object' for this. It really should go into userInfo. I believe there's only one component that posts this notification in BOXAPIDataOperation. It'd be great if we could just clean it up.

boxtcarpel: Agreed 100%. We had that discussion, that is IOS-12164. Thuy did not want to do that for 3.8.X releases, so we'll do it in Master for Hamilton. I want to do this next week.

PR #259, opened 2016-09-02
IOS-12165: fix crash upon launching EMM because we try to extract use…

thuyBox:
…rID from a string userInfo

same as https://github.com/box/box-ios-sdk/pull/258


boxcla:

PR #267, opened 2016-09-30
Auth VC should respect app users mode.

RicoYao:


boxcla:

BoxContentSDK/BoxContentSDK/OAuth2/BOXAuthorizationViewController.m
```diff
@@ -139,7 +142,14 @@ - (void)viewDidLoad
 {
     [super viewDidLoad];
 
-    [self loadAuthorizationURL];
+    if ([self.SDKClient.session isKindOfClass:[BOXOAuth2Session class]]) {
```
boxtcarpel, line 145: We couldn't do this based on the token delegate?

BoxContentSDK/BoxContentSDK/OAuth2/BOXAuthorizationViewController.m
```diff
@@ -139,7 +142,14 @@ - (void)viewDidLoad
...
+        [self loadAuthorizationURL];
+    } else if ([self.SDKClient.session isKindOfClass:[BOXAppUserSession class]]) {
+        __weak BOXAuthorizationViewController *me = self;
+        [self.SDKClient autheticateAppUserWithCompletionBlock:^(BOXUser *user, NSError *error) {
```
boxtcarpel, line 149: If we're using a BOXAppUserSession, don't we know for a fact that this is what will get called by ' - (void)authenticateWithCompletionBlock:(void (^)(BOXUser *user, NSError *error))completion;' ?

BoxContentSDK/BoxContentSDK/OAuth2/BOXAuthorizationViewController.m
RicoYao: I wanted to match the way we're checking this in other classes.

BoxContentSDK/BoxContentSDK/OAuth2/BOXAuthorizationViewController.m
RicoYao: Discussed in-person. We're good with this current approach as is.

boxtcarpel:
+2 LGTM

💍

PR #269, opened 2016-10-03
Make sure that cancelling a folder items request actually works

boxtcarpel:
Needed to make sure we propagate the cancel to the paginated items request.

### DIFF
--- a/BoxContentSDK/BoxContentSDK/Requests/BOXFolderItemsRequest.m
+++ b/BoxContentSDK/BoxContentSDK/Requests/BOXFolderItemsRequest.m
@@ -15,7 +15,10 @@
 #import "BOXContentSDKErrors.h"
 
 @interface BOXFolderItemsRequest ()
+
 @property (nonatomic) BOOL isCancelled;
+@property (nonatomic, readwrite, strong) BOXFolderPaginatedItemsRequest *paginatedRequest;
+
 @end
 
 @implementation BOXFolderItemsRequest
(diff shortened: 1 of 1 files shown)


boxcla:

crousselle:
good catch, +2

BoxContentSDK/BoxContentSDK/Requests/BOXFolderItemsRequest.m
```diff
@@ -95,6 +100,8 @@ - (void)performRequestWithCached:(BOXItemsBlock)cacheBlock refreshed:(BOXItemsBl
         NSMutableArray *results = [[NSMutableArray alloc] init];
 
         BOXItemsBlock localRefreshBlock = ^(NSArray *items, NSError *error){
+            self.paginatedRequest = nil;
```
thuyBox, line 103: should we call this as first thing in recursiveFetch/paginatedRefreshItemFetch block instead of in localRefreshBlock?

boxtcarpel: Well, I'm trying to avoid changing it too much, or too hastily.
When we create and set off a new paginated request we set it as the paginated request, and if we complete the existing one we nil our paginated request. Seems complete to me on handling each leaf in the tree.
What is your concern? Seem like we're missing anything?

thuyBox:
+2!

boxtcarpel:
Thanks!

PR #280, opened 2016-11-16
Make sure the base URL used for loadHTMLString:baseURL: is an NSURL

boxtcarpel:


boxcla:

boxtcarpel:
@adempsey please review retroactively. I also still need to test this in a scenario where an HTML error is returned as such, for now pushing this to unblock testing and development.

adempsey:
+2

PR #281, opened 2016-11-17
Use about:blank instead of box.com for the base URL in web views

boxtcarpel:
The base URL gets loaded when the string passed in to loadHTMLString:baseURL:
is not valid HTML. In those cases, we do not want to show the box.com landing
page as that will be confusing to users.


boxcla:

boxtcarpel:
Tested auth on iPhone and iPad, login success and failure. On success, we get a white screen before the login interface goes away, as opposed to the box.com page that was shown before.

thuyBox:
+2 what tests did you perform?

PR #288, opened 2016-12-21
Add collections field to bookmarks full field listing

adempsey:
A change was made recently to add `collections` to the full response header for requested bookmarks, but the corresponding test wasn't updated to account for it.


boxcla:

scottliubox:
+2

PR #289, opened 2016-12-23
Allow API URLs to be settable for development.

RicoYao:


boxcla:

BoxContentSDK/BoxContentSDK/Clients/BOXContentClient.h
```diff
@@ -180,4 +170,13 @@
  **/ 
 - (void)setSharedLinkStorageDelegate:(id <BOXSharedLinkStorageProtocol>)delegate;
 
+/**
+ *  API base URLs.
+ **/
++ (NSString *)APIBaseURL;
```
boxtcarpel, line 176: Some confusion with these being named xURL, but being NSString objects

BoxContentSDK/BoxContentSDK/Clients/BOXContentClient.m
```diff
@@ -410,4 +402,57 @@ + (void)resetInstancesForTesting
...
+
++ (NSString *)APIUploadBaseURL
+{
+    if (staticAPIUploadBaseURL.length == 0) {
+        staticAPIUploadBaseURL = BOXDefaultAPIUploadBaseURL;
+    }
+    return staticAPIUploadBaseURL;
+}
+
++ (void)setAPIBaseURL:(NSString *)APIBaseURL
```
boxtcarpel, line 437: Should you be able to do this on an instance of client, is opposed to all clients? (in which case might need to adjust the static approach)

BoxContentSDK/BoxContentSDK/OAuth2/BOXOAuth2Session.m
```diff
@@ -299,7 +299,7 @@ - (void)revokeCredentials
 - (void)sendRevokeRequest
 {
     // We don't go through any of our regular queues/operations because those get shut down upon logout.
-    NSMutableURLRequest *request = [NSMutableURLRequest requestWithURL:[NSURL URLWithString:[NSString stringWithFormat:@"%@/oauth2/revoke", BOXAPIBaseURL]]];
+    NSMutableURLRequest *request = [NSMutableURLRequest requestWithURL:[NSURL URLWithString:[NSString stringWithFormat:@"%@/oauth2/revoke", self.APIBaseURLString]]];
```
boxtcarpel: No version number on revoke API call
https://docs.box.com/reference#revoke

BoxContentSDK/BoxContentSDK/Requests/BOXRequest.m
```diff
@@ -14,6 +14,7 @@
 #import "BOXAppUserSession.h"
 #import "NSString+BOXContentSDKAdditions.h"
 #import "UIDevice+BOXContentSDKAdditions.h"
+#import "BOXContentClient.h"
```
boxtcarpel, line 17: I guess we have to have this dependency...

BoxContentSDK/BoxContentSDK/Requests/BOXRequest.m
```diff
@@ -135,8 +112,7 @@ - (NSURL *)URLWithResource:(NSString *)resource
     return [self URLWithResource:resource ID:ID
                      subresource:subresource
                            subID:subID
-                         baseURL:self.baseURL
-                      APIVersion:self.APIVersion];
+                         baseURL:self.baseURL];
```
boxtcarpel, line 115: I guess originally our idea was that these would be settable at this layer.

BoxContentSDK/BoxContentSDK/Clients/BOXContentClient.m
```diff
@@ -22,6 +22,11 @@
 #import "BOXUserRequest.h"
 #import "BOXContentClient+User.h"
 
+// Default API URLs
+NSString *const BOXDefaultAPIBaseURL = @"https://api.box.com/2.0";
```
boxtcarpel, line 26: Do you think that including the version number could cause confusion. You'd have to know to set the base URL with the version number (this would have to be well documented, I can see this being easy to miss even for our own use).

Then this becomes a problem when the version number is not desirable... is this really the true base URL?
Should we split it to baseURL and baseURLWithVersion?

BoxContentSDK/BoxContentSDK/Clients/BOXContentClient.m
RicoYao: I wanted it to be static for all clients, and in fact I intentionally added code to proactively set the URL on all existing session objects in memory. There would have to be a **very** compelling use-case to justify allowing the URLs to differ between client/session instances IMHO.

BoxContentSDK/BoxContentSDK/Requests/BOXRequest.m
RicoYao: Could you please clarify what action you want taken based on that comment.

BoxContentSDK/BoxContentSDK/Requests/BOXRequest.m
RicoYao: Could you please clarify what action you want taken based on that comment.

BoxContentSDK/BoxContentSDK/Clients/BOXContentClient.h
RicoYao: I was carrying over naming based on the property name. I could change it to end with 'URLString' instead, but probably unnecessary? Let me know if you prefer that.

BoxContentSDK/BoxContentSDK/Clients/BOXContentClient.m
RicoYao: Makes sense, I will:
1. Add a const for BOXDefaultAPIBaseURLWithoutVersion = https://api.box.com
2. Add class methods 'APIBaseURLWithoutVersion' (getter) and 'setAPIBaseURLWithoutVersion' (setter)

BoxContentSDK/BoxContentSDK/OAuth2/BOXOAuth2Session.m
RicoYao: Nice catch. Will fix.

BoxContentSDK/BoxContentSDK/Requests/BOXRequest.m
boxtcarpel: I'm not sure we'd do things differently.
I guess one option would be to try and make the URL-switching happen around the request object perhaps at the 'prepareRequest'. That way there is still an awareness of the Request in the Client, but not the awareness of the Client in the Request. Not sure if that is feasible or if so, that's it's definitely better. I'll leave it to you.

BoxContentSDK/BoxContentSDK/Requests/BOXRequest.m
boxtcarpel: See above comment.

BoxContentSDK/BoxContentSDK/Clients/BOXContentClient.h
boxtcarpel: Yeah, I saw that it was already that way. Not a priority right now and for this PR, I don't think. But just calling it out so it's in our consciousness. Ultimately, and as we introduce new things, we want to be unambiguous.

BoxContentSDK/BoxContentSDK/Clients/BOXContentClient.m
RicoYao: See the new way that I've distinguished btwn the URLs.

BoxContentSDK/BoxContentSDK/Clients/BOXContentClient.h
```diff
@@ -180,4 +170,15 @@
...
++ (NSString *)OAuth2BaseURL;
++ (NSString *)APIAuthBaseURL;
```
boxtcarpel, line 178: As discussed offline, a bit of a better way to differentiate APIAuthBaseURL from OAuth2BaseURL without having to dig into the code around them (whether that's in naming, comments, etc).

BoxContentSDK/BoxContentSDK/OAuth2/BOXOAuth2Session.m
```diff
@@ -18,6 +18,7 @@
 #import "BOXUser_Private.h"
 #import "NSDate+BOXContentSDKAdditions.h"
 #import "BOXAbstractSession_Private.h"
+#import "BOXContentClient.h"
```
boxtcarpel, line 21: What are your thoughts on this as a reason to pass in the URL/s to use?

Instead of querying up into the content client - it's given what it needs only...

BoxContentSDK/BoxContentSDK/Clients/BOXContentClient.m
boxtcarpel: Sounds good. Maybe still a comment in the accessors that the 'APIBaseURL' and 'APIUploadBaseURL' are expected to contain the API version number within them.

boxtcarpel (approved):

PR #290, opened 2016-12-23
Minor cleanup for recent item additions

adempsey:


boxcla:

BoxContentSDK/BoxContentSDK/Protocols/BOXRecentItemProtocol.h
```diff
@@ -39,6 +26,6 @@
  URL for the shared link that was used to access the item by the user.
  nil if not applicable
  */
-@property (nonatomic, readwrite, strong) NSURL *sharedLink;
+@property (nonatomic, readwrite, strong) NSURL *sharedLinkURL;
```
boxtcarpel: Should these be read/write in the protocol?

BoxContentSDK/BoxContentSDK/Models/BOXRecentFile.h
```diff
@@ -0,0 +1,23 @@
...
+
+#import <BoxContentSDK/BoxContentSDK.h>
+#import "BOXRecentItemProtocol.h"
+
+/**
+ Represents a file recently interacted with by the current user.
+ */
+@interface BOXRecentFile : BOXFile <BOXRecentItemProtocol>
+
+#pragma mark - BOXRecentItemProtocol
```
boxtcarpel: Don't need this unless we're overwriting read/write

BoxContentSDK/BoxContentSDK/Models/BOXRecentFile.h
```diff
@@ -0,0 +1,23 @@
...
+//
+//
+
+#import <BoxContentSDK/BoxContentSDK.h>
+#import "BOXRecentItemProtocol.h"
+
+/**
+ Represents a file recently interacted with by the current user.
+ */
+@interface BOXRecentFile : BOXFile <BOXRecentItemProtocol>
```
boxtcarpel: Why do we need this concrete class and a protocol? It would seem the class would be sufficient (wouldn't support bookmarks, but we already have that problem and the protocol only helps a little).

BoxContentSDK/BoxContentSDK/BOXContentSDKConstants.m
```diff
@@ -104,6 +104,7 @@
 BOXAPIItemType *const BOXAPIItemTypeCollaboration = @"collaboration";
 BOXAPIItemType *const BOXAPIItemTypeGroup = @"group";
 BOXAPIItemType *const BOXAPIItemTypeFileVersion = @"file_version";
+BOXAPIItemType *const BOXAPIItemTypeRecentItem = @"recent_item";
```
boxtcarpel, line 107: Still need this?

boxtcarpel (approved):

BoxContentSDK/BoxContentSDK/BOXContentSDKConstants.m
adempsey: Not directly right now, but may be worth keeping around for some purpose in the future.

PR #293, opened 2017-01-06
support nsurlsession in BOXAPIDataOperation (download request)

thuyBox:
make BOXFileThumbnailRequest working with nsurlsession


boxcla:

BoxContentSDK/BoxContentSDK/BOXNSURLSessionDelegate.m
```diff
@@ -20,8 +21,108 @@
  */
 
 
+@interface BOXNSURLSessionDelegate()
+
+//a map to associate a session with its BOXAPIOperation
+//during session/task's delegate callbacks, we call appropriate methods on operation
+@property (nonatomic, readwrite, strong) NSMutableDictionary *sessionIdToOperation;
```
jlawton: This can be `readonly`

BoxContentSDK/BoxContentSDK/BOXNSURLSessionDelegate.m
```diff
@@ -20,8 +21,108 @@
...
+{
+    if (_sessionIdToOperation == nil) {
+        _sessionIdToOperation = [NSMutableDictionary new];
+    }
+    return _sessionIdToOperation;
+}
+
+- (void)mapSessionTaskId:(NSUInteger)sessionTaskId withOperation:(BOXAPIOperation *)operation
+{
+    [self.sessionIdToOperation setObject:operation forKey:@(sessionTaskId)];
```
jlawton: Is this always called mutually exclusively to itself or any delegate callbacks? Or do we have to lock access to this dictionary?

Also, is there any reason we're preferring method syntax over subscripting for all these dictionary accesses?

BoxContentSDK/BoxContentSDK/BOXNSURLSessionManager.m
```diff
@@ -47,19 +48,19 @@ - (NSURLSession *)defaultSession
 
 - (NSURLSession *)backgroundSession
```
jlawton: `advancedSession`. Same for calls.

BoxContentSDK/BoxContentSDK/BOXNSURLSessionDelegate.m
jlawton: Could we add type annotations?

BoxContentSDK/BoxContentSDK/BOXNSURLSessionDelegate.m
```diff
@@ -20,8 +21,108 @@
...
+ * nil, which implies that no error occurred and this task is complete.
+ */
+- (void)URLSession:(NSURLSession *)session task:(NSURLSessionTask *)task
+didCompleteWithError:(nullable NSError *)error
+{
+    BOXAPIOperation *operation = [self.sessionIdToOperation objectForKey:@(task.taskIdentifier)];
+    if (operation != nil && [operation isKindOfClass:[BOXAPIDataOperation class]]) {
+        BOXAPIDataOperation *dataOperation = (BOXAPIDataOperation *)operation;
+        [dataOperation finishURLSessionTaskWithData:nil response:task.response error:error];
+    }
```
jlawton: Should we be removing from `sessionIdToOperation`?

BoxContentSDK/BoxContentSDK/Operations/BOXAPIDataOperation.h
```diff
@@ -62,9 +62,17 @@ typedef void (^BOXAPIDataProgressBlock)(long long expectedTotalBytes, unsigned l
  * @warning If you are manually reading from this output stream (for example with
  * a `CFStreamCreateBoundPair`) do not let data sit in the stream or you risk causing
  * a large file to buffer entirely in memory.
+ *
+ * If destinationPath is provided, outputStream will be ignored
```
jlawton, line 66: :( This is another case where Swift would make the interface nicer.

BoxContentSDK/BoxContentSDK/BOXNSURLSessionDelegate.m
```diff
@@ -20,8 +21,108 @@
...
+}
+
+#pragma mark - NSURLSessionDataDelegate
+
+- (void)URLSession:(NSURLSession *)session dataTask:(NSURLSessionDataTask *)dataTask
+didReceiveResponse:(NSURLResponse *)response
+ completionHandler:(void (^)(NSURLSessionResponseDisposition disposition))completionHandler
+{
+    BOXAPIOperation *operation = [self.sessionIdToOperation objectForKey:@(dataTask.taskIdentifier)];
+    if (operation != nil && [operation isKindOfClass:[BOXAPIDataOperation class]]) {
```
jlawton: :) Doing class checks here *really* means that a consumer would have trouble creating meaningful subclasses. Maybe we should be thinking about a protocol or something.

BoxContentSDK/BoxContentSDK/Operations/BOXAPIDataOperation.m
```diff
@@ -79,7 +79,20 @@ - (NSData *)encodeBody:(NSDictionary *)bodyDictionary
...
 - (BOOL)shouldUseSessionTask
 {
-    return NO;
+    return YES;
+}
+
+- (NSURLSessionTask *)createSessionTask
+{
+    NSURLSessionTask *sessionTask;
+    __weak BOXAPIDataOperation *weakSelf = self;
```
jlawton: Making this weak here won't make any difference.

BoxContentSDKSampleApp/BoxContentSDKSampleApp/BOXSampleThumbnailsHelper.m
```diff
@@ -41,6 +41,7 @@ - (instancetype)init
 
 - (void)storeThumbnailForItemWithID:(NSString *)itemID userID:(NSString *)userID thumbnail:(UIImage *)thumbnail
 {
+    if (thumbnail == nil) return;
```
jlawton: I'm surprised this is even called with `nil`. Is it intended to be? And if so, is it intended to be a noop?

thuyBox: yeah, it's possible to call this method with nil, calling with nil will cause a crash. This class is used in the sample app, and meant for caching thumbnails so we don't need to re-download. Made change to remove cache if nil

BoxContentSDK/BoxContentSDK/BOXNSURLSessionManager.m
thuyBox: done

BoxContentSDK/BoxContentSDK/BOXNSURLSessionDelegate.m
thuyBox: done

BoxContentSDK/BoxContentSDK/BOXNSURLSessionManager.m
```diff
@@ -7,27 +7,39 @@
...
-@property (nonatomic, readwrite, strong) NSURLSession *backgroundSession;
+//Advanced NSURLSession to be used by NSURLSessionTask which needs a delegate to handle specific tasks
+//during the life of a session/task, and to be run in the background e.g. download, upload
+@property (nonatomic, readwrite, strong) NSURLSession *advancedSession;
 
-//Delegate for background NSURLSession to handle its callbacks
-@property (nonatomic, readwrite, strong) BOXNSURLSessionDelegate *sessionDelegate;
+//a map to associate a session task with its task delegate
+//during session/task's delegate callbacks, we call appropriate methods on task delegate
+@property (nonatomic, readonly, strong) NSMapTable *sessionTaskIdToTaskDelegate;
```
jlawton: Can we specify the key and value types? Something like `NSMapTable<NSNumber *, id<BOXNSURLSessionTaskDelegate>> *`

BoxContentSDK/BoxContentSDK/BOXNSURLSessionManager.m
```diff
@@ -45,49 +57,145 @@ - (NSURLSession *)defaultSession
...
+ * still be called.
+ */
+- (void)URLSession:(NSURLSession *)session downloadTask:(NSURLSessionDownloadTask *)downloadTask
+didFinishDownloadingToURL:(NSURL *)location
+{
+    id<BOXNSURLSessionTaskDelegate> taskDelegate = [self.sessionTaskIdToTaskDelegate objectForKey:@(downloadTask.taskIdentifier)];
+    if ([taskDelegate conformsToProtocol:@protocol(BOXNSURLSessionDownloadTaskDelegate)]) {
+        id<BOXNSURLSessionDownloadTaskDelegate> downloadTaskDelegate = taskDelegate;
+
+        if ([downloadTaskDelegate respondsToSelector:@selector(destinationPath)]) {
```
jlawton: Should we also do a `nil` check on the destination path? This probably shouldn't happen, but you know, it's not guaranteed by the type. :)

BoxContentSDK/BoxContentSDK/BOXNSURLSessionManager.h
```diff
@@ -10,6 +10,57 @@
...
+/**
+ * The location for output file. If provided, outputStream will be ignored
+ * Using destinationPath to consume data will allow request to be executed in the background if the app is killed/suspended and resume upon app restarts/resumes
+ */
+- (NSString *)destinationPath;
+
+/**
+ * To be called to report ongoing progress of the task
+ */
+- (void)progressWithExpectedTotalBytes:(long long)expectedTotalBytes totalBytesReceived:(unsigned long long)totalBytesReceived;
```
jlawton: The order of received before expected makes more sense to me, like `progressWithTotalBytesRecieved:expected:`. Why is one of these signed, and one unsigned? I'm guessing you get a negative expected if we don't know? Should we be using `int64_t` to be compatible with NSURLSession?

BoxContentSDK/BoxContentSDK/Operations/BOXAPIDataOperation.m
```diff
@@ -156,6 +168,12 @@ - (void)performProgressCallback
...
 
+- (void)finish
+{
+    [super finish];
+    [self close];
+}
+
 - (void)dealloc
 {
     if (self.outputStream) {
```
jlawton, line 179: Odd that we're `nil`ling the stream delegate elsewhere, but not here.

BoxContentSDK/BoxContentSDK/BOXNSURLSessionManager.h
```diff
@@ -10,6 +10,57 @@
...
+
+@protocol BOXNSURLSessionDownloadTaskDelegate <BOXNSURLSessionTaskDelegate>
+
+@optional
+
+/**
+ * Output stream to write data into. If destinationPath is provided, outputStream will be ignored
+ * Using outputStream to consume data will not allow the request to be executed in the background if the app is killed/suspended
+ */
+- (NSOutputStream *)outputStream;
```
jlawton: Is this ever actually used through the protocol?

BoxContentSDK/BoxContentSDK/BOXNSURLSessionManager.m
```diff
@@ -45,49 +57,145 @@ - (NSURLSession *)defaultSession
...
+- (void)URLSession:(NSURLSession *)session dataTask:(NSURLSessionDataTask *)dataTask
+didReceiveResponse:(NSURLResponse *)response
+ completionHandler:(void (^)(NSURLSessionResponseDisposition disposition))completionHandler
+{
+    id<BOXNSURLSessionTaskDelegate> taskDelegate = [self.sessionTaskIdToTaskDelegate objectForKey:@(dataTask.taskIdentifier)];
+    if ([taskDelegate respondsToSelector:@selector(processIntermediateResponse:)]) {
+        [taskDelegate processIntermediateResponse:response];
+    }
+
+    if (completionHandler != nil) {
```
jlawton: `completionHandler` is defined as `nonnull`

BoxContentSDK/BoxContentSDK/Operations/BOXAPIDataOperation.m
```diff
@@ -224,10 +234,48 @@ - (void)connection:(NSURLConnection *)connection didReceiveData:(NSData *)data
         // If we received an error, don't write the response data to the output stream
         [super connection:connection didReceiveData:data];
     } else {
+        [self processIntermediateData:data];
+    }
+}
+
+
```
jlawton: Lots of space here.

thuyBox: done

BoxContentSDK/BoxContentSDK/BOXNSURLSessionManager.h
thuyBox: done

BoxContentSDK/BoxContentSDK/BOXNSURLSessionManager.h
thuyBox: yes, we will use it for download using data session task type

BoxContentSDK/BoxContentSDK/BOXNSURLSessionManager.m
thuyBox: done

BoxContentSDK/BoxContentSDK/BOXNSURLSessionManager.m
thuyBox: done

BoxContentSDK/BoxContentSDK/Operations/BOXAPIDataOperation.m
thuyBox: done

BoxContentSDK/BoxContentSDK/BOXNSURLSessionManager.h
```diff
@@ -92,6 +113,16 @@ NS_ASSUME_NONNULL_BEGIN
  */
 - (NSURLSessionUploadTask *)createUploadTask:(NSURLRequest *)request fromFile:(NSURL *)fileURL;
 
+/**
+ * Associate a session task with its task delegate to handle callbacks for it
+ */
+- (void)associateSessionTaskId:(NSUInteger)sessionTaskId withTaskDelegate:(id <BOXNSURLSessionTaskDelegate> )taskDelegate;
```
jlawton, line 114: May want to call out that the delegate is not retained.

thuyBox: done

BoxContentSDK/BoxContentSDK/BOXNSURLSessionManager.h
```diff
@@ -10,21 +10,98 @@
...
+ * Output stream to write data into. If destinationPath is provided, outputStream will be ignored
+ * Using outputStream to consume data will not allow the request to be executed in the background if the app is killed/suspended
+ */
+- (NSOutputStream *)outputStream;
+
+/**
+ * The location for output file. If provided, outputStream will be ignored
+ * Using destinationPath to consume data will allow request to be executed in the background if the app is killed/suspended and resume upon app restarts/resumes
+ */
+- (NSString *)destinationPath;
```
jlawton: Should document that a file will appear at the destination path on success, before finish… is called, and if not, there must have been an error. Also, this object is responsible for cleaning up the destination path in finish…

jlawton: What are the rules for this path? Do we expect the directory to exist, and the file not to? What happens if either of those are false?

jlawton: I ask, because there's no error passing mechanism for a failure here.

thuyBox: done

BoxContentSDK/BoxContentSDK/BOXNSURLSessionManager.h
```diff
@@ -10,21 +10,102 @@
...
+
+/**
+ * The location for output file. If provided, outputStream will be ignored
+ * Using destinationPath to consume data will allow request to be executed in the background
+ * if the app is killed/suspended and resume upon app restarts/resumes
+ * If download finishes successfully, i.e. at the time finishURLSessionTaskWithResponse:error:
+ *   method gets called, if no error, the downloaded file will be available at destinationPath
+ * The task delegate will be responsible for the downloaded file onwards
+ */
+- (NSString *)destinationPath;
```
jlawton: This should be `nullable`, I think. Same for `outputStream`

BoxContentSDK/BoxContentSDK/BOXNSURLSessionManager.h
```diff
@@ -10,21 +10,102 @@
...
+@optional
+
+/**
+ * Output stream to write data into. If destinationPath is provided, outputStream will be ignored
+ * Using outputStream to consume data will not allow the request to be executed in the background if the app is killed/suspended
+ */
+- (NSOutputStream *)outputStream;
+
+/**
+ * The location for output file. If provided, outputStream will be ignored
```
jlawton: How about something like this?
```
/**
 * The location into which to download the file. If provided and non-`nil`,
 * `outputStream` will be ignored.
 *
 * Downloads to a destination path can occur in the background, and managed by
 * the system even while the app is not running. The app may be launched in the
 * background to handle the completion of such a download.
 *
 * Upon successful download, when `-finishURLSessionTaskWithResponse:error:` is
 * called on this delegate, a file will already exist at the `downloadPath`.
 * The delegate is then considered the owner of the file.
 *
 * The path should be the name of a non-existent file in a writable directory.
 * If there was an error writing the file, that error would not be reflected in
 * `-finishURLSessionTaskWithResponse:error:`, so it must be inferred by the
 * missing file.
 */
```

BoxContentSDK/BoxContentSDK/BOXNSURLSessionManager.h
thuyBox: per your question about how to propage error if we fail to move the file in session manager, I think it's better to have an explicit method instead of destinationPath like - (void)didFinishDownloadingToURL:(NSURL *)location to notify about downloaded file exists and let the task delegate handle accordingly and handle error if any.

BoxContentSDK/BoxContentSDK/Operations/BOXAPIDataOperation.m
```diff
@@ -294,6 +294,17 @@ - (void)progressWithTotalBytesWritten:(int64_t)totalBytesWritten totalBytesExpec
     }
 }
 
+- (void)didFinishDownloadingToURL:(NSURL *)location
+{
+    if (self.destinationPath != nil) {
+        NSError *error = nil;
+        [[NSFileManager defaultManager] moveItemAtPath:location.path toPath:self.destinationPath error:&error];
+        if (self.error == nil) {
```
jlawton, line 302: Is there a case where `self.error != nil`, and this gets called? I guess I'm fine with this check, anyway.

jlawton (approved):

BoxContentSDK/BoxContentSDK/Operations/BOXAPIDataOperation.m
thuyBox: for this type of operation, we don't expect it to be. It's more of a safe check to honor existing error if it comes before this.

PR #302, opened 2017-01-20
add support for reconnecting with background download upon app restart

thuyBox:
integrate background download reconnection in sample app

### DIFF
--- a/BoxContentSDK/BoxContentSDK/BOXNSURLSessionManager.h
+++ b/BoxContentSDK/BoxContentSDK/BOXNSURLSessionManager.h
@@ -59,14 +59,21 @@ NS_ASSUME_NONNULL_BEGIN
 @protocol BOXNSURLSessionManagerDelegate <NSObject>
 
 /**
+ * Notify delegate about download progress
+ */
+- (void)downloadTask:(NSURLSessionDownloadTask *)downloadTask
+   totalBytesWritten:(int64_t)totalBytesWritten
+totalBytesExpectedToWrite:(int64_t)totalBytesExpectedToWrite;
+
+/**
  * Sent when a download session task has finish downloading into a url location
  */
-- (void)downloadTask:(NSUInteger)sessionTaskId didFinishDownloadingToURL:(NSURL *)location;
+- (void)downloadTask:(NSURLSessionTask *)sessionTask didFinishDownloadingToURL:(NSURL *)location;
 
 /**
  * Sent when a session task finishes
  */
-- (void)finishURLSessionTask:(NSUInteger)sessionTaskId withResponse:(NSURLResponse *)response error:(NSError *)error;
+- (void)finishURLSessionTask:(NSURLSessionTask *)sessionTask withResponse:(NSURLResponse *)response error:(NSError *)error;
 
 @end
 
@@ -76,11 +83,13 @@ NS_ASSUME_NONNULL_BEGIN
 @interface BOXNSURLSessionManager : NSObject
 
 /**
- * A delegate to handle callbacks from session tasks that do not have associated task delegates
- * This is possible if the background tasks were created outside of this BOXNSURLSessionManager (e.g. app restarts)
- * A task delegate can always be re-associated back with a session task by calling associateSessionTaskId:withTaskDelegate:
+ * This method needs to be called once to set up the manager to be ready to perform other public methods
+ * @param delegate  handle callbacks from session tasks that do not have associated task delegates
+ *                  possible if the background tasks were created outside of BOXNSURLSessionManager (e.g. app restarts)
+ *                  A task delegate can always be re-associated with a session task by calling
+ *                  associateSessionTaskId:withTaskDelegate:
  */
-@property (nonatomic, strong, readwrite) id<BOXNSURLSessionManagerDelegate> delegate;
+- (void)setUpWithDelegate:(id<BOXNSURLSessionManagerDelegate>)delegate;
 
 /**
  Create a NSURLSessionDataTask which does not need to be run in background,
--- a/BoxContentSDK/BoxContentSDK/BOXNSURLSessionManager.h
+++ b/BoxContentSDK/BoxContentSDK/BOXNSURLSessionManager.h
@@ -84,38 +84,39 @@ totalBytesExpectedToWrite:(int64_t)totalBytesExpectedToWrite;
 
 /**
  * This method needs to be called once to set up the manager to be ready to perform other public methods
- * @param delegate  handle callbacks from session tasks that do not have associated task delegates
- *                  possible if the background tasks were created outside of BOXNSURLSessionManager (e.g. app restarts)
- *                  A task delegate can always be re-associated with a session task by calling
- *                  associateSessionTaskId:withTaskDelegate:
+ * @param defaultDelegate   handle callbacks from session tasks that do not have associated task delegates
+ *                          possible if the background tasks were created outside of BOXNSURLSessionManager
+ *                          (e.g. app restarts)
+ *                          A task delegate can always be re-associated with a session task by calling
+ *                          associateSessionTaskId:withTaskDelegate:
  */
-- (void)setUpWithDelegate:(id<BOXNSURLSessionManagerDelegate>)delegate;
+- (void)setUpWithDefaultDelegate:(id<BOXNSURLSessionManagerDelegate>)defaultDelegate;
 
 /**
  Create a NSURLSessionDataTask which does not need to be run in background,
  and its completionHandler will be called upon completion of the task
  */
-- (NSURLSessionDataTask *)createDataTask:(NSURLRequest *)request completionHandler:(void (^)(NSData * data, NSURLResponse * response, NSError * error))completionHandler;
+- (NSURLSessionDataTask *)createDataTaskWithRequest:(NSURLRequest *)request completionHandler:(void (^)(NSData * data, NSURLResponse * response, NSError * error))completionHandler;
 
 /**
  Create a NSURLSessionDownloadTask which can be run in the background and download to an outputstream
  */
-- (NSURLSessionDownloadTask *)createDownloadTaskWithRequest:(NSURLRequest *)request taskDelegate:(id <BOXNSURLSessionDownloadTaskDelegate>)taskDelegate;
+- (NSURLSessionDownloadTask *)createBackgroundDownloadTaskWithRequest:(NSURLRequest *)request taskDelegate:(id <BOXNSURLSessionDownloadTaskDelegate>)taskDelegate;
 
 /**
  Create a NSURLSessionDataTask which can be run to download data into a destination path but not run in the background
  */
-- (NSURLSessionDataTask *)createDataTaskForDownload:(NSURLRequest *)request taskDelegate:(id <BOXNSURLSessionDownloadTaskDelegate>)taskDelegate;
+- (NSURLSessionDataTask *)createNonBackgroundDownloadTaskWithRequest:(NSURLRequest *)request taskDelegate:(id <BOXNSURLSessionDownloadTaskDelegate>)taskDelegate;
 
 /**
  Create a NSURLSessionDownloadTask to be resumed
  */
-- (NSURLSessionDownloadTask *)createDownloadTaskWithResumeData:(NSData *)resumeData;
+- (NSURLSessionDownloadTask *)createBackgroundDownloadTaskWithResumeData:(NSData *)resumeData;
 
 /**
  Create a NSURLSessionUploadTask which can be run in background
  */
-- (NSURLSessionUploadTask *)createUploadTask:(NSURLRequest *)request fromFile:(NSURL *)fileURL;
+- (NSURLSessionUploadTask *)createBackgroundUploadTaskWithRequest:(NSURLRequest *)request fromFile:(NSURL *)fileURL;
 
 /**
  * Associate a session task with its task delegate to handle callbacks for it, taskDelegate is not retained
--- a/BoxContentSDK/BoxContentSDK/BOXNSURLSessionManager.h
+++ b/BoxContentSDK/BoxContentSDK/BOXNSURLSessionManager.h
@@ -118,6 +118,11 @@ NS_ASSUME_NONNULL_BEGIN
  */
 - (void)dessociateSessionTaskId:(NSUInteger)sessionTaskId;
 
+/**
+ * Asynchronously calls a completion callback with all background upload, and download tasks in a session.
+ */
+- (void)pendingBackgroundDownloadUploadSessionTasks:(void (^)(NSArray<NSURLSessionUploadTask *> * _Nonnull uploadTasks, NSArray<NSURLSessionDownloadTask *> * _Nonnull downloadTasks))completion;
+
 @end
 
 NS_ASSUME_NONNULL_END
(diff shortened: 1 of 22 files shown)


boxcla:

BoxContentSDK/BoxContentSDK/BOXNSURLSessionManager.h
```diff
@@ -118,6 +118,11 @@ NS_ASSUME_NONNULL_BEGIN
  */
 - (void)dessociateSessionTaskId:(NSUInteger)sessionTaskId;
 
+/**
+ * Asynchronously calls a completion callback with all background upload, and download tasks in a session.
+ */
+- (void)pendingBackgroundDownloadUploadSessionTasks:(void (^)(NSArray<NSURLSessionUploadTask *> * _Nonnull uploadTasks, NSArray<NSURLSessionDownloadTask *> * _Nonnull downloadTasks))completion;
```
jlawton, line 134: I think inside `NS_ASSUME_NONNULL` you shouldn't need `_Nonnull` in block parameters, if you want to save some line length.

BoxContentSDK/BoxContentSDK/Operations/BOXAPIDataOperation.m
```diff
@@ -389,6 +398,7 @@ - (id)copyWithZone:(NSZone *)zone
     operationCopy.successBlock = [self.successBlock copy];
     operationCopy.failureBlock = [self.failureBlock copy];
     operationCopy.progressBlock = [self.progressBlock copy];
+    operationCopy.sessionTaskReplacedBlock = [self.sessionTaskReplacedBlock copy];
```
jlawton: These block properties should all be declared `copy` (as block properties always should be), and then you don't need the copy calls here, as the setter will do that.

BoxContentSDK/BoxContentSDK/Operations/BOXAPIJSONOperation.m
```diff
@@ -27,6 +27,7 @@ - (id)copyWithZone:(NSZone *)zone
     operationCopy.success = [self.success copy];
     operationCopy.failure = [self.failure copy];
     operationCopy.timesReenqueued = self.timesReenqueued;
+    operationCopy.sessionTaskReplacedBlock = [self.sessionTaskReplacedBlock copy];
```
jlawton: Block properties should be `copy`.

BoxContentSDK/BoxContentSDK/Operations/BOXAPIOperation.h
```diff
@@ -161,6 +162,8 @@ typedef void (^BOXAPIDataFailureBlock)(NSURLRequest *request, NSHTTPURLResponse
  */
 @property (nonatomic, readwrite, strong) NSError *error;
 
+@property (nonatomic, readwrite, strong) BOXSessionTaskReplacedBlock sessionTaskReplacedBlock;
```
jlawton: `copy`

BoxContentSDK/BoxContentSDK/Requests/BOXFileDownloadRequest.m
```diff
@@ -13,7 +13,8 @@ @interface BOXFileDownloadRequest ()
 @property (nonatomic, readonly, strong) NSString *destinationPath;
 @property (nonatomic, readonly, strong) NSOutputStream *outputStream;
 @property (nonatomic, readonly, strong) NSString *fileID;
-
+@property (nonatomic, readwrite, strong) NSURLSessionDownloadTask *downloadTask;
+@property (nonatomic, readwrite, strong) BOXSessionTaskReplacedBlock downloadTaskReplacedBlock;
```
jlawton: `copy`

BoxContentSDK/BoxContentSDK/BOXNSURLSessionManager.h
thuyBox: done

BoxContentSDK/BoxContentSDK/Operations/BOXAPIOperation.h
thuyBox: done

BoxContentSDK/BoxContentSDK/Operations/BOXAPIDataOperation.m
thuyBox: done

BoxContentSDK/BoxContentSDK/Operations/BOXAPIJSONOperation.m
thuyBox: done

BoxContentSDK/BoxContentSDK/BOXNSURLSessionManager.h
```diff
@@ -76,11 +83,13 @@ NS_ASSUME_NONNULL_BEGIN
...
- * This is possible if the background tasks were created outside of this BOXNSURLSessionManager (e.g. app restarts)
- * A task delegate can always be re-associated back with a session task by calling associateSessionTaskId:withTaskDelegate:
+ * This method needs to be called once to set up the manager to be ready to perform other public methods
+ * @param delegate  handle callbacks from session tasks that do not have associated task delegates
+ *                  possible if the background tasks were created outside of BOXNSURLSessionManager (e.g. app restarts)
+ *                  A task delegate can always be re-associated with a session task by calling
+ *                  associateSessionTaskId:withTaskDelegate:
  */
-@property (nonatomic, strong, readwrite) id<BOXNSURLSessionManagerDelegate> delegate;
+- (void)setUpWithDelegate:(id<BOXNSURLSessionManagerDelegate>)delegate;
```
boxtcarpel: Can the delegate change? Or does it have to be the same for the lifetime of the app?

BoxContentSDK/BoxContentSDK/BOXNSURLSessionManager.m
```diff
@@ -57,7 +62,7 @@ - (NSURLSession *)defaultSession
     return _defaultSession;
 }
 
-- (NSURLSession *)advancedSession
+- (NSURLSession *)createAdvancedSession
```
boxtcarpel: James and I had spoken on the naming here. Advanced session is kind of generic, could mean anything. If we don't like backgroundSession, what about longRunningSession or something else like that? Something that clarifies what it is for.

BoxContentSDK/BoxContentSDK/BOXNSURLSessionManager.m
```diff
@@ -19,6 +19,11 @@ @interface BOXNSURLSessionManager() <NSURLSessionDelegate, NSURLSessionTaskDeleg
 //during the life of a session/task, and to be run in the background e.g. download, upload
 @property (nonatomic, readwrite, strong) NSURLSession *advancedSession;
 
+//A delegate to handle callbacks from session tasks that do not have associated task delegates
+//This is possible if the background tasks were created outside of this BOXNSURLSessionManager (e.g. app restarts)
+//A task delegate can always be re-associated back with a session task by calling associateSessionTaskId:withTaskDelegate:
+@property (nonatomic, strong, readwrite) id<BOXNSURLSessionManagerDelegate> delegate;
```
boxtcarpel: Minor, naming. I wonder if we should rename this delegate and protocol to be a bit more clear. This isn't a regular delegate where these callbacks always happen on these events. It's some sort of fallback delegate where these callbacks only happen when there isn't some other more specific listener listening...

BoxContentSDK/BoxContentSDK/Clients/BOXContentClient+File.h
```diff
@@ -8,6 +8,7 @@
 
 #import "BOXContentSDKConstants.h"
 #import "BOXContentClient.h"
+#import "BOXAPIDataOperation.h"
```
boxtcarpel: Do we need to know about "BOXAPIDataOperation" here? Maybe just "BOXAPIOperation" for the defenition of BOXSessionTaskReplacedBlock ?

BoxContentSDK/BoxContentSDK/Clients/BOXContentClient+File.h
```diff
@@ -134,6 +135,23 @@
                                                     fromLocalFilePath:(NSString *)localFilePath;
 
 /**
+ *  Generate a request to upload a new version of a file from a local file.
```
boxtcarpel: Incorrect comment

BoxContentSDK/BoxContentSDK/Clients/BOXContentClient+File.m
```diff
@@ -20,6 +20,8 @@
 #import "BOXPreflightCheckRequest.h"
 #import "BOXFileUploadNewVersionRequest.h"
 
+#import "BOXContentClient+File.h"
```
boxtcarpel: That's already here... at the top

BoxContentSDK/BoxContentSDK/Operations/BOXAPIDataOperation.h
```diff
@@ -130,6 +130,8 @@ typedef void (^BOXAPIDataProgressBlock)(long long expectedTotalBytes, unsigned l
  */
 @property (nonatomic, readwrite, strong) NSString *modelID;
 
+- (id)initWithURL:(NSURL *)URL HTTPMethod:(NSString *)HTTPMethod body:(NSDictionary *)body queryParams:(NSDictionary *)queryParams session:(BOXAbstractSession *)session sessionTask:(NSURLSessionTask *)sessionTask;
```
boxtcarpel: minor: colon align for readability

boxtcarpel: Also, is this only viable for data operations?
Also, having session and sessionTask next to one another is confusing because they refer to different kinds of sessions. maybe ...:session:URLSessionTask: ?

BoxContentSDK/BoxContentSDK/Operations/BOXAPIOperation.m
```diff
@@ -348,6 +350,16 @@ - (void)executeOperation
     }
 }
 
+- (void)setSessionTask:(NSURLSessionTask *)sessionTask
+{
+    NSURLSessionTask *oldSessionTask = _sessionTask;
+    _sessionTask = sessionTask;
+    if (_sessionTaskReplacedBlock != nil) {
+        _sessionTaskReplacedBlock(oldSessionTask, _sessionTask);
```
boxtcarpel, line 358: do you want to call this if oldSessionTask was nil, or only if it was non-nil?

BoxContentSDK/BoxContentSDK/Operations/BOXAPIOperation.m
```diff
@@ -527,8 +539,12 @@ - (void)processResponse:(NSURLResponse *)response
 
 - (void)finishURLSessionTaskWithData:(NSData *)data response:(NSURLResponse *)response error:(NSError *)error
 {
-    [self processResponse:response];
-    [self processResponseData:data];
+    if (response != nil) {
+        [self processResponse:response];
```
boxtcarpel: why are these not safe to call with nil?

BoxContentSDK/BoxContentSDK/Requests/BOXFileDownloadRequest.h
```diff
@@ -4,7 +4,7 @@
 //
 
 #import "BOXRequestWithSharedLinkHeader.h"
-
+#import "BOXAPIOperation.h"
 @interface BOXFileDownloadRequest : BOXRequestWithSharedLinkHeader
```
boxtcarpel: minor: should have an empty line here

BoxContentSDK/BoxContentSDK/BOXNSURLSessionManager.h
thuyBox: delegate is a one-time set for nsurlsessions. As long as you use the same BOXNSURLSessionManager, the delegate is the same.

BoxContentSDK/BoxContentSDK/BOXNSURLSessionManager.m
thuyBox: done

BoxContentSDK/BoxContentSDK/Clients/BOXContentClient+File.h
thuyBox: done

BoxContentSDK/BoxContentSDK/Clients/BOXContentClient+File.h
thuyBox: done

BoxContentSDK/BoxContentSDK/Clients/BOXContentClient+File.m
thuyBox: done

BoxContentSDK/BoxContentSDK/Operations/BOXAPIDataOperation.h
thuyBox: done

BoxContentSDK/BoxContentSDK/Operations/BOXAPIOperation.m
thuyBox: we should notify whenever there is new sessionTask regardless of oldSessionTask being nil or not. We could check if oldSessionTask != _sessionTask but not sure if it's really needed.

BoxContentSDK/BoxContentSDK/Requests/BOXFileDownloadRequest.h
thuyBox: done

BoxContentSDK/BoxContentSDK/Operations/BOXAPIOperation.m
thuyBox: I don't recall the exact reason, given that we might have a valid intermediate HTTPResponse before this finish call, we might have wanted to keep the previous response if the last response is nil. But we can always bring it back if needed. Removed.

BoxContentSDK/BoxContentSDK/BOXNSURLSessionManager.m
thuyBox: renamed to defaultDelegate

BoxContentSDK/BoxContentSDK/Operations/BOXAPIMultipartToJSONOperation.m
```diff
@@ -7,6 +7,7 @@
 //
 
 #import "BOXAPIMultipartToJSONOperation.h"
+#import "BOXAPIOperation_Private.h"
```
boxtcarpel, line 10: Is this needed? Doesn't appear to be

boxtcarpel (approved):

BoxContentSDK/BoxContentSDK/Operations/BOXAPIMultipartToJSONOperation.m
thuyBox: for sessionTask to be visible

boxtcarpel: I was looking at it on the master branch, not the nsurlsession branch...

PR #303, opened 2017-01-20
Parse recent item interaction dates from ISO-8601 format

adempsey:


boxcla:

jdizoglio (approved):

PR #313, opened 2017-02-01
Incorrect object name called in Sample App

Roshanalexander:
BOXFileDownloadRequest is instantiated in download file section with the name boxRequest. But it's called as 'request' where it should be 'boxRequest'.


boxcla:
Verified that @Roshanalexander has signed the CLA. Thanks for the pull request!

boxtcarpel:
Thanks, @Roshanalexander! It is definitely important for our documentation to be correct and up to date!

PR #316, opened 2017-02-06
Use synchronized BOOL instead of dispatch_once

### LINKED ISSUE #315: dispatch_once_t cannot be an instance variable
BOXAPIMultipartToJSONOperation.m uses dispatch_once_t as an instance variable which does not work.

https://github.com/box/box-ios-sdk/blob/master/BoxContentSDK/BoxContentSDK/Operations/BOXAPIMultipartToJSONOperation.m#L368

This StackOverflow question is a surprisingly good an thorough analysis why this doesn't work:
http://stackoverflow.com/questions/13856037/can-i-declare-dispatch-once-t-predicate-as-a-member-variable-instead-of-static/19845164#19845164

Greg Parker, who works on the Objective-C Runtime at Apple.

>dispatch_once_t must not be an instance variable. The implementation of dispatch_once() requires that the dispatch_once_t is zero, and has never been non-zero. The previously-not-zero case would need additional memory barriers to work correctly, but dispatch_once() omits those barriers for performance reasons.

Xcode 8.3 Beta 1 now has a new Analyzer Warning step which we found while working on Box integration for PSPDFKit's PDF Viewer for iOS: https://pdfviewer.io

screen shot 2017-02-06 at 22 53 52

I'll follow up with a PR that fixes this.

steipete:
Fixes #315.

CLA is signed.

### DIFF
--- a/BoxContentSDK/BoxContentSDK/Operations/BOXAPIMultipartToJSONOperation.m
+++ b/BoxContentSDK/BoxContentSDK/Operations/BOXAPIMultipartToJSONOperation.m
@@ -365,7 +365,7 @@ static NSString * BOXAPIMultipartContentTypeHeader(void)
 
 @interface BOXAPIMultipartToJSONOperation ()
 {
-    dispatch_once_t _pred;
+    BOOL _streamsInitialized;
 }
 
 @property (nonatomic, readwrite, strong) NSMutableArray *formPieces;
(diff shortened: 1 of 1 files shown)


boxcla:
Verified that @steipete has signed the CLA. Thanks for the pull request!

boxtcarpel (approved):

PR #317, opened 2017-02-09
Update NSURLConnection to NSURLSession in the auth flow code

boxtcarpel:


boxcla:

BoxContentSDK/BoxContentSDK/OAuth2/BOXAuthorizationViewController.m
```diff
@@ -484,12 +461,38 @@ - (void)webViewDidFinishLoad:(UIWebView *)webView
...
+didReceiveResponse:(nonnull NSURLResponse *)response
+ completionHandler:(nonnull void (^)(NSURLSessionResponseDisposition))completionHandler
 {
-    BOXLog(@"connection %@ did receive authentication challenge %@", connection, challenge);
+    BOXLog(@"URLSessionDataTask %@ did receive response %@", dataTask, response);
+    if ([response isKindOfClass:[NSHTTPURLResponse class]]) {
+        BOXLog(@"HTTP Headers were: %@", [(NSHTTPURLResponse *)response allHeaderFields]);
+    }
+    self.connectionResponse = response;
+    [self.connectionData setLength:0];
```
thuyBox, line 476: should we only reset the connectionData after loading it to the web, i.e. inside - (void)URLSession:(NSURLSession *)session task:(NSURLSessionTask *)task didCompleteWithError:(NSError *)error?

BoxContentSDK/BoxContentSDK/OAuth2/BOXAuthorizationViewController.m
```diff
@@ -103,6 +100,11 @@ - (instancetype)initWithSDKClient:(BOXContentClient *)SDKClient
         _connectionData = [[NSMutableData alloc] init];
         _hostsThatCanUseWebViewDirectly = [NSMutableSet set];
 
+        NSURLSessionConfiguration *sessionConfiguration = [NSURLSessionConfiguration ephemeralSessionConfiguration];
+        _URLSession = [NSURLSession sessionWithConfiguration:sessionConfiguration
+                                                    delegate:self
+                                               delegateQueue:nil];
```
thuyBox, line 106: minor: create a delegateQueue ourselves to have control over maxConcurrentOperationCount to 1.

BoxContentSDK/BoxContentSDK/OAuth2/BOXAuthorizationViewController.m
```diff
@@ -504,45 +507,86 @@ - (void)connection:(NSURLConnection *)connection willSendRequestForAuthenticatio
         // Allow a certificate if its status was evaluated successfully and the result is that it should be trusted
         BOOL shouldTrustServer = (status == errSecSuccess && (trustResult == kSecTrustResultProceed || trustResult == kSecTrustResultUnspecified));
 
-        [self completeServerTrustAuthenticationChallenge:challenge shouldTrust:shouldTrustServer];
+        if (shouldTrustServer) {
+            SecTrustRef serverTrust = [[challenge protectionSpace] serverTrust];
+            NSURLCredential *serverTrustCredential = [NSURLCredential credentialForTrust:serverTrust];
+            if (completionHandler) {
+                completionHandler(NSURLSessionAuthChallengeUseCredential, serverTrustCredential);
+                //FIXME: Or should we use NSURLSessionAuthChallengePerformDefaultHandling
```
thuyBox: what needs to be done to remove this FIXME? from the documentation, with NSURLSessionAuthChallengePerformDefaultHandling, the provided credential parameter is ignored, which we don't want?

BoxContentSDK/BoxContentSDK/OAuth2/BOXAuthorizationViewController.m
```diff
@@ -504,45 +507,86 @@ - (void)connection:(NSURLConnection *)connection willSendRequestForAuthenticatio
...
+            } else {
+                [[challenge sender] useCredential:serverTrustCredential
```
thuyBox: could we test assuming that completionHandler is nil, and just call this directly to see if session task can still finish successfully?

BoxContentSDK/BoxContentSDK/OAuth2/BOXAuthorizationViewController.m
```diff
@@ -504,45 +507,86 @@ - (void)connection:(NSURLConnection *)connection willSendRequestForAuthenticatio
...
                                              (self.ntlmAuthFailures < kMaxNTLMAuthFailuresPriorToExit));
 
             // Handle the authentication challenge as a basic HTTP authentication challenge
             // (certificate-based, among other methods, is not currently supported).
             if (shouldApplyiOS8_NTLM_WAR) {
                 BOXLog(@"Applying iOS8 NTLM WAR, ntlmAuthFaulures = %ld", (long)self.ntlmAuthFailures);
                 self.ntlmAuthFailures++;
-                [[challenge sender] continueWithoutCredentialForAuthenticationChallenge:challenge];
+                if (completionHandler) {
+                    completionHandler(NSURLSessionAuthChallengeUseCredential, NULL);
```
thuyBox, line 573: should this be NSURLSessionAuthChallengePerformDefaultHandling which will ignore the provided credential parameter automatically? is there an NTLM test env we can test with?

BoxContentSDK/BoxContentSDK/OAuth2/BOXAuthorizationViewController.m
```diff
@@ -504,45 +507,86 @@ - (void)connection:(NSURLConnection *)connection willSendRequestForAuthenticatio
...
+                } else {
+                    [[challenge sender] continueWithoutCredentialForAuthenticationChallenge:challenge];
+                }
             } else if (([challenge previousFailureCount] > 0 &&
                         self.authenticationChallengeCredential == nil) ||
                        [challenge previousFailureCount] > 1) {
                 BOXLog(@"Have %ld previous failures", (long)[challenge previousFailureCount]);
+                if (completionHandler) {
+                    completionHandler(NSURLSessionAuthChallengeCancelAuthenticationChallenge, NULL);
```
thuyBox, line 582: not sure what's the impact of calling completionHandler earlier in the flow, should we call it at the very end, i.e. after calling failAuthenticationWithURLSessionTask?

BoxContentSDK/BoxContentSDK/OAuth2/BOXAuthorizationViewController.m
```diff
@@ -274,7 +250,7 @@ - (void)failAuthenticationWithConnection:(NSURLConnection *)connection
             self.completionBlock(self, nil, self.connectionError);
         }
     }
-    
+
```
thuyBox, line 253: should we call the self.completionBlock only when sessionTask finishes in the callback - (void)URLSession:(NSURLSession *)session task:(NSURLSessionTask *)task didCompleteWithError:(NSError *)error?

if given our contract defines completionBlock as the very last step to be called processing user's request. We don't know what might be a side effect of executing completionBlock's logic to the remaining life cycle of session task.

BoxContentSDK/BoxContentSDK/OAuth2/BOXAuthorizationViewController.m
```diff
@@ -552,62 +596,77 @@ - (void)connection:(NSURLConnection *)connection willSendRequestForAuthenticatio
...
+                                                                                     [[self.authenticationChallenge sender] cancelAuthenticationChallenge:self.authenticationChallenge];
+                                                                                 }
+                                                                                 [alertController dismissViewControllerAnimated:YES completion:nil];
+                                                                             }];
                         UIAlertAction *submitAction = [UIAlertAction actionWithTitle:NSLocalizedString(@"Submit", @"Alert view button: submit button for authentication challenge")
                                                                                style:UIAlertActionStyleDefault
                                                                              handler:^(UIAlertAction *action) {
                                                                                  [alertController dismissViewControllerAnimated:YES completion:nil];
-                                                                                 
+
```
thuyBox, line 636: minor: not sure if calling this too early could prevent the rest from execution, or should we call [alertController dismissViewControllerAnimated:YES completion:nil]; at the end of the block.

BoxContentSDK/BoxContentSDK/OAuth2/BOXAuthorizationViewController.m
```diff
@@ -616,52 +675,27 @@ - (void)connection:(NSURLConnection *)connection willSendRequestForAuthenticatio
...
+                                MIMEType:[self.connectionResponse MIMEType]
+                        textEncodingName:[self.connectionResponse textEncodingName]
+                                 baseURL:[self.connectionResponse URL]];
+
+        self.connectionResponse = nil;
+    } else {
+        // Failure
+        BOXLog(@"URLSessionTask %@ did fail with error %@", task, error);
+        if ([error code] != NSURLErrorUserCancelledAuthentication) {
+            ///
```
thuyBox: remove empty comment

BoxContentSDK/BoxContentSDK/OAuth2/BOXAuthorizationViewController.m
```diff
@@ -616,52 +675,27 @@ - (void)connection:(NSURLConnection *)connection willSendRequestForAuthenticatio
...
+            [self failAuthenticationWithURLSessionTask:task message:[error localizedDescription] error:error];
```
thuyBox: given previous comment inside failAuthenticationWithURLSessionTask method, we might need to call self.completionBlock after this.

BoxContentSDK/BoxContentSDK/OAuth2/BOXAuthorizationViewController.m
```diff
@@ -616,52 +675,27 @@ - (void)connection:(NSURLConnection *)connection willSendRequestForAuthenticatio
...
+        // Success
+        BOXLog(@"URLSessionTask %@ did finish loading. Requesting that the webview load the data (%lu bytes) with reponse %@", task, (unsigned long)[self.connectionData length], self.connectionResponse);
+        [self setWebViewCanBeUsedDirectly:YES forHost:task.currentRequest.URL.host];
+        [self setWebViewCanBeUsedDirectly:YES forHost:[self.connectionResponse URL].host];
+        [(UIWebView *)self.view loadData:self.connectionData
+                                MIMEType:[self.connectionResponse MIMEType]
+                        textEncodingName:[self.connectionResponse textEncodingName]
+                                 baseURL:[self.connectionResponse URL]];
+
+        self.connectionResponse = nil;
```
thuyBox, line 689: should we call self.completionBlock?

BoxContentSDK/BoxContentSDK/OAuth2/BOXAuthorizationViewController.m
boxtcarpel: Do you think we should do this? I think it's unnecessary...

BoxContentSDK/BoxContentSDK/OAuth2/BOXAuthorizationViewController.m
boxtcarpel: We should discuss offline, as I'm not sure exactly what you mean. The completion block here is only called at the last step as-is, on handling the failure... That in turn either gets called from didCompleteWithError (with an error that is not 'canceled') or in other scenarios where we want to handle the failure and then trigger a cancel (see the challenge handling).

BoxContentSDK/BoxContentSDK/OAuth2/BOXAuthorizationViewController.m
boxtcarpel: We need to reset it here because on receiving a response a new stream of data might be coming in.

BoxContentSDK/BoxContentSDK/OAuth2/BOXAuthorizationViewController.m
boxtcarpel: Well, I'm not sure it would be a valid test, because the system will still have a completion block with additional code it might be expected to run.

As long as the system passes us a completion block, it's not valid to not call it (as per the documentation).

We expect it will always have a completion block, but in case it doesn't, we know that the challenge can be responded to in this way.

BoxContentSDK/BoxContentSDK/OAuth2/BOXAuthorizationViewController.m
boxtcarpel: This was just for me to remember during development. I think it might work either way (the default handling ignores the credential, because it doesn't need it, it just does the default action).

But what I ended up going with is the closest analog to what we did before. I will remove the fixme.

BoxContentSDK/BoxContentSDK/OAuth2/BOXAuthorizationViewController.m
boxtcarpel: PerformDefaultHandling's function isn't to ignore the credential, it is to do the default action for that kind of challenge - which doesn't require a credential, which is why it ignores that parameter.
Here we want to continue without credential, which is a different thing (from docs: Use the specified credential, which may be nil.)

BoxContentSDK/BoxContentSDK/OAuth2/BOXAuthorizationViewController.m
boxtcarpel: We're just triggering a cancel, then handling the error on our end instead.

BoxContentSDK/BoxContentSDK/OAuth2/BOXAuthorizationViewController.m
boxtcarpel: It does not prevent the execution of the rest of the block.

BoxContentSDK/BoxContentSDK/OAuth2/BOXAuthorizationViewController.m
boxtcarpel: No, the auth flow has not yet completed, we're continuing the flow by loading the currently fetched data into the web view.

The completion block is only called at the end of the auth flow (due to success or errors).

BoxContentSDK/BoxContentSDK/OAuth2/BOXAuthorizationViewController.m
boxtcarpel: done

BoxContentSDK/BoxContentSDK/OAuth2/BOXAuthorizationViewController.m
boxtcarpel: Not sure what you mean, that is all handled in the failAuthentication method. The completion block gets called when it needs to be.

BoxContentSDK/BoxContentSDK/OAuth2/BOXAuthorizationViewController.m
thuyBox: should we continue appending data since we only process all the data together once at the end? Is there expectation that only the last data received should be processed?

boxtcarpel: Yeah, that is the expectation. I think we should keep that as-is for now since it follows the previous pattern which is working.

thuyBox:
+2 👍 👍

PR #322, opened 2017-02-14
fix sample app's build issues

thuyBox:


boxcla:

BoxContentSDK/BoxContentSDK/Operations/BOXAPIOperation.h
```diff
@@ -7,11 +7,9 @@
 //
 
 #import <Foundation/Foundation.h>
-
-#import "BOXOAuth2Session.h"
-
 #import "BOXContentSDKConstants.h"
 #import "BOXURLSessionManager.h"
+#import "BOXAbstractSession.h"
```
boxtcarpel: Can just forward declare the class, no? (actual import can be in the .m).

boxtcarpel: Which is covered by the subclass 'BOXOAuth2Session' which is already imported.

boxtcarpel (approved):

BoxContentSDK/BoxContentSDK/Operations/BOXAPIOperation.h
thuyBox: done

boxtcarpel (approved):

PR #325, opened 2017-02-16
Resolve issue with auth failing in the sample app

boxtcarpel:
The problem stemmed from using the default client instead of the instance
of BOXContentClient currently in-use by the sample app (which supports
multiple accounts and thus client objects).
Also fixes a problem with the configuration of the BOXURLSessionManager
repeatedly for each instance of BOXContentClient


boxcla:

thuyBox:
+2

PR #328, opened 2017-02-17
Resolve an issue with network calls not working in the sample app

boxtcarpel:
Make sure the content client object obtained by the sample app has
initialized its URLSessionManager correctly.

This will likely go away as we move to a centralized session manager.


boxcla:

thuyBox:
+2

PR #332, opened 2017-02-24
make BOXURLSessionManager a sharedInstance

thuyBox:


boxcla:

BoxContentSDK/BoxContentSDK/BOXContentSDK.h
```diff
@@ -35,7 +35,6 @@
 #import "BOXOAuth2Session.h"
 #import "BOXParallelOAuth2Session.h"
 #import "BOXAppUserSession.h"
-#import "BOXURLSessionManager.h"
```
boxtcarpel: What if one wants to extend the SDK with new functionality? Might knowledge the session manager be needed?

BoxContentSDK/BoxContentSDK/BOXURLSessionManager.h
```diff
@@ -81,14 +81,22 @@ totalBytesExpectedToSend:(int64_t)totalBytesExpectedToSend;
...
 /**
- * This method needs to be called once to set up the manager to be ready to perform other public methods
+ * Should only use sharedInstance instead of creating a new instance of BOXURLSessionManager
+ * BOXURLSessionManager is responsible for a unique background NSURLSession for the app
+ * with BOXURLSessionManager itself as delegate
+ */
++ (BOXURLSessionManager *)sharedInstance;
+
+/**
+ * This method needs to be called once to set up the manager to be ready to support background upload/download tasks.
```
boxtcarpel: Should we support calling this more than once? Can the delegate theoretically change?

BoxContentSDKSampleApp/BoxContentSDKSampleApp/BOXSampleAccountsViewController.m
```diff
@@ -40,7 +40,7 @@ - (void)viewDidLoad {
         BOXContentClient *client = [BOXContentClient clientForUser:user];
         if (   ([client.session isKindOfClass:[BOXOAuth2Session class]] && !self.isAppUsers)
             || ([client.session isKindOfClass:[BOXAppUserSession class]] && self.isAppUsers)) {
-            [client.urlSessionManager setUpWithDefaultDelegate:((id<BOXURLSessionManagerDelegate>)[[UIApplication sharedApplication] delegate])];
+            [client setUpToSupportBackgroundTasksWithDefaultDelegate:((id<BOXURLSessionManagerDelegate>)[[UIApplication sharedApplication] delegate])];
```
boxtcarpel, line 43: We might need to do this logic a bit differently if it is only expected to be called once (it's currently being called here for a new client, or below for retrieving an existing client - both of those things can happen a bunch of times in the life of the sample app).

BoxContentSDK/BoxContentSDK/BOXURLSessionManager.h
thuyBox: agree, this could be called more than once. done

BoxContentSDK/BoxContentSDK/BOXContentSDK.h
thuyBox: leave it public. done

BoxContentSDKSampleApp/BoxContentSDKSampleApp/BOXSampleAccountsViewController.m
thuyBox: remove the need to be called once

BoxContentSDK/BoxContentSDK/BOXURLSessionManager.m
```diff
@@ -106,17 +114,10 @@ - (NSURLSession *)createBackgroundSession
...
-        if (self.hasCompletedSetup == NO) {
-            self.defaultDelegate = delegate;
-            self.defaultSession = [self createDefaultSession];
-            self.progressSession = [self createProgressSession];
-            self.backgroundSession = [self createBackgroundSession];
-            self.hasCompletedSetup = YES;
-        }
-    }
+    self.defaultDelegate = delegate;
+    [self createBackgroundSession];
```
boxtcarpel, line 120: is this safe to call repeatedly or should we protect just this part?

boxtcarpel (approved):

BoxContentSDK/BoxContentSDK/BOXURLSessionManager.m
thuyBox: it's safe since we check for _backgroundSession == nil first, and NSURLSession will always give us the same instance of that background session of a given id after it was created

PR #333, opened 2017-03-02
Fixes for data-handling in NSURLSession implemenation especially re errors

boxtcarpel:


boxcla:

BoxContentSDK/BoxContentSDK/Operations/BOXAPIOperation.m
```diff
@@ -555,7 +555,17 @@ - (void)finishURLSessionTaskWithData:(NSData *)data response:(NSURLResponse *)re
 
 - (void)sessionTask:(NSURLSessionTask *)sessionTask didFinishWithResponse:(NSURLResponse *)response error:(NSError *)error
 {
-    [self finishURLSessionTaskWithData:nil response:response error:error];
+    [self finishURLSessionTaskWithData:self.responseData response:response error:error];
```
boxtcarpel, line 558: Note these pretty significant behavior changes here.

BoxContentSDK/BoxContentSDK/Operations/BOXAPIOperation.m
```diff
@@ -555,7 +555,17 @@ - (void)finishURLSessionTaskWithData:(NSData *)data response:(NSURLResponse *)re
...
+}
+
+- (void)processIntermediateResponse:(NSURLResponse *)response
```
thuyBox, line 561: I think we could remove this delegate method processIntermediateResponse completely together with BOXURLSessionManager's implementation of delegate method URLSession:dataTask:didReceiveResponse:completionHandler: given that we have no need to intercept this method for the ability that it provides, and we still have access to response at the last callback which exists for both background/non-background URLSession:task:didCompleteWithError:, this also helps avoid processing response twice.

boxtcarpel: Sometimes you need the response first... We might need to solve this another way

thuyBox:
+2, we'll follow-up with avoiding processing response twice for fore-ground download/upload

PR #334, opened 2017-03-02
Add missing response processing in stream operation

boxtcarpel:


boxcla:

BoxContentSDK/BoxContentSDK/Operations/BOXStreamOperation.m
```diff
@@ -144,12 +144,12 @@ - (void)abortWithError:(NSError *)error
...
+- (void)processIntermediateResponse:(NSURLResponse *)response
 {
-    [super sessionTask:sessionTask didFinishWithResponse:response error:error];
-
-    if (error != nil &&
-        error.code == BOXContentSDKAPIErrorAccepted) {
+    [super processIntermediateResponse:response];
+    
+    if (self.error != nil &&
+        self.error.code == BOXContentSDKAPIErrorAccepted) {
```
thuyBox, line 161: if we decide to remove processIntermediateResponse per https://github.com/box/box-ios-sdk/pull/333#pullrequestreview-24827686, we might want to keep it as before

thuyBox: BOXStreamOperation uses a sessionTask which takes in a completionHandler (and no other interaction needed). It inherits the same createSessionTask method from BOXAPIOperation.

- (NSURLSessionTask *)createSessionTask
{
    __weak BOXAPIOperation *weakSelf = self;
    NSURLSessionTask *sessionTask = [self.session.urlSessionManager createDataTaskWithRequest:self.APIRequest
                                                completionHandler:^(NSData *data, NSURLResponse *response, NSError *error) {
                                                    [weakSelf finishURLSessionTaskWithData:data response:response error:error];
                                                }];
    return sessionTask;
}

We should call sessionTask:didFinishWithResponse:error: inside sessionTask's completionHandler, so that subclasses like BOXStreamOperation can override.

thuyBox: if we want BOXStreamOperation to behave more like fore-ground download to report on progress, we want similar setup as BOXAIDataOperation. We can override createSessionTask to call [self.session.urlSessionManager createNonBackgroundDownloadTaskWithRequest:self.APIRequest taskDelegate:self];
and BOXStreamOperation inherits BOXURLSessionDownloadTaskDelegate

BoxContentSDK/BoxContentSDK/Operations/BOXStreamOperation.m
```diff
@@ -15,7 +15,7 @@
 #define MAX_REENQUE_DELAY 60
 #define REENQUE_BASE_DELAY 0.2
 
-@interface BOXStreamOperation ()
+@interface BOXStreamOperation () <BOXURLSessionDownloadTaskDelegate>
```
boxtcarpel, line 18: Need to explore if the task used should be NSURLSessionStreamTask

BoxContentSDK/BoxContentSDK/Operations/BOXAPIOperation.m
```diff
@@ -540,7 +540,9 @@ - (void)processResponse:(NSURLResponse *)response
 
 - (void)finishURLSessionTaskWithData:(NSData *)data response:(NSURLResponse *)response error:(NSError *)error
 {
-    [self processResponse:response];
+    if (self.HTTPResponse == nil) {
```
thuyBox, line 543: could we add a comment why we need to have this check? I assume around duplicated processing of the response?

thuyBox:
@irvingruan fyi

BoxContentSDK/BoxContentSDK/Operations/BOXAPIOperation.m
boxtcarpel: Right, if the response was already handled, then we don't want to handle it at "finish" time, otherwise it may need to still be handled. I can add a comment.

thuyBox:
+2

PR #336, opened 2017-03-07
Fix various issues with uploads

boxtcarpel:


boxcla:

BoxContentSDK/BoxContentSDK/Operations/BOXAPIOperation.m
```diff
@@ -540,13 +540,15 @@ - (void)processResponse:(NSURLResponse *)response
 
 - (void)finishURLSessionTaskWithData:(NSData *)data response:(NSURLResponse *)response error:(NSError *)error
 {
-    if (self.HTTPResponse == nil) {
+    if (self.HTTPResponse == nil && response != nil) {
```
thuyBox, line 550: what is the issue if we call processResponse with a nil response?

BoxContentSDK/BoxContentSDK/Operations/BOXAPIOperation.m
```diff
@@ -540,13 +540,15 @@ - (void)processResponse:(NSURLResponse *)response
...
         // If the response has not already been handled (some, but not all, operations get
-        // 'processResponse:' earlier in the process) handle it here.
+        // 'processResponse:' earlier in the process), handle it here.
+        // Also, do not handle a nil response. That could happen if there was a failure on the whole
+        // network connection, for example. In that case, there is no response.
         [self processResponse:response];
     }
 
-    if (data != nil) {
+    if ([data length] > 0) {
```
thuyBox: could you add a comment why we have to check for non-empty data?

would there be a legitimate case that data is empty and needs to be processed accordingly inside processResponseData, e.g. to set error accordingly?

BoxContentSDK/BoxContentSDK/Operations/BOXAPIMultipartToJSONOperation.m
```diff
@@ -169,9 +191,8 @@ - (NSDictionary *)HTTPHeaders
 
 - (void)abortWithError:(NSError *)error
 {
-    [self.connection cancel];
+    self.error = error;
```
boxtcarpel, line 194: This was an issue if stream setup would fail

BoxContentSDK/BoxContentSDK/Operations/BOXAPIMultipartToJSONOperation.m
```diff
@@ -92,7 +110,9 @@ - (NSURLSessionTask *)createSessionTask
     return sessionTask;
 }
 
-- (void)progressWithTotalBytesSent:(int64_t)totalBytesSent totalBytesExpectedToSend:(int64_t)totalBytesExpectedToSend
+- (void)sessionTask:(NSURLSessionTask *)sessionTask
```
boxtcarpel, line 113: name wasn't updated with a previous refactor

BoxContentSDK/BoxContentSDK/Operations/BOXAPIOperation.m
boxtcarpel: If we call processResponse with a nil response, it will always generate an error. If there is no response to process, we should not go through that - particularly here on 'finish' on a 'processIntermediateResponse' a nil response SHOULD generate an error but in this case it can be valid (for example a network error where the failure happens before reaching the server, like no connection or a DNS problem.)

BoxContentSDK/BoxContentSDK/Operations/BOXAPIOperation.m
boxtcarpel: Yeah, I'll add a comment. Basically the data gets initialized to [NSData data] which is non-nil empty data. If at the end it is still empty, we do not want to process the data because that could result in a false-negative error (like being unable to parse JSON from the data). Especially for JSON operations this is common because the network data gets stored in a separate variable and the data is left as-is.

BoxContentSDK/BoxContentSDK/Operations/BOXAPIMultipartToJSONOperation.m
irvingruan: Is there not a callback we need to invoke to propagate up the error or is assigning it to self.error sufficient?

boxtcarpel: The callback would get called from canceling the task. If this is so early in the process that a task does not yet exist (which can happen on initial stream setup), then setting the error is sufficient because it gets checked in that setup process.

BoxContentSDK/BoxContentSDK/Operations/BOXAPIOperation.m
```diff
@@ -540,18 +540,26 @@ - (void)processResponse:(NSURLResponse *)response
...
+        // (the responseData property that feeds into this gets initialized as empty NSData),
+        // ignore it. Otherwise we need to process the data.
         [self processResponseData:data];
     }
+
     if (self.error == nil) {
+        // If we do not already have an error set, use the passed in error.
         self.error = error;
     }
+
```
thuyBox, line 570: can we add a FIXME to review bringing this logic for if (self.error == nil) {self.error = error;} to the top of this method given the input error is coming from client, and we have expectation to preserve self.error if it was set earlier?

BoxContentSDK/BoxContentSDK/Operations/BOXAPIOperation.m
```diff
@@ -540,18 +540,26 @@ - (void)processResponse:(NSURLResponse *)response
 
 - (void)finishURLSessionTaskWithData:(NSData *)data response:(NSURLResponse *)response error:(NSError *)error
 {
-    if (self.HTTPResponse == nil) {
+    if (self.HTTPResponse == nil && response != nil) {
         // If the response has not already been handled (some, but not all, operations get
-        // 'processResponse:' earlier in the process) handle it here.
+        // 'processResponse:' earlier in the process), handle it here.
+        // Also, do not handle a nil response. That could happen if there was a failure on the whole
+        // network connection, for example. In that case, there is no response.
```
thuyBox, line 554: can we add comment for processResponse in BOXAPIOperation.h/.m that this method is expecting a valid response for processing, else that is considered an error.

Separately, should we check if self.error is nil inside processResponse before overriding it?

boxtcarpel: done. For the error handling, we'll do that in the scope of the investigation mentioned in the FIXME:

BoxContentSDK/BoxContentSDK/Operations/BOXAPIOperation.m
boxtcarpel: done

thuyBox:
+2 ! 👍

PR #338, opened 2017-03-08
Remove deprecated NSURLConnection functionality

boxtcarpel:


boxcla:

BoxContentSDK/BoxContentSDK/Operations/BOXAPIDataOperation.m
```diff
@@ -228,39 +223,8 @@ - (void)close
 - (void)abortWithError:(NSError *)error
 {
     [self close];
-    [self.connection cancel];
+    self.error = error;
```
thuyBox: should it be set if (self.error != nil)?

BoxContentSDK/BoxContentSDK/Operations/BOXAPIOperation.h
```diff
@@ -47,12 +47,12 @@ typedef void (^BOXSessionTaskReplacedBlock)(NSURLSessionTask *oldSessionTask, NS
...
  * 2xx range, with an exception made for 202. 202, 3xx, 4xx, and 5xx are all treated as errors.
  *
- * NSURLConnectionDataDelegate
+ * BOXURLSessionTaskDelegate
  * ===========================
- * BOXAPIOperation instances issue API calls using NSURLConnection. Each BOXAPIOperation acts
- * as the delegate for its own API call. [BOXAPIOperations](BOXAPIOperation) are not concurrent
- * NSOperations, so they keep the runloop they are running on open during the lifetime of the
- * request to enable receiving delegate callbacks.
+ * BOXAPIOperation instances issue API calls using NSURLSession. Each BOXAPIOperation acts
```
thuyBox, line 52: NSURLSessionTask

BoxContentSDK/BoxContentSDK/Operations/BOXStreamOperation.m
```diff
@@ -148,7 +148,9 @@ - (long long)contentLength
 
 - (void)abortWithError:(NSError *)error
 {
+    self.error = error;
```
thuyBox: should we check if (self.error == nil) before overriding?

BoxContentSDK/BoxContentSDK/Operations/BOXAPIDataOperation.m
boxtcarpel: I think the expectation is that the abort error gets set

BoxContentSDK/BoxContentSDK/Operations/BOXAPIOperation.h
boxtcarpel: I phrased it like this on purpose, saying we issue API calls using 'nsurlsession' (the general networking mechanism) and the operations are the delegates of their respective session task (see the next line).

What do you think? Would it be clearer to change it?

irvingruan: Since BOXAPIOperation is fairly high level, I think it makes sense to keep the abstraction terminology here. NSURLSession is as much as a class as it is a recognizable networking concept to iOS developers. Once we get into nitty gritty API calls, it makes more sense to use NSURLSessionTask if it's mentioned.

BoxContentSDK/BoxContentSDK/Operations/BOXAPIOperation.m
```diff
@@ -323,18 +312,10 @@ - (void)executeOperation
...
-            if (self.shouldUseSessionTask == YES) {
-                if (self.sessionTask == nil) {
-                    self.sessionTask = [self createSessionTask];
-                }
-                [self.sessionTask resume];
-            } else {
-                self.connection = [[NSURLConnection alloc] initWithRequest:self.APIRequest delegate:self];
-                BOXLog(@"Starting %@", self);
-                [self startURLConnection];
+            if (self.sessionTask == nil) {
```
irvingruan, line 315: Minor: We should use _sessionTask here when we're checking == nil. `createSessionTask` could be renamed `sessionTask`.

BoxContentSDK/BoxContentSDK/Operations/BOXStreamOperation.m
```diff
@@ -148,7 +148,9 @@ - (long long)contentLength
...
     [self.sessionTask cancel];
+    //FIXME: Remove?
```
irvingruan: Minor: let's add the date and author next to the FIXME. This adds context for future developers.

BoxContentSDK/BoxContentSDK/Operations/BOXAPIOperation.m
boxtcarpel: I disagree. As long as we don't have a lazy instantiation accessor there is no need to use the ivar directly and in fact that could have undesirable side-effects. If that changes we'd have to review usage of the accessor, but until then using the property is the correct approach.

BoxContentSDK/BoxContentSDK/Operations/BOXStreamOperation.m
boxtcarpel: I actually meant to call this out in the PR. I'd like to resolve this before pushing ideally.

I think this was from a previous change I made, where I called the finish directly because the URLConnection code was doing it like that before, however as of a recent change I made for the same thing in another class, I don't think we need to. I'll run another test and also wanted your guys' thoughts in case I was missing something.

BoxContentSDK/BoxContentSDK/Operations/BOXAPIDataOperation.m
thuyBox: previously we call [self connection:self.connection didFailWithError:error]; when abort, and inside didFailWithError:, we check if error!=nil before setting self.error

BoxContentSDK/BoxContentSDK/Operations/BOXAPIOperation.h
thuyBox: sounds good

BoxContentSDK/BoxContentSDK/Operations/BOXStreamOperation.m
thuyBox: similarly to previous comment

BoxContentSDK/BoxContentSDK/Operations/BOXStreamOperation.m
thuyBox: could you give some context to the recent change that makes this unnecessary?

thuyBox:
+2

BoxContentSDK/BoxContentSDK/Operations/BOXStreamOperation.m
boxtcarpel: We spoke offline. Basically, canceling the session task will call the didFinish:... (see documentation).

PR #339, opened 2017-03-08
add BOXURLSessionCacheClient to persist data for background download/…

thuyBox:
…upload tasks to assist reconnection

### DIFF
--- a/BoxContentSDK/BoxContentSDK/BOXContentSDKConstants.h
+++ b/BoxContentSDK/BoxContentSDK/BOXContentSDKConstants.h
@@ -443,9 +443,9 @@ extern NSString *const BOXAPIEnterpriseEventTypeItemSync;
 extern NSString *const BOXAPIEnterpriseEventTypeItemUnsync;
 
 //urlsessiontask cache dir, file prefix
-extern NSString *const BOXURLSessionTaskCacheDirName;
-extern NSString *const BOXURLSessionTaskCacheSessionsDirName;
-extern NSString *const BOXURLSessionTaskCacheUsersDirName;
+extern NSString *const BOXURLSessionTaskCacheDirectoryName;
+extern NSString *const BOXURLSessionTaskCacheSessionsDirectoryName;
+extern NSString *const BOXURLSessionTaskCacheUsersDirectoryName;
 extern NSString *const BOXURLSessionTaskCacheDestinationFilePath;
 extern NSString *const BOXURLSessionTaskCacheResumeData;
 extern NSString *const BOXURLSessionTaskCacheResponse;
--- a/BoxContentSDK/BoxContentSDK/BOXContentSDKConstants.h
+++ b/BoxContentSDK/BoxContentSDK/BOXContentSDKConstants.h
@@ -442,6 +442,24 @@ extern NSString *const BOXAPIEnterpriseEventTypeCollaborationExpiration;
 extern NSString *const BOXAPIEnterpriseEventTypeItemSync;
 extern NSString *const BOXAPIEnterpriseEventTypeItemUnsync;
 
+//urlsessiontask cache dir, file prefix
+extern NSString *const BOXURLSessionTaskCacheDirName;
+extern NSString *const BOXURLSessionTaskCacheSessionsDirName;
+extern NSString *const BOXURLSessionTaskCacheUsersDirName;
+extern NSString *const BOXURLSessionTaskCacheDestinationFilePath;
+extern NSString *const BOXURLSessionTaskCacheResumeData;
+extern NSString *const BOXURLSessionTaskCacheResponse;
+extern NSString *const BOXURLSessionTaskCacheResponseData;
+extern NSString *const BOXURLSessionTaskCacheError;
+
+typedef NS_ENUM(NSUInteger, BOXURLSessionTaskCacheFileType) {
+    BOXURLSessionTaskCacheFileTypeDestinationFilePath = 0,
+    BOXURLSessionTaskCacheFileTypeResumeData = 1,
+    BOXURLSessionTaskCacheFileTypeResponse = 2,
+    BOXURLSessionTaskCacheFileTypeResponseData = 3,
+    BOXURLSessionTaskCacheFileTypeError = 4
+};
+
 typedef NS_ENUM(NSUInteger, BOXThumbnailSize) {
     BOXThumbnailSize32 = 32,
     BOXThumbnailSize64 = 64,
--- a/BoxContentSDK/BoxContentSDK/BOXContentSDKConstants.m
+++ b/BoxContentSDK/BoxContentSDK/BOXContentSDKConstants.m
@@ -424,8 +424,8 @@ NSString *const BOXAPIEnterpriseEventTypeItemUnsync = @"ITEM_UNSYNC";
 
 
 //urlsessiontask cache dir, file prefix
-NSString *const BOXURLSessionTaskCacheDirName = @"BOXURLSessionCache";
-NSString *const BOXURLSessionTaskCacheSessionsDirName = @"sessions";
+NSString *const BOXURLSessionTaskCacheDirectoryName = @"BOXURLSessionCache";
+NSString *const BOXURLSessionTaskCacheSessionsDirectoryName = @"sessions";
 NSString *const BOXURLSessionTaskCacheUsersDirName = @"users";
 NSString *const BOXURLSessionTaskCacheDestinationFilePath = @"destinationFilePath";
 NSString *const BOXURLSessionTaskCacheResumeData = @"resumeData";
--- a/BoxContentSDK/BoxContentSDK/BOXContentSDKConstants.m
+++ b/BoxContentSDK/BoxContentSDK/BOXContentSDKConstants.m
@@ -421,3 +421,14 @@ NSString *const BOXAPIEnterpriseEventTypeCollaborationInvite = @"COLLABORATION_I
 NSString *const BOXAPIEnterpriseEventTypeCollaborationExpiration = @"COLLABORATION_EXPIRATION";
 NSString *const BOXAPIEnterpriseEventTypeItemSync = @"ITEM_SYNC";
 NSString *const BOXAPIEnterpriseEventTypeItemUnsync = @"ITEM_UNSYNC";
+
+
+//urlsessiontask cache dir, file prefix
+NSString *const BOXURLSessionTaskCacheDirName = @"BOXURLSessionCache";
+NSString *const BOXURLSessionTaskCacheSessionsDirName = @"sessions";
+NSString *const BOXURLSessionTaskCacheUsersDirName = @"users";
+NSString *const BOXURLSessionTaskCacheDestinationFilePath = @"destinationFilePath";
+NSString *const BOXURLSessionTaskCacheResumeData = @"resumeData";
+NSString *const BOXURLSessionTaskCacheResponse = @"response";
+NSString *const BOXURLSessionTaskCacheResponseData = @"responseData";
+NSString *const BOXURLSessionTaskCacheError = @"error";
--- a/BoxContentSDK/BoxContentSDK/BOXContentSDKErrors.h
+++ b/BoxContentSDK/BoxContentSDK/BOXContentSDKErrors.h
@@ -77,6 +77,12 @@ typedef NS_ENUM(NSUInteger, BOXContentSDKStreamError) {
     BOXContentSDKStreamErrorReadFailed = 30001
 };
 
+typedef NS_ENUM(NSUInteger, BOXContentSDKCacheError) {
+    BOXContentSDKSCacheErrorArchiveFailed = 40000,
+    BOXContentSDKSCacheErrorFileCreateFailed = 40001,
+    BOXContentSDKSCacheErrorNoValidSessionTaskGivenUserIdAssociateId = 40002
+};
+
 extern NSString *const BOXAuthTokenRequestErrorInvalidGrant; // Invalid refresh token
 extern NSString *const BOXAuthTokenRequestErrorInvalidToken; // Invalid access token
 extern NSString *const BOXAuthTokenRequestErrorInvalidRequest; // Possibly a missing access token
--- a/BoxContentSDK/BoxContentSDK/BOXContentSDKErrors.h
+++ b/BoxContentSDK/BoxContentSDK/BOXContentSDKErrors.h
@@ -77,10 +77,11 @@ typedef NS_ENUM(NSUInteger, BOXContentSDKStreamError) {
     BOXContentSDKStreamErrorReadFailed = 30001
 };
 
-typedef NS_ENUM(NSUInteger, BOXContentSDKCacheError) {
-    BOXContentSDKSCacheErrorArchiveFailed = 40000,
-    BOXContentSDKSCacheErrorFileCreateFailed = 40001,
-    BOXContentSDKSCacheErrorNoValidSessionTaskGivenUserIdAssociateId = 40002
+typedef NS_ENUM(NSUInteger, BOXContentSDKURLSessionCacheError) {
+    BOXContentSDKURLSessionCacheErrorFileCreateFailed = 50000,
+    BOXContentSDKURLSessionCacheErrorNoValidSessionForUserId = 50001,
+    BOXContentSDKURLSessionCacheErrorNoValidUserIdOrAssociateId = 50002,
+    BOXContentSDKURLSessionCacheErrorNoValidBackgroundSessionIdOrSessionTaskId = 50003
 };
 
 extern NSString *const BOXAuthTokenRequestErrorInvalidGrant; // Invalid refresh token
--- a/BoxContentSDK/BoxContentSDK/BOXURLSessionCacheClient.h
+++ b/BoxContentSDK/BoxContentSDK/BOXURLSessionCacheClient.h
@@ -1,0 +1,153 @@
+//
+//  BOXURLSessionCacheClient.h
+//  BoxContentSDK
+//
+//  Created by Thuy Nguyen on 3/3/17.
+//  Copyright © 2017 Box. All rights reserved.
+//
+
+#import <Foundation/Foundation.h>
+
+@interface BOXURLSessionTaskCachedInfo : NSObject
+
+@property (nonatomic, strong, readwrite) NSString *backgroundSessionId;
+@property (nonatomic, assign, readwrite) NSUInteger sessionTaskId;
+@property (nonatomic, strong, readwrite) NSString *destinationFilePath;
+@property (nonatomic, strong, readwrite) NSData *resumeData;
+@property (nonatomic, strong, readwrite) NSURLResponse *response;
+@property (nonatomic, strong, readwrite) NSData *responseData;
+@property (nonatomic, strong, readwrite) NSError *error;
+
+@end
+
+@protocol BOXURLSessionCacheClientDelegate <NSObject>
+
+// allow delegate to encrypt data before BOXURLSessionCacheClient persists it to disk
+- (NSData *)encryptData:(NSData *)data;
+
+// allow delegate to decrypt data which was encrypted by encryptData method above
+- (NSData *)decryptData:(NSData *)data;
+
+@end
+
+/**
+ * BOXURLSessionCacheClient is used to persist data for background download/upload session tasks to provide to task delegates
+ * for processing once the task delegates are ready.
+ *
+ * BOXURLSessionCacheClient will require a root directory to cache its data at initialization, and it will create the sub directory
+ * structures to persist data accordingly.
+ *
+ * 1. To store data specific to a background session task uniquely identified by background session id and session task id,
+ * we use sub-directory /sessions/$backgroundSessionId/$sessionTaskId, which will contain up to 5 files:
+ *
+ * - destinationFilePath: store the destination file path of downloaded file, applicable for background download task only
+ * - resumeData:          store the resume data of background download task which allows us to resume
+ *                        download from the point before the task was cancelled, applicable for background
+ *                        download task only
+ * - responseData:        store the response data (could contain error from server)
+ * - response:            store NSURLResponse
+ * - error:               store client-side NSError
+ *
+ * 2. To keep track of whose user and its associateId the session task belongs to, we save backgroundSessionId and sessionTaskId
+ * as a file name under sub-directory /users/$userId/$associateId with format $backgroundSessionId-$sessionTaskId
+ */
+@interface BOXURLSessionCacheClient : NSObject
+
+@property (nonatomic, weak, readwrite) id<BOXURLSessionCacheClientDelegate> delegate;
+
+/**
+ * Cache the relationship between the session task and the user who started it as well as its equivalent associateId
+ *
+ * @param userId                Id of user started the session task
+ * @param associateId           Id to associate with the session task
+ * @param backgroundSessionId   Id of the background session
+ * @param sessionTaskId         Id of the session task
+ * @param error                 error if fail to get
+ *
+ * @return YES if succeeded, NO if failed
+ */
+- (BOOL)cacheUserId:(NSString *)userId associateId:(NSString *)associateId backgroundSessionId:(NSString *)backgroundSessionId sessionTaskId:(NSUInteger)sessionTaskId error:(NSError **)error;
+
+/**
+ * Cache destinationFilePath of a background download session task
+ *
+ * @param backgroundSessionId   Id of the background session
+ * @param sessionTaskId         Id of the session task
+ * @param destinationFilePath   destination file path of downloaded file, applicable for background download task only
+ * @param error                 error if fail to cache
+ *
+ * @return YES if succeeded, NO if failed
+ */
+- (BOOL)cacheBackgroundSessionId:(NSString *)backgroundSessionId sessionTaskId:(NSUInteger)sessionTaskId destinationFilePath:(NSString *)destinationFilePath error:(NSError **)error;
+
+/**
+ * Cache resumeData of of a background download session task
+ *
+ * @param backgroundSessionId   Id of the background session
+ * @param sessionTaskId         Id of the session task
+ * @param resumeData            resume data to resume background download session task from
+ * @param error                 error if fail to cache
+ *
+ * @return YES if succeeded, NO if failed
+ */
+- (BOOL)cacheBackgroundSessionId:(NSString *)backgroundSessionId sessionTaskId:(NSUInteger)sessionTaskId resumeData:(NSData *)resumeData error:(NSError **)error;
+
+/**
+ * Cache responseData of a background session task
+ *
+ * @param backgroundSessionId   Id of the background session
+ * @param sessionTaskId         Id of the session task
+ * @param responseData          response data from the session task
+ * @param error                 error if fail to cache
+ *
+ * @return YES if succeeded, NO if failed
+ */
+- (BOOL)cacheBackgroundSessionId:(NSString *)backgroundSessionId sessionTaskId:(NSUInteger)sessionTaskId responseData:(NSData *)responseData error:(NSError **)error;
+
+/**
+ * Cache response of a background session task
+ *
+ * @param backgroundSessionId   Id of the background session
+ * @param sessionTaskId         Id of the session task
+ * @param response              NSURLResponse from the session task
+ * @param error                 error if fail to cache
+ *
+ * @return YES if succeeded, NO if failed
+ */
+- (BOOL)cacheBackgroundSessionId:(NSString *)backgroundSessionId sessionTaskId:(NSUInteger)sessionTaskId response:(NSURLResponse *)response error:(NSError **)error;
+
+/**
+ * Cache client-side error of a background session task
+ *
+ * @param backgroundSessionId   Id of the background session
+ * @param sessionTaskId         Id of the session task
+ * @param taskError             store client-side NSError of the session task
+ * @param error                 error if fail to cache
+ *
+ * @return YES if succeeded, NO if failed
+ */
+- (BOOL)cacheBackgroundSessionId:(NSString *)backgroundSessionId sessionTaskId:(NSUInteger)sessionTaskId taskError:(NSData *)taskError error:(NSError **)error;
+
+/**
+ * Get all cached data of the background session task associated with this userId and associateId
+ *
+ * @param userId            Id of user started the session task
+ * @param associateId       Id to associate with the session task
+ * @param error             error if fail to get
+ *
+ * @return BOXURLSessionTaskCache   all cached data of the session task
+ */
+- (BOXURLSessionTaskCachedInfo *)getCacheForUserId:(NSString *)userId associateId:(NSString *)associateId error:(NSError **)error;
+
+/**
+ * Delete all cached data of the background session task associated with this userId and associateId
+ *
+ * @param userId            Id of user started the session task
+ * @param associateId       Id to associate with the session task
+ * @param error             error if fail to delete
+ *
+ * @return YES if succeeded, NO if failed
+ */
+- (BOOL)deleteCachedInfoForUserId:(NSString *)userId associateId:(NSString *)associateId error:(NSError **)error;
+
+@end
--- a/BoxContentSDK/BoxContentSDK/BOXURLSessionCacheClient.h
+++ b/BoxContentSDK/BoxContentSDK/BOXURLSessionCacheClient.h
@@ -9,6 +9,7 @@
 #import <Foundation/Foundation.h>
 
 @interface BOXURLSessionTaskCachedInfo : NSObject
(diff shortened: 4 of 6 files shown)


boxcla:

BoxContentSDK/BoxContentSDK/BOXContentSDKConstants.h
```diff
@@ -442,6 +442,24 @@ extern NSString *const BOXAPIEnterpriseEventTypeCollaborationExpiration;
 extern NSString *const BOXAPIEnterpriseEventTypeItemSync;
 extern NSString *const BOXAPIEnterpriseEventTypeItemUnsync;
 
+//urlsessiontask cache dir, file prefix
+extern NSString *const BOXURLSessionTaskCacheDirName;
```
irvingruan: Minor: would prefer Dir to expand to Directory for clarity. Also, proper caps and space after // for consistency.

BoxContentSDK/BoxContentSDK/BOXContentSDKErrors.h
```diff
@@ -77,6 +77,12 @@ typedef NS_ENUM(NSUInteger, BOXContentSDKStreamError) {
     BOXContentSDKStreamErrorReadFailed = 30001
 };
 
+typedef NS_ENUM(NSUInteger, BOXContentSDKCacheError) {
+    BOXContentSDKSCacheErrorArchiveFailed = 40000,
+    BOXContentSDKSCacheErrorFileCreateFailed = 40001,
+    BOXContentSDKSCacheErrorNoValidSessionTaskGivenUserIdAssociateId = 40002
```
irvingruan: Minor: The naming for the last one is a bit confusing. Is `BOXContentSDKCacheErrorNoValidSessionForUserId` inaccurate?

BoxContentSDK/BoxContentSDK/BOXURLSessionCacheClient.h
```diff
@@ -0,0 +1,153 @@
...
+//
+//  Created by Thuy Nguyen on 3/3/17.
+//  Copyright © 2017 Box. All rights reserved.
+//
+
+#import <Foundation/Foundation.h>
+
+@interface BOXURLSessionTaskCachedInfo : NSObject
+
+@property (nonatomic, strong, readwrite) NSString *backgroundSessionId;
```
irvingruan: Copy for strings.

BoxContentSDK/BoxContentSDK/BOXURLSessionCacheClient.h
```diff
@@ -0,0 +1,153 @@
...
+@property (nonatomic, assign, readwrite) NSUInteger sessionTaskId;
+@property (nonatomic, strong, readwrite) NSString *destinationFilePath;
+@property (nonatomic, strong, readwrite) NSData *resumeData;
```
irvingruan, line 17: What does `resumeData` mean?

BoxContentSDK/BoxContentSDK/BOXURLSessionCacheClient.m
```diff
@@ -0,0 +1,297 @@
...
+//
+
+#import "BOXURLSessionCacheClient.h"
+#import "BOXContentSDKConstants.h"
+#import "BOXLog.h"
+#import "BOXContentSDKErrors.h"
+
+@interface BOXURLBackgroundSessionIdAndSessionTaskId : NSObject
+
+@property (nonatomic, strong, readwrite) NSString *backgroundSessionId;
```
irvingruan: Copy for strings.

BoxContentSDK/BoxContentSDK/BOXURLSessionCacheClient.m
```diff
@@ -0,0 +1,297 @@
...
+@property (nonatomic, assign, readwrite) NSUInteger sessionTaskId;
+
+@end
+
+@interface BOXURLSessionCacheClient()
+
+@property (nonatomic, strong, readwrite) NSString *cacheDir;
```
irvingruan: Copy for strings.

BoxContentSDK/BoxContentSDK/BOXURLSessionCacheClient.m
```diff
@@ -0,0 +1,297 @@
...
+@end
+
+@implementation BOXURLSessionCacheClient
+
+- (id)initWithCacheRootDir:(NSString *)cacheRootDir
+{
+    self = [super init];
+    if (self != nil) {
+        BOOL isDir = NO;
+        NSString *cacheDir = [cacheRootDir stringByAppendingPathComponent:BOXURLSessionTaskCacheDirName];
```
irvingruan: We should either have a default and/or check if `cacheRootDir` is nil.

BoxContentSDK/BoxContentSDK/BOXURLSessionCacheClient.m
```diff
@@ -0,0 +1,297 @@
...
+        if ([[NSFileManager defaultManager] fileExistsAtPath:cacheDir isDirectory:&isDir] == NO || isDir == NO) {
+            NSError *error;
+            [[NSFileManager defaultManager] createDirectoryAtPath:cacheDir withIntermediateDirectories:YES attributes:nil error:&error];
+            BOXAssert(error != nil, @"Failed to create cacheDir %@ with error %@", cacheDir, error);
```
irvingruan: Do we want to establish backup policy here in case we cannot create directory?

BoxContentSDK/BoxContentSDK/BOXURLSessionCacheClient.m
```diff
@@ -0,0 +1,297 @@
...
+    return [self cacheBackgroundSessionId:backgroundSessionId sessionTaskId:sessionTaskId data:data type:BOXURLSessionTaskCacheFileTypeResponse error:error];
+}
+
+- (BOOL)cacheBackgroundSessionId:(NSString *)backgroundSessionId sessionTaskId:(NSUInteger)sessionTaskId taskError:(NSData *)taskError error:(NSError **)error
+{
+    NSData *data = [NSKeyedArchiver archivedDataWithRootObject:taskError];
+    return [self cacheBackgroundSessionId:backgroundSessionId sessionTaskId:sessionTaskId data:data type:BOXURLSessionTaskCacheFileTypeError error:error];
+}
+
+- (BOXURLSessionTaskCachedInfo *)getCacheForUserId:(NSString *)userId associateId:(NSString *)associateId error:(NSError **)error
```
irvingruan: Minor: let's just use `cacheInfoForUserId` method name, instead of Java's standard.

BoxContentSDK/BoxContentSDK/BOXURLSessionCacheClient.m
```diff
@@ -0,0 +1,297 @@
...
+{
+    BOXURLBackgroundSessionIdAndSessionTaskId *backgroundSessionIdAndSessionTaskId = [self backgroundSessionIdAndSessionTaskIdGivenUserId:userId associateId:associateId];
+
+    if (backgroundSessionIdAndSessionTaskId == nil) {
+        if (error != nil) {
+        (*error) = [[NSError alloc] initWithDomain:BOXURLSessionTaskCacheError code:BOXContentSDKSCacheErrorNoValidSessionTaskGivenUserIdAssociateId userInfo:nil];
```
irvingruan: Minor: why parenthesis around error?

BoxContentSDK/BoxContentSDK/BOXURLSessionCacheClient.m
```diff
@@ -0,0 +1,297 @@
...
+
+    NSString *backgroundSessionId = backgroundSessionIdAndSessionTaskId.backgroundSessionId;
+    NSUInteger sessionTaskId = backgroundSessionIdAndSessionTaskId.sessionTaskId;
+
+    NSString *dir = [self dirPathOfSessionTaskWithBackgroundSessionId:backgroundSessionId sessionTaskId:sessionTaskId];
+
+    BOXURLSessionTaskCachedInfo *cachedInfo = [[BOXURLSessionTaskCachedInfo alloc] init];
+
+    //get all files under sessions/$backgroundSessionId/$sessionTaskId
+    NSArray *files = [[NSFileManager defaultManager] contentsOfDirectoryAtPath:dir error:error];
```
irvingruan: Minor: filePaths instead of files.

BoxContentSDK/BoxContentSDK/BOXURLSessionCacheClient.m
```diff
@@ -0,0 +1,297 @@
...
+        }
+    }
+    return cachedInfo;
+}
+
+- (BOOL)deleteCachedInfoForUserId:(NSString *)userId associateId:(NSString *)associateId error:(NSError **)error
+{
+    BOXURLBackgroundSessionIdAndSessionTaskId *backgroundSessionIdAndSessionTaskId = [self backgroundSessionIdAndSessionTaskIdGivenUserId:userId associateId:associateId];
+    if (backgroundSessionIdAndSessionTaskId == nil) {
+        return YES;
```
irvingruan: Do we want to return NO here or that if it isn't found, it's expected that the contract here will return YES?

BoxContentSDK/BoxContentSDK/BOXURLSessionCacheClient.m
```diff
@@ -0,0 +1,297 @@
...
+
+    if (success == YES) {
+        //clean up users/$userId/$associateId/*
+        dir = [self dirPathOfUserSessionTaskGivenUserId:userId associateId:associateId];
+        success = [self deleteDirectory:dir error:error];
+    }
+    return success;
+}
+
+#pragma mark - private helpers
```
irvingruan, line 228: Minor: proper caps.

BoxContentSDK/BoxContentSDK/BOXURLSessionCacheClient.m
```diff
@@ -0,0 +1,297 @@
...
+    if (outError != nil) {
+        (*outError) = error;
+    }
+    return success;
+}
+
+- (BOOL)cacheAndEncryptData:(NSData *)data atPath:(NSString *)path error:(NSError **)outError
+{
+    NSData *finalData = data;
+    if ([self.delegate respondsToSelector:@selector(encryptData:)]) {
```
irvingruan: If the delegate doesn't respond to the selector, can we really say that we encrypted the file? Might want to reflect an "attempt" abstraction in the method signature if this isn't always the case.

BoxContentSDK/BoxContentSDK/BOXURLSessionCacheClient.m
```diff
@@ -0,0 +1,297 @@
...
+- (BOOL)createDirForBackgroundSessionId:(NSString *)backgroundSessionId sessionTaskId:(NSUInteger)sessionTaskId error:(NSError **)error
+{
+    NSString *path = [self dirPathOfSessionTaskWithBackgroundSessionId:backgroundSessionId sessionTaskId:sessionTaskId];
+    return [self createDirectory:path error:error];
+}
+
+- (BOOL)createDirectory:(NSString *)path error:(NSError **)error
+{
+    BOOL isDir = NO;
+    BOOL success = YES;
```
irvingruan, line 294: We should default success to NO.

BoxContentSDK/BoxContentSDK/BOXURLSessionCacheClient.m
```diff
@@ -0,0 +1,297 @@
...
+    if ([[NSFileManager defaultManager] fileExistsAtPath:path isDirectory:&isDir] == NO || isDir == NO) {
+        success = [[NSFileManager defaultManager] createDirectoryAtPath:path withIntermediateDirectories:YES attributes:nil error:error];
+    }
+    return success;
+}
+
+- (BOOL)deleteDirectory:(NSString *)directoryPath error:(NSError **)error
+{
+    BOOL isDir = NO;
+    BOOL success = YES;
```
irvingruan, line 304: We should default success to NO.

BoxContentSDK/BoxContentSDK/BOXURLSessionCacheClient.m
```diff
@@ -0,0 +1,297 @@
...
+- (BOOL)createFile:(NSString *)path error:(NSError **)error
+{
+    BOOL success = [[NSFileManager defaultManager] createFileAtPath:path contents:nil attributes:nil];
+    if (success == NO && error != nil) {
+        (*error) = [[NSError alloc] initWithDomain:BOXContentSDKErrorDomain code:BOXContentSDKSCacheErrorFileCreateFailed userInfo:nil];
+    }
+    return success;
+}
+
+// path sessions/$backgroundSessionId/$sessionTaskId
```
irvingruan: MInor: can we properly caps/format these short method comments? e.g. `// Filepath Pattern: ...`

BoxContentSDK/BoxContentSDK/BOXURLSessionCacheClient.m
```diff
@@ -0,0 +1,297 @@
...
+- (NSString *)dirPathOfSessionTaskWithBackgroundSessionId:(NSString *)backgroundSessionId sessionTaskId:(NSUInteger)sessionTaskId
+{
+    return [[[self.cacheDir stringByAppendingPathComponent:BOXURLSessionTaskCacheSessionsDirName] stringByAppendingPathComponent:backgroundSessionId] stringByAppendingPathComponent:[NSString stringWithFormat:@"%lu", (unsigned long)sessionTaskId]];
```
irvingruan: stringByAppendingString is hard to parse. Let's use NSString stringWithFormat. Easier to read/change.

BoxContentSDK/BoxContentSDK/BOXURLSessionCacheClient.m
```diff
@@ -0,0 +1,297 @@
...
+}
+
+// path users/$userId/$associateId/$backgroundSessionId-$sessionTaskId
+- (NSString *)dirPathOfUserSessionTaskGivenUserId:(NSString *)userId associateId:(NSString *)associateId
+{
+    return [[[self.cacheDir stringByAppendingPathComponent:BOXURLSessionTaskCacheUsersDirName] stringByAppendingPathComponent:userId] stringByAppendingPathComponent:associateId];
```
irvingruan: Same for this.

BoxContentSDK/BoxContentSDK/BOXURLSessionCacheClient.m
```diff
@@ -0,0 +1,297 @@
...
+    NSString *fileName = [self fileNameGivenBackgroundSessionId:backgroundSessionId sessionTaskId:sessionTaskId];
+    return [path stringByAppendingPathComponent:fileName];
+}
+
+- (NSString *)fileNameGivenBackgroundSessionId:(NSString *)backgroundSessionId sessionTaskId:(NSUInteger)sessionTaskId
+{
+    return [NSString stringWithFormat:@"%@-%lu", backgroundSessionId, (unsigned long)sessionTaskId];
+}
+
+- (BOXURLBackgroundSessionIdAndSessionTaskId *)parseBackgroundSessionIdAndSessionTaskIdFileName:(NSString *)name
```
irvingruan, line 348: Minor: this method name is hard to read. Can we simplify this to something else?

BoxContentSDK/BoxContentSDK/BOXURLSessionCacheClient.m
```diff
@@ -0,0 +1,297 @@
...
+{
+    NSArray *arr = [name componentsSeparatedByString:@"-"];
+    BOXURLBackgroundSessionIdAndSessionTaskId *res = nil;
+    if (arr.count == 2) {
```
irvingruan, line 352: Can we get an example of what a string might look like to parse? Also, let's find a different variable name for `arr`.

BoxContentSDK/BoxContentSDK/BOXURLSessionCacheClient.m
```diff
@@ -0,0 +1,297 @@
...
+    }
+    return res;
+}
+
+- (BOXURLBackgroundSessionIdAndSessionTaskId *)backgroundSessionIdAndSessionTaskIdGivenUserId:(NSString *)userId associateId:(NSString *)associateId
+{
+    NSString *dir = [self dirPathOfUserSessionTaskGivenUserId:userId associateId:associateId];
+    NSError *error;
+    BOXURLBackgroundSessionIdAndSessionTaskId *res = nil;
+    NSArray *arr = [[NSFileManager defaultManager] contentsOfDirectoryAtPath:dir error:&error];
```
irvingruan: Arr -> filePaths

BoxContentSDK/BoxContentSDK/BOXURLSessionCacheClient.m
```diff
@@ -0,0 +1,297 @@
...
+        res.sessionTaskId = [arr[1] unsignedIntegerValue];
+    }
+    return res;
+}
+
+- (BOXURLBackgroundSessionIdAndSessionTaskId *)backgroundSessionIdAndSessionTaskIdGivenUserId:(NSString *)userId associateId:(NSString *)associateId
+{
+    NSString *dir = [self dirPathOfUserSessionTaskGivenUserId:userId associateId:associateId];
+    NSError *error;
+    BOXURLBackgroundSessionIdAndSessionTaskId *res = nil;
```
irvingruan: What's res? Let's use backgroundSessionIdAdSessionTaskId.

BoxContentSDK/BoxContentSDK/BOXContentSDKConstants.m
```diff
@@ -421,3 +421,14 @@
 NSString *const BOXAPIEnterpriseEventTypeCollaborationExpiration = @"COLLABORATION_EXPIRATION";
 NSString *const BOXAPIEnterpriseEventTypeItemSync = @"ITEM_SYNC";
 NSString *const BOXAPIEnterpriseEventTypeItemUnsync = @"ITEM_UNSYNC";
+
+
+//urlsessiontask cache dir, file prefix
+NSString *const BOXURLSessionTaskCacheDirName = @"BOXURLSessionCache";
+NSString *const BOXURLSessionTaskCacheSessionsDirName = @"sessions";
+NSString *const BOXURLSessionTaskCacheUsersDirName = @"users";
+NSString *const BOXURLSessionTaskCacheDestinationFilePath = @"destinationFilePath";
```
boxtcarpel: postfix the following values with 'Key'?

BoxContentSDK/BoxContentSDK/BOXContentSDKErrors.h
```diff
@@ -77,6 +77,12 @@ typedef NS_ENUM(NSUInteger, BOXContentSDKStreamError) {
     BOXContentSDKStreamErrorReadFailed = 30001
 };
 
+typedef NS_ENUM(NSUInteger, BOXContentSDKCacheError) {
+    BOXContentSDKSCacheErrorArchiveFailed = 40000,
```
boxtcarpel: overlapping with BOXContentSDKAppUserError.

Also, do we need one for 'SDKCacheErrorUnknown'?

Also, should the name be more specific to the networking/session/whatever cache for clarity?

BoxContentSDK/BoxContentSDK/BOXURLSessionCacheClient.h
```diff
@@ -0,0 +1,153 @@
...
+@property (nonatomic, strong, readwrite) NSURLResponse *response;
+@property (nonatomic, strong, readwrite) NSData *responseData;
+@property (nonatomic, strong, readwrite) NSError *error;
+
+@end
+
+@protocol BOXURLSessionCacheClientDelegate <NSObject>
```
boxtcarpel, line 24: generic name, but this is specifically an encryption delegate, we should call that out in the name.

BoxContentSDK/BoxContentSDK/BOXURLSessionCacheClient.h
```diff
@@ -0,0 +1,153 @@
...
+
+// allow delegate to encrypt data before BOXURLSessionCacheClient persists it to disk
+- (NSData *)encryptData:(NSData *)data;
```
boxtcarpel: These should be optional, no?

BoxContentSDK/BoxContentSDK/BOXURLSessionCacheClient.h
```diff
@@ -0,0 +1,153 @@
...
+ * - responseData:        store the response data (could contain error from server)
+ * - response:            store NSURLResponse
+ * - error:               store client-side NSError
+ *
+ * 2. To keep track of whose user and its associateId the session task belongs to, we save backgroundSessionId and sessionTaskId
+ * as a file name under sub-directory /users/$userId/$associateId with format $backgroundSessionId-$sessionTaskId
+ */
+@interface BOXURLSessionCacheClient : NSObject
+
+@property (nonatomic, weak, readwrite) id<BOXURLSessionCacheClientDelegate> delegate;
```
boxtcarpel: Should have a comment explaining that it is OK to leave this unset and that it would just mean the networking cache information will not be encrypted.

BoxContentSDK/BoxContentSDK/BOXURLSessionCacheClient.h
```diff
@@ -0,0 +1,153 @@
...
+ *
+ * @return BOXURLSessionTaskCache   all cached data of the session task
+ */
+- (BOXURLSessionTaskCachedInfo *)getCacheForUserId:(NSString *)userId associateId:(NSString *)associateId error:(NSError **)error;
+
+/**
+ * Delete all cached data of the background session task associated with this userId and associateId
+ *
+ * @param userId            Id of user started the session task
+ * @param associateId       Id to associate with the session task
```
boxtcarpel: can you pass nil for this to remove all information for a particular user? If so, should that be explained here in the comments?

BoxContentSDK/BoxContentSDK/BOXURLSessionCacheClient.h
```diff
@@ -0,0 +1,153 @@
...
+ * @param error             error if fail to delete
+ *
+ * @return YES if succeeded, NO if failed
+ */
+- (BOOL)deleteCachedInfoForUserId:(NSString *)userId associateId:(NSString *)associateId error:(NSError **)error;
```
boxtcarpel, line 161: Minor: Apple's naming convention involves capitalizing full acronyms (like URL), in this case ID. Also, for ID Apple tends to go with 'Identifier'

BoxContentSDK/BoxContentSDK/BOXURLSessionCacheClient.m
```diff
@@ -0,0 +1,297 @@
...
+@implementation BOXURLSessionCacheClient
+
+- (id)initWithCacheRootDir:(NSString *)cacheRootDir
+{
+    self = [super init];
+    if (self != nil) {
+        BOOL isDir = NO;
+        NSString *cacheDir = [cacheRootDir stringByAppendingPathComponent:BOXURLSessionTaskCacheDirName];
+        if ([[NSFileManager defaultManager] fileExistsAtPath:cacheDir isDirectory:&isDir] == NO || isDir == NO) {
+            NSError *error;
```
boxtcarpel: Should initialize to nil, to form the standard/habit on the team - even though Apple improved this for objects, for primitives this matters.

BoxContentSDK/BoxContentSDK/BOXURLSessionCacheClient.m
boxtcarpel: +1 we should not be using 'get' in Obj-C generally speaking, it has an association with inserting the value into a variable passed in by reference. If you want to indicate that it's being loaded from disk, etc, you can use 'fetch...'

BoxContentSDK/BoxContentSDK/BOXURLSessionCacheClient.m
boxtcarpel: Please note this isn't appendingString, it's appendingPathComponent which is the right thing to use for correct path formatting. Using stringWithFormat with manual slashes etc is error-prone and not change-safe.

We can still break this up for better readability if we think it's too much as-is.

BoxContentSDK/BoxContentSDK/BOXURLSessionCacheClient.h
thuyBox: thinking we don't need to be specific for encryption only, there might be other use cases. Added "@optional"

BoxContentSDK/BoxContentSDK/BOXURLSessionCacheClient.h
thuyBox: done

BoxContentSDK/BoxContentSDK/BOXURLSessionCacheClient.h
thuyBox: done

BoxContentSDK/BoxContentSDK/BOXURLSessionCacheClient.h
thuyBox: as of now, don't think we have to support deleting all cache for a userId, so I'll make it explicit that associateId cannot be nil

BoxContentSDK/BoxContentSDK/BOXContentSDKConstants.h
thuyBox: done

BoxContentSDK/BoxContentSDK/BOXContentSDKErrors.h
thuyBox: done

BoxContentSDK/BoxContentSDK/BOXContentSDKErrors.h
thuyBox: done

BoxContentSDK/BoxContentSDK/BOXURLSessionCacheClient.h
thuyBox: done

BoxContentSDK/BoxContentSDK/BOXURLSessionCacheClient.h
thuyBox: I have comment explaining those properties, I can add a @see below note

BoxContentSDK/BoxContentSDK/BOXURLSessionCacheClient.m
thuyBox: done

BoxContentSDK/BoxContentSDK/BOXURLSessionCacheClient.m
thuyBox: done

BoxContentSDK/BoxContentSDK/BOXURLSessionCacheClient.m
thuyBox: done

BoxContentSDK/BoxContentSDK/BOXURLSessionCacheClient.m
thuyBox: done

BoxContentSDK/BoxContentSDK/BOXURLSessionCacheClient.m
thuyBox: we attempt to create dir whenever we want to cache data again if dir doesn't exist.

BoxContentSDK/BoxContentSDK/BOXURLSessionCacheClient.m
thuyBox: done

BoxContentSDK/BoxContentSDK/BOXURLSessionCacheClient.m
thuyBox: done

BoxContentSDK/BoxContentSDK/BOXURLSessionCacheClient.m
thuyBox: done

BoxContentSDK/BoxContentSDK/BOXURLSessionCacheClient.m
thuyBox: added the check for error from backgroundSessionIdAndSessionTaskIdGivenUserId method, if there is an error, we return NO instead

BoxContentSDK/BoxContentSDK/BOXURLSessionCacheClient.m
thuyBox: done

BoxContentSDK/BoxContentSDK/BOXURLSessionCacheClient.m
thuyBox: we don't want to since the dir might exists already

BoxContentSDK/BoxContentSDK/BOXURLSessionCacheClient.m
thuyBox: we don't want to since the dir might already be deleted

BoxContentSDK/BoxContentSDK/BOXURLSessionCacheClient.m
thuyBox: done

BoxContentSDK/BoxContentSDK/BOXURLSessionCacheClient.m
thuyBox: done

BoxContentSDK/BoxContentSDK/BOXURLSessionCacheClient.m
thuyBox: done

BoxContentSDK/BoxContentSDK/BOXURLSessionCacheClient.m
thuyBox: done

BoxContentSDK/BoxContentSDK/BOXContentSDKConstants.m
thuyBox: don't think we need to for this case

BoxContentSDK/BoxContentSDK/BOXURLSessionCacheClient.h
boxtcarpel: IMO, unless we know/really think we're going to use it for something else, lets be specific at this point, it will make this stuff easier to consume/understand

BoxContentSDK/BoxContentSDK/BOXURLSessionCacheClient.m
```diff
@@ -0,0 +1,378 @@
...
+        res.backgroundSessionId = arr[0];
+        res.sessionTaskId = [arr[1] unsignedIntegerValue];
+    }
+    return res;
+}
+
+- (BOXURLBackgroundSessionIdAndSessionTaskId *)backgroundSessionIdAndSessionTaskIdGivenUserId:(NSString *)userId associateId:(NSString *)associateId error:(NSError **)error
+{
+    NSString *dir = [self dirPathOfUserSessionTaskGivenUserId:userId associateId:associateId];
+    NSError *err;
```
boxtcarpel, line 362: minor: initialize value

boxtcarpel (commented):
+1

BoxContentSDK/BoxContentSDK/BOXURLSessionCacheClient.h
irvingruan: To me, it sounds confusing because "resume" is a verb. Maybe "resumableData"?

BoxContentSDK/BoxContentSDK/BOXURLSessionCacheClient.m
irvingruan: If it already exists, it should return NO IMO. Technically, it did *not* create the directory this time, which is what the method says.

BoxContentSDK/BoxContentSDK/BOXURLSessionCacheClient.h
```diff
@@ -0,0 +1,163 @@
...
+@property (nonatomic, strong, readwrite) NSData *responseData;
+@property (nonatomic, strong, readwrite) NSError *error;
+
+@end
+
+@protocol BOXURLSessionCacheClientDelegate <NSObject>
+
+@optional
+// allow delegate to encrypt data before BOXURLSessionCacheClient persists it to disk
+- (NSData *)encryptData:(NSData *)data;
```
irvingruan, line 28: As with delegate patterns per Apple's recommendations, we should include the object in the delegate method, `- (NSData *)sessionCacheClient:(BOXURLSessionCacheClient *)client willEncryptData:(NSData *)data`.

PR #349, opened 2017-03-14
cancel on going session tasks and clean up all session tasks' info fo…

thuyBox:
…r a user upon logout

### DIFF
--- a/BoxContentSDK/BoxContentSDK/BOXURLSessionCacheClient.m
+++ b/BoxContentSDK/BoxContentSDK/BOXURLSessionCacheClient.m
@@ -404,6 +404,48 @@
     return success;
 }
 
+- (NSDictionary *)associateIdToBackgroundSessionIdAndSessionTaskIdsForUserId:(NSString *)userId error:(NSError **)outError
+{
+    if (userId == nil) {
+        return nil;
+    }
+
+    NSError *error = nil;
+    NSArray *associateIds = [self associateIdsForUserId:userId error:&error];
+    NSMutableDictionary *associateIdToBackgroundSessionIdAndSessionTaskIds = [NSMutableDictionary new];
+
+    //iterate through users/$userId/$associateId subdirs to get its backgroundSessionId and sessionTaskId
+    for (NSString *associateId in associateIds) {
+
+        BOXURLBackgroundSessionIdAndSessionTaskId *backgroundSessionIdAndSessionTaskId = [self backgroundSessionIdAndSessionTaskIdForUserId:userId associateId:associateId error:&error];
+
+        if (backgroundSessionIdAndSessionTaskId != nil) {
+            associateIdToBackgroundSessionIdAndSessionTaskIds[associateId] = backgroundSessionIdAndSessionTaskId;
+        } else if (error == nil) {
+            //we do not have valid backgroundSessionId and sessionTaskId for a previously seen userId and associateId,
+            //something must have failed, clean up the cache dir for future usage
+            [self deleteCachedInfoForUserId:userId associateId:associateId error:&error];
+        }
+    }
+    if (outError != nil) {
+        *outError = error;
+    }
+    return [associateIdToBackgroundSessionIdAndSessionTaskIds copy];
+}
+
+// Return associateIds under users/$userId dir
+- (NSArray *)associateIdsForUserId:(NSString *)userId error:(NSError **)error
+{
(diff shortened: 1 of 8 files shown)


boxcla:

BoxContentSDK/BoxContentSDK/BOXURLSessionCacheClient.m
```diff
@@ -404,6 +404,48 @@ - (BOOL)deleteCachedInfoForUserId:(NSString *)userId associateId:(NSString *)ass
...
+    NSMutableDictionary *associateIdToBackgroundSessionIdAndSessionTaskIds = [NSMutableDictionary new];
+
+    //iterate through users/$userId/$associateId subdirs to get its backgroundSessionId and sessionTaskId
+    for (NSString *associateId in associateIds) {
+
+        BOXURLBackgroundSessionIdAndSessionTaskId *backgroundSessionIdAndSessionTaskId = [self backgroundSessionIdAndSessionTaskIdForUserId:userId associateId:associateId error:&error];
+
+        if (backgroundSessionIdAndSessionTaskId != nil) {
+            associateIdToBackgroundSessionIdAndSessionTaskIds[associateId] = backgroundSessionIdAndSessionTaskId;
+        } else if (error == nil) {
```
boxtcarpel, line 424: What if the error is not nil? Any errors for which we'll still do the cleanup?

BoxContentSDK/BoxContentSDK/BOXURLSessionCacheClient.m
```diff
@@ -404,6 +404,48 @@ - (BOOL)deleteCachedInfoForUserId:(NSString *)userId associateId:(NSString *)ass
...
+            //we do not have valid backgroundSessionId and sessionTaskId for a previously seen userId and associateId,
+            //something must have failed, clean up the cache dir for future usage
+            [self deleteCachedInfoForUserId:userId associateId:associateId error:&error];
```
boxtcarpel, line 427: if we fail to delete, that's the error we'd pass back for the failure to get the ID listing?

BoxContentSDK/BoxContentSDK/BOXURLSessionManager.m
```diff
@@ -401,6 +401,48 @@ - (BOOL)cleanUpSessionTaskInfoGivenUserId:(NSString *)userId associateId:(NSStri
...
+
+    if (userId != nil) {
+        NSError *error = nil;
+
+        //cancel on-going session tasks associated with this userId
+        [self cancelOnGoingSessionTasksForUserId:userId error:&error];
+        BOXAssert(error == nil, @"Failed to cancel on-going session tasks with error %@", error);
+
+        //clean up memory entries and cached info of all session tasks associated with this userId
+        NSArray *associateIds = [self.cacheClient associateIdsForUserId:userId error:&error];
```
boxtcarpel, line 416: Why is this needed? Can't we just blow away the whole user directory?

boxtcarpel (approved):

BoxContentSDK/BoxContentSDK/BOXURLSessionManager.m
thuyBox: we could, but that only cleans up the cache dir, not in-memory info in self.backgroundSessionIdToSessionTask, so we just re-use the method that clean up all info relating to a session task

BoxContentSDK/BoxContentSDK/BOXURLSessionCacheClient.m
thuyBox: yes, essentially, we report back any errors we encounter during the method execution. Let me know if you think we should pass nil instead of &error

BoxContentSDK/BoxContentSDK/BOXURLSessionCacheClient.m
thuyBox: don't think there are other errors that we need to clean up. Let me know if you do.

BoxContentSDK/BoxContentSDK/BOXURLSessionCacheClient.m
boxtcarpel: I wonder if in that case we should create an indicative error for 'we failed to get the info and then we failed the cleanup', and then also have one for 'we failed to get the info', because if we're in this conditional that means we don't have an error object but we did not get the item, so it is a form of failure.

PR #350, opened 2017-03-15
additional cleaning up of data upon logging out and accept nil respon…

thuyBox:
…se data for a successful bg upload
and support reconnection to background sessions created from extensions

### DIFF
--- a/BoxContentSDK/BoxContentSDK/Operations/BOXAPIMultipartToJSONOperation.m
+++ b/BoxContentSDK/BoxContentSDK/Operations/BOXAPIMultipartToJSONOperation.m
@@ -132,6 +132,16 @@ totalBytesExpectedToSend:(int64_t)totalBytesExpectedToSend
     }
 }
 
+- (void)sessionTask:(NSURLSessionTask *)sessionTask didFinishWithResponse:(NSURLResponse *)response responseData:(NSData *)responseData error:(NSError *)error
+{
+    @synchronized (self) {
+        if ([self shouldRunInBackground] == YES) {
+            [[NSFileManager defaultManager] removeItemAtPath:self.uploadMultipartCopyFilePath error:nil];
+        }
+        [super sessionTask:sessionTask didFinishWithResponse:response responseData:responseData error:error];
+    }
+}
+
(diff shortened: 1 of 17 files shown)


boxcla:

BoxContentSDK/BoxContentSDK/Operations/BOXAPIMultipartToJSONOperation.m
```diff
@@ -132,6 +132,16 @@ - (void)sessionTask:(NSURLSessionTask *)sessionTask processIntermediateData:(NSD
     }
 }
 
+- (void)sessionTask:(NSURLSessionTask *)sessionTask didFinishWithResponse:(NSURLResponse *)response responseData:(NSData *)responseData error:(NSError *)error
+{
+    @synchronized (self) {
+        if ([self shouldRunInBackground] == YES) {
+            [[NSFileManager defaultManager] removeItemAtPath:self.uploadMultipartCopyFilePath error:nil];
```
boxtcarpel, line 139: Do we not need the copy on retries for failures/etc?

boxtcarpel (approved):

BoxContentSDK/BoxContentSDK/Operations/BOXAPIMultipartToJSONOperation.m
boxtcarpel: We also discussed offline about centralizing cleanup (finish?)

boxtcarpel:
We also spoke offline about forcing cleanup, making sure more that everything on disk goes away at logout.

BoxContentSDK/BoxContentSDK/Operations/BOXAPIMultipartToJSONOperation.m
thuyBox: we won't be able to retry upload using the same operation once it's finished. User can create a new request and the temp file will be re-created.

thuyBox:
will follow up with a PR to force clean up of userId cache upon logout even if not empty, and handle background tasks without a userId gracefully.

PR #352, opened 2017-03-15
support resumable background download

thuyBox:


boxcla:

BoxContentSDK/BoxContentSDK/BOXURLSessionCacheClient.h
```diff
@@ -308,9 +308,30 @@
...
 
+/**
+ * Return resumeData of a completed session task associated with the userId and associateId
+ *
+ * @param userId        Id of user started the session task. Cannot be nil
+ * @param associateId   Id associate with the session task for this user
+ *
+ * @return resume data for background download task which has completed
+ */
+- (NSData *)resumeDataForUserId:(NSString *)userId associateId:(NSString *)associateId;
```
irvingruan, line 324: Minor: resumableData, resume is not an adjective

BoxContentSDK/BoxContentSDK/Operations/BOXAPIDataOperation.h
```diff
@@ -77,6 +77,12 @@ typedef void (^BOXAPIDataProgressBlock)(long long expectedTotalBytes, unsigned l
  */
 @property (nonatomic, readwrite, strong) NSString *destinationPath;
 
+/**
+ * This indicates whether background download should be cancelled with intention to resume
+ * so we maintain resume information to allow a later operation to resume the download from where it was left off
+ */
+@property (nonatomic, readwrite, assign) BOOL allowResume;
```
irvingruan, line 84: Minor: resumableDownloadsEnabled

BoxContentSDK/BoxContentSDK/BOXURLSessionCacheClient.h
boxtcarpel: I'm not sure the data itself is resumable, it's not meant as a description of the data but rather a classification of the data, this is 'resume' data, to be used for resuming the operation... I think it could possibly be improved on, but I don't think resumableData is clearer

BoxContentSDK/BoxContentSDK/Operations/BOXAPIOperation.m
```diff
@@ -53,7 +53,7 @@ static BOOL BoxOperationStateTransitionIsValid(BOXAPIOperationState fromState, B
 
 @interface BOXAPIOperation()
 
-- (void)cancelConnection;
+- (void)cancelSessionTask;
```
boxtcarpel, line 56: We can rename this if we want, but then it's again tech-specific. I left it because I was just thinking of it as an abstraction "cancelConnection" It's a network connection no matter what technology we use under the covers to do that.

Is there a real need for this to be specific for session task? Does it do sessionTask-unique things?

boxtcarpel: Hm... I guess maybe with the resume stuff...

boxtcarpel (approved):

BoxContentSDK/BoxContentSDK/Operations/BOXAPIOperation.m
thuyBox: yeah, it is cancelling session task explicitly and potentially producing resume data

PR #362, opened 2017-03-17
Fix the lost behavior on receiving 202 responses for network calls

boxtcarpel:
moved from performSelector: to dispatch_after.

From Apple's documentation for performSelector:withObject:afterDelay:
Special Considerations
This method registers with the runloop of its current context,
and depends on that runloop being run on a regular basis to perform
correctly. One common context where you might call this method and
end up registering with a runloop that is not automatically run on
a regular basis is when being invoked by a dispatch queue.
If you need this type of functionality when running on a dispatch
queue, you should use dispatch_after and related methods to get
the behavior you want.


boxcla:

crousselle:
+2

PR #374, opened 2017-04-04
Login page fails to load after a device trust failure

boxtcarpel:
The login view controller was not getting deallocated, because the
NSURLSession was holding a strong reference to it as the delegate.

### DIFF
--- a/BoxContentSDK/BoxContentSDK/OAuth2/BOXAuthorizationViewController.m
+++ b/BoxContentSDK/BoxContentSDK/OAuth2/BOXAuthorizationViewController.m
@@ -36,6 +36,8 @@ typedef void (^BOXAuthCancelBlock)(BOXAuthorizationViewController *authorization
 @property (nonatomic, readwrite, assign) NSHTTPCookieAcceptPolicy preexistingCookiePolicy;
 
 @property (nonatomic, readwrite, strong) BOXContentClient *SDKClient;
+//NOTE: For this object to get properly deallocated, the URL Sesssion object has to be invalidated before
+// the completion blocks are called (since it holds a strong reference to its delegate until it is invalidated).
 @property (nonatomic, readwrite, copy) BOXAuthCompletionBlock completionBlock;
 @property (nonatomic, readwrite, copy) BOXAuthCancelBlock cancelBlock;
 @property (nonatomic, readwrite, strong) NSURL *authorizeURL;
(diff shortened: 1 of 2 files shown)


boxcla:

BoxContentSDK/BoxContentSDK/OAuth2/BOXAuthorizationViewController.m
```diff
@@ -149,7 +150,8 @@ - (void)viewDidLoad
     } else if ([self.SDKClient.session isKindOfClass:[BOXAppUserSession class]]) {
         __weak BOXAuthorizationViewController *me = self;
         [self.SDKClient autheticateAppUserWithCompletionBlock:^(BOXUser *user, NSError *error) {
-            self.completionBlock(me, user, error);
+            [me.URLSession invalidateAndCancel];
```
boxtcarpel: One alternate approach I was considering is creating a 'close/dismiss' method for the auth view controller, or even a 'pre-close/dismiss cleanup' method (since we don't control how it is presented) which the caller would have to use. This is more likely to work well for all 3rd parties without changes, but this has the issue of someone dismissing this VC by some outside mechanism (not driven by the auth VC). One possibility would be to have both - this approach, plus a pre-dismiss method that the caller can use before externally doing away with this VC. What do you guys think?

BoxContentSDK/BoxContentSDK/OAuth2/BOXAuthorizationViewController.m
```diff
@@ -102,7 +104,7 @@ - (instancetype)initWithSDKClient:(BOXContentClient *)SDKClient
 
         NSURLSessionConfiguration *sessionConfiguration = [NSURLSessionConfiguration defaultSessionConfiguration];
         _URLSession = [NSURLSession sessionWithConfiguration:sessionConfiguration
-                                                    delegate:self
+                                                    delegate:self //NOTE: The URL Session holds a strong reference on its delegate until it is invalidated
```
boxtcarpel, line 107: Per Apple's docs, this this is a 'strong' reference, which does not get cleaned up until the delegate is set to something else or the session is invalidated.

crousselle (approved):

BoxContentSDK/BoxContentSDK/OAuth2/BOXAuthorizationViewController.m
adempsey: Setting `URLSession` to `nil` explicitly in `dealloc` wouldn't work in this case?

boxtcarpel: dealloc would never get called.

adempsey (approved):

adempsey (approved):

PR #377, opened 2017-04-14
Migrate HLS representations api to url_template

jdizoglio:


boxcla:

BoxContentSDK/BoxContentSDK/BOXContentSDKConstants.h
```diff
@@ -161,18 +162,19 @@ extern BOXRepresentationType *const BOXRepresentationTypeHLS;
 extern BOXRepresentationType *const BOXRepresentationTypeCrocodoc;
 extern BOXRepresentationType *const BOXRepresentationTypeDICOM;
 
+// Representations URL Template
+extern NSString *const BOXRepresentationTemplateKeyAccessPath;
+
+// Representation Template Value
+extern NSString *const BOXRepresentationTemplateValueAccessPath;
```
jlawton: Maybe change this constant name to reference HLS specifically, since we'll have to make this more generic in the future.

jdizoglio: agree, done.

BoxContentSDK/BoxContentSDKTests/file_all_fields_representations.json
```diff
@@ -0,0 +1,108 @@
...
+                    },
+                    "links": {
+                        "info": {
+                            "url": "https:\/\/api.box.com\/2.0\/internal_files\/125041766103\/versions\/133811379927\/representations\/hls"
+                        },
+                        "content": {
+                            "type": "manifest",
+                            "url": "https:\/\/dl.boxcloud.com\/api\/2.0\/internal_files\/125041766103\/versions\/133811379927\/representations\/hls\/content\/master.m3u8"
+                        }
+                    },
```
ahmedcharles: Remove links.

BoxContentSDK/BoxContentSDKTests/file_all_fields_representations.json
```diff
@@ -0,0 +1,108 @@
...
+                    "info": {
+                        "url": "https:\/\/api.box.com\/2.0\/internal_files\/125041766103\/versions\/133811379927\/representations\/hls"
+                    },
+                    "content": {
+                        "url_template": "https:\/\/dl.boxcloud.com\/api\/2.0\/internal_files\/125041766103\/versions\/133811379927\/representations\/hls\/content\/{+asset_path}"
+                    },
+                    "details": {
+                    
+                    },
```
ahmedcharles: Remove details.

BoxContentSDK/BoxContentSDKTests/file_all_fields_representations.json
```diff
@@ -0,0 +1,108 @@
...
+                    {
+                    "type": "collection",
+                    "id": "10047",
+                    "name": "Favorites",
+                    "collection_type": "favorites"
+                    }
+                    ],
+    "representations": {
+        "entries": [
+        {
```
ahmedcharles: The whitespace here is sort of weird.

BoxContentSDK/BoxContentSDKTests/file_all_fields_representations.json
```diff
@@ -0,0 +1,108 @@
...
+            "representation": "hls",
+                    "properties": {
+
```
ahmedcharles: Don't really need the extra newlines?

BoxContentSDK/BoxContentSDK/Models/BOXRepresentation.m
```diff
@@ -41,27 +46,26 @@ - (instancetype)initWithJSON:(NSDictionary *)JSONResponse
                                                                     nullAllowed:NO];
         self.details = detailsJSON;
         
-        NSDictionary *linksJSON = [NSJSONSerialization box_ensureObjectForKey:BOXAPIObjectKeyLinks
+        NSDictionary *contentJSON = [NSJSONSerialization box_ensureObjectForKey:BOXAPIObjectKeyContent
                                                                  inDictionary:JSONResponse
                                                               hasExpectedType:[NSDictionary class]
                                                                   nullAllowed:NO];
```
ahmedcharles: Is the indentation off here now (ObjC newbie question)?

BoxContentSDK/BoxContentSDK/Models/BOXRepresentation.m
```diff
@@ -41,27 +46,26 @@ - (instancetype)initWithJSON:(NSDictionary *)JSONResponse
...
         
-        NSDictionary *contentJSON = linksJSON[BOXAPIObjectKeyContent];
-        
-        NSString *contentURLString = [NSJSONSerialization box_ensureObjectForKey:BOXAPIObjectKeyURL
+        NSString *contentURLString = [NSJSONSerialization box_ensureObjectForKey:BOXAPIObjectKeyURLTemplate
                                                              inDictionary:contentJSON
                                                           hasExpectedType:[NSString class]
                                                               nullAllowed:NO];
         if (contentURLString.length > 0) {
+            if ([self.type isEqualToString:BOXRepresentationTypeHLS]) {
```
ahmedcharles: This should probably return `nil` (which I think is the error indicator) if it's not HLS, because otherwise, your `[NSURL URLWithString:contentURLString]` will fail below due to invalid url characters (`{}` are guaranteed to be part of the url_template).

I know this isn't changing, but `name` or `representationName` would be a better identifier than `type`.

BoxContentSDK/BoxContentSDK/Models/BOXRepresentation.m
jdizoglio: colon aligned. Need to adjust with first colon.

BoxContentSDK/BoxContentSDK/Models/BOXRepresentation.m
jdizoglio: URLWithString will set contentURL to nil for this failure case. 

But this brings up a good point, we are not handling mp4. Can I append {+asset_path} as 'preview.mp4' for mp4 reps?
Ex:
"url_template": "https:\/\/dl3.boxcloud.com\/api\/2.0\/internal_files\/125041766103\/versions\/133811379927\/representations\/mp4\/content\/{+asset_path}"

With contentURL as nil means we are currently we are failing over to our old api call.

BoxContentSDK/BoxContentSDK/Models/BOXRepresentation.m
jdizoglio: the new xcode 8.3 isn't play nice. Seems to be centering our code. Will leave as is.

jdizoglio: looks like it formats correctly in the PR.

BoxContentSDK/BoxContentSDKTests/file_all_fields_representations.json
jdizoglio: done

BoxContentSDK/BoxContentSDKTests/file_all_fields_representations.json
jdizoglio: done

BoxContentSDK/BoxContentSDKTests/file_all_fields_representations.json
jdizoglio: done

BoxContentSDK/BoxContentSDKTests/file_all_fields_representations.json
jdizoglio: done

jdizoglio: xcode 8.3 seems to have formatting issues, attempted to format by hand and getting different results in git than in xcode. Will leave as is.

BoxContentSDK/BoxContentSDK/Models/BOXRepresentation.m
jdizoglio: I was going to add mp4 checks here, however, down stream we bypass reps and fallback. Will migrate this when we integrate the reset of reps.

As far as setting contentURL to nil, we are ok because URLWithString returns nil for failure. In other words, the result will be nil for all but HLS types.

jdizoglio:
Updated the file request header to reflect that we only support HLS streaming through BOXRepresentation at this time.

BoxContentSDK/BoxContentSDK/BOXContentSDKConstants.m
```diff
@@ -274,6 +282,7 @@
 NSString *const BOXAPIObjectKeyCanSeeManagedUsers = @"can_see_managed_users";
 NSString *const BOXAPIObjectKeyIsSyncEnabled = @"is_sync_enabled";
 NSString *const BOXAPIObjectKeyStatus = @"status";
+NSString *const BOXAPIObjectKeyState = @"state";
```
boxtcarpel, line 285: what about 'BOXAPIMetadataObjectKeyAccessPath' ?

BoxContentSDK/BoxContentSDK/Models/BOXRepresentation.m
```diff
@@ -41,27 +46,26 @@ - (instancetype)initWithJSON:(NSDictionary *)JSONResponse
...
+        NSDictionary *contentJSON = [NSJSONSerialization box_ensureObjectForKey:BOXAPIObjectKeyContent
+                                                                   inDictionary:JSONResponse
+                                                                hasExpectedType:[NSDictionary class]
+                                                                    nullAllowed:NO];
         
-        NSString *contentURLString = [NSJSONSerialization box_ensureObjectForKey:BOXAPIObjectKeyURL
-                                                             inDictionary:contentJSON
-                                                          hasExpectedType:[NSString class]
-                                                              nullAllowed:NO];
+        NSString *contentURLString = [NSJSONSerialization box_ensureObjectForKey:BOXAPIObjectKeyURLTemplate
```
boxtcarpel, line 54: should we update the name of the variable?

BoxContentSDK/BoxContentSDK/Requests/BOXFileRequest.h
```diff
@@ -10,6 +10,11 @@
 @property (nonatomic, readwrite, assign) BOOL requestAllFileFields;
 @property (nonatomic, readwrite, strong) NSArray *notMatchingEtags;//If-None-Match: Array of strings representing etag values
 
+/*
+ * Set this string to "[hls]" request a representation endpoint for HLS video streaming.
```
boxtcarpel: lets talk about how best to expose this. I think we can abstract it a bit.

BoxContentSDK/BoxContentSDK/Requests/BOXFileRequest.m
```diff
@@ -50,9 +50,16 @@ - (BOXAPIOperation *)createOperation
     NSDictionary *queryParameters = nil;
 
     if (self.requestAllFileFields) {
-        queryParameters = @{BOXAPIParameterKeyFields: [self fullFileFieldsParameterString]};
-    }
+        NSString *fieldString = [self fullFileFieldsParameterString];
+        
+        // Include the representations field in the request supporting the request for file representation
+        if ([self.httpHeaderValueXRepHints length] > 0) {
```
boxtcarpel: this strikes me as confusing, as per the comment above, lets review this.

BoxContentSDK/BoxContentSDK/Requests/BOXRequest.m
```diff
@@ -325,8 +325,7 @@ - (NSString *)fullFolderFieldsParameterString
                        BOXAPIObjectKeyExtension,
                        BOXAPIObjectKeyIsPackage,
                        BOXAPIObjectKeyAllowedSharedLinkAccessLevels,
-                       BOXAPIObjectKeyCollections,
-                       BOXAPIObjectKeyRepresentations];
+                       BOXAPIObjectKeyCollections];
```
boxtcarpel, line 328: what's the impact of this change? I thought we'd already made the performance related change on this, is this in addition or was my understanding incorrect?

BoxContentSDK/BoxContentSDK/Requests/BOXFileRequest.h
ahmedcharles: I'm also not sure about how to write an API that exposes this well. But we probably don't have time to make an awesome API with this change, so I suppose I'm ok with it for now if we can revisit later.

BoxContentSDK/BoxContentSDK/BOXContentSDKConstants.m
ahmedcharles: I'm not sure what this comment means?

BoxContentSDK/BoxContentSDK/Models/BOXRepresentation.m
```diff
@@ -41,27 +46,26 @@ - (instancetype)initWithJSON:(NSDictionary *)JSONResponse
...
+                                                                    inDictionary:contentJSON
+                                                                 hasExpectedType:[NSString class]
+                                                                     nullAllowed:NO];
         if (contentURLString.length > 0) {
+            if ([self.type isEqualToString:BOXRepresentationTypeHLS]) {
+                contentURLString = [contentURLString stringByReplacingOccurrencesOfString:BOXRepresentationTemplateKeyAccessPath withString:BOXRepresentationTemplateValueHLSManifiest];
+            }
             self.contentURL = [NSURL URLWithString:contentURLString];
```
ahmedcharles, line 62: I thought I commented on this before. But if you don't do the replacement, this will fail. So, the if check should check for hls and fail if it isn't.

BoxContentSDK/BoxContentSDK/BOXContentSDKConstants.m
jdizoglio: should be removed.

BoxContentSDK/BoxContentSDK/Requests/BOXFileRequest.m
ahmedcharles: Given the implementation above, this makes sense to me. (I'm surprised that there isn't an `isEmpty` function that could be used instead, perhaps? But that's a different thing.)

The idea here is that `fields=representations` should only be requested if `X-Rep-Hints` will be provided. How the API is structured to provide the content of `X-Rep-Hints` is debatable, but in the end, it comes down to a string used as a header and if it's empty, we don't want to ask for the `representations` field.

BoxContentSDK/BoxContentSDK/Requests/BOXRequest.m
ahmedcharles: This change makes it so that we only ask for `representations` when `X-Rep-Hints` is provided and non-empty. That's what the server side API expects/supports.

BoxContentSDK/BoxContentSDKTests/file_all_fields_representations.json
```diff
@@ -0,0 +1,95 @@
...
+       "id": "10047",
+       "name": "Favorites",
+       "collection_type": "favorites"
+     }
+    ],
+    "representations": {
+       "entries": [
+        {
+        "representation": "hls",
+          "properties": {
```
ahmedcharles: Nit: whitespace. This is indented 2 spaces too much. 💯

BoxContentSDK/BoxContentSDK/Models/BOXRepresentation.m
jdizoglio: It is a string type in our json. Will leave as named.

BoxContentSDK/BoxContentSDK/Requests/BOXFileRequest.h
jdizoglio: Per our offline conversation. Moving the request construction into our content SDK file request. Until we integrate representations into our app, we will make this a high definition video request. A bool will serve this need.

BoxContentSDK/BoxContentSDK/Requests/BOXFileRequest.m
jdizoglio: This changed per our team discussion. We will use a bool to build the high definition video request. The comments are updated to support the change.

BoxContentSDK/BoxContentSDK/Requests/BOXRequest.m
jdizoglio: Per discussion we are moving representation to support high definition video file request only.

BoxContentSDK/BoxContentSDK/Models/BOXRepresentation.m
jdizoglio: I had responded, git commit seems to have removed. In objc, nil is a valid value and set by URLWithString. Expecting nil here is intentional.

BoxContentSDK/BoxContentSDKTests/file_all_fields_representations.json
jdizoglio: used jq. Should be good now.

BoxContentSDK/BoxContentSDK/Models/BOXRepresentation.m
ahmedcharles: I mean, wouldn't the entire BoxRepresentation be nil rather than just this field? What's the benefit of having a non-nil BoxRepresentation which can't be used?

BoxContentSDK/BoxContentSDK/BOXContentSDKConstants.m
boxtcarpel: There was an extern that was never defined, probably left over from a refactor/cleanup. It has since been removed. Thanks Jim.

ahmedcharles:
I'm fine with this change, though I'm still unsure about the error handling. If everyone else thinks it's fine, then I'll punt, since I don't know enough either way. :)

boxtcarpel (approved):

BoxContentSDK/BoxContentSDKTests/BOXFileTests.m
```diff
@@ -176,8 +176,88 @@ - (void)test_that_file_with_all_fields_is_parsed_correctly_from_json
...
+    BOXFile *file = [[BOXFile alloc] initWithJSON:dictionary];
+    
+    XCTAssertEqualObjects(@"123", file.modelID);
+    XCTAssertEqualObjects(@"Halloweenie-3.jpg", file.name);
+    XCTAssertEqualObjects(@"0", file.etag);
+    XCTAssertNil(file.sequenceID);
+    XCTAssertEqualObjects(@"", file.itemDescription);
+    
+    NSNumber *fileSize = [NSNumber numberWithInteger:622592];
+    XCTAssertEqualObjects(fileSize, file.size);
```
jlawton, line 198: I think we could jut write this as `XCTAssertEqualObjects(@622592, file.size);`, since it's only used once.

jdizoglio:
test this please

BoxContentSDK/BoxContentSDKTests/BOXFileTests.m
```diff
@@ -176,8 +176,88 @@ - (void)test_that_file_with_all_fields_is_parsed_correctly_from_json
...
+    [self assertModel:owner isEquivalentTo:file.owner];
+    XCTAssertEqualObjects(@"13338532", file.owner.modelID);
+    
+    BOXFolderMini *parentFolder = [[BOXFolderMini alloc] initWithJSON:dictionary[BOXAPIObjectKeyParent]];
+    [self assertModel:parentFolder isEquivalentTo:file.parentFolder];
+    XCTAssertEqualObjects(@"0", file.parentFolder.modelID);
+    
+    XCTAssertEqualObjects(@"active", file.status);
+    XCTAssertEqualObjects(@"f34df557706317b50ee400e072eb731032fe27d3", file.SHA1);
+    XCTAssertEqualObjects([NSNumber numberWithInteger:0], file.commentCount);
```
jlawton, line 233: Could also use `@0`

PR #381, opened 2017-04-24
Adding support for single file collaboration.

scottliubox:


boxcla:

BoxContentSDK/BoxContentSDK.xcodeproj/project.xcworkspace/xcshareddata/BoxContentSDK.xcscmblueprint
```diff
@@ -4,12 +4,12 @@
```
crousselle, line 4: probably should reverse this, isn't this in the .gitignore ??

BoxContentSDK/BoxContentSDK/Requests/BOXCollaborationCreateRequest.m
```diff
@@ -8,12 +8,19 @@
 #import "BOXAPIJSONOperation.h"
 #import "BOXCollaboration.h"
 
+@interface BOXCollaborationCreateRequest()
+
+@property (nonatomic, readwrite, strong) BOXAPIItemType *itemType;
```
crousselle, line 13: minor - for consistency, either we should have the 2 public readonly properties here, or none (we're missing the item ID). Since we're only manipulating the ivar, it might be okay without ?

BoxContentSDK/BoxContentSDK/Requests/BOXFileCollaborationsRequest.h
```diff
@@ -0,0 +1,37 @@
...
+ For example, after requesting the first 100 records, you can provide the nextMarker to fetch another set of
+ records beginning at the 101st item in the listing. The next available marker ID is returned in the response header
+ of each paged recent items request.
+ */
+@property (nonatomic, readwrite, strong) NSString *nextMarker;
+
+- (instancetype)initWithFileID:(NSString *)fileID;
+
+//Perform API request and any cache update only if refreshBlock is not nil
+- (void)performRequestWithCompletion:(BOXFileCollaborationArrayCompletionBlock)completionBlock;
```
crousselle, line 31: can we add a regular method documentation here ?

BoxContentSDK/BoxContentSDK/Requests/BOXFileCollaborationsRequest.h
```diff
@@ -0,0 +1,37 @@
...
+
+//Perform API request and any cache update only if refreshBlock is not nil
```
crousselle, line 33: same here

BoxContentSDK/BoxContentSDK/Requests/BOXCollaborationCreateRequest.m
crousselle: or maybe I'm wrong, I thought you could not use ivar without readwrite, but I guess I was mistaken ?

BoxContentSDK/BoxContentSDKTests/BOXCollaborationCreateRequestTests.m
```diff
@@ -21,7 +21,8 @@ @implementation BOXCollaborationCreateRequestTests
 - (void)test_that_request_with_user_id_has_expected_URLRequest
```
crousselle, line 21: we should add testing coverage to the `BOXCollaborationCreateRequest` created via the new file specific initializers

crousselle:
nice work, super clean ! A few minor things

BoxContentSDK/BoxContentSDK/Requests/BOXCollaborationCreateRequest.m
scottliubox: Personally I like to have a properties instead of manipulating iVars directly. But yes, I'll add a readwrite version of itemID.

crousselle: yeah I prefer this as well :)

BoxContentSDK/BoxContentSDK/Requests/BOXFileCollaborationsRequest.h
scottliubox: This is actually the standard doc we have for all performRequest methods in all of our request classes. I do agree that we can make it better in the future.

BoxContentSDK/BoxContentSDKTests/BOXCollaborationCreateRequestTests.m
scottliubox: Both file and folder methods actually use the same initializers, just with a different string for itemType. I just reviewed the test cases and I think we have pretty good coverage there.

crousselle (approved):

PR #385, opened 2017-04-28
IOS-14005 - Migrate Recents API request to use 'next_marker' since 'm…

irvingruan:
…arker' will be deprecated.

### DIFF
--- a/BoxContentSDK/BoxContentSDK/Requests/BOXRecentItemsRequest.m
+++ b/BoxContentSDK/BoxContentSDK/Requests/BOXRecentItemsRequest.m
@@ -41,7 +41,7 @@
     }
 
     if (self.nextMarker != nil) {
-        queryParameters[BOXAPIParameterKeyMarker] = self.nextMarker;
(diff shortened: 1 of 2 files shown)


boxcla:

BoxContentSDK/BoxContentSDK/Requests/BOXRecentItemsRequest.m
```diff
@@ -41,7 +41,7 @@ - (BOXAPIOperation *)createOperation
     }
 
     if (self.nextMarker != nil) {
-        queryParameters[BOXAPIParameterKeyMarker] = self.nextMarker;
```
adempsey, line 44: Should we mark the `BOXAPIParameterKeyMarker` key as deprecated too to prevent confusion?

irvingruan: Good point.

adempsey (approved):

PR #387, opened 2017-05-03
Fix warnings for iOS 10.3

adempsey:
Also some cleanup

### DIFF
--- a/BoxContentSDK/BoxContentSDK/AppToApp/BOXAppToAppAnnotationBuilder.m
+++ b/BoxContentSDK/BoxContentSDK/AppToApp/BOXAppToAppAnnotationBuilder.m
@@ -10,7 +10,7 @@
 #import "BOXAppToAppApplication.h"
 
 #import "BOXLog.h"
-#import "NSString+BoxURLHelper.h"
+#import "NSString+BOXURLHelper.h"
 
 
 @interface BOXAppToAppAnnotationBuilder()
@@ -103,7 +103,7 @@
 
     if (result != nil)
     {
-        result = [result stringByReplacingPercentEscapesUsingEncoding:NSUTF8StringEncoding];
+        result = [result stringByRemovingPercentEncoding];
         [info removeObjectForKey:key];
     }
 
@@ -156,7 +156,7 @@
 {
     NSString *result = string;
     // percent de-encode the string
-    result = [result stringByReplacingPercentEscapesUsingEncoding:NSUTF8StringEncoding];
+    result = [result stringByRemovingPercentEncoding];
 
     return result;
 }
--- a/BoxContentSDK/BoxContentSDK/BOXURLRequestSerialization.m
+++ b/BoxContentSDK/BoxContentSDK/BOXURLRequestSerialization.m
@@ -1393,7 +1393,7 @@ typedef enum {
         return nil;
     }
 
-    self.format = (NSPropertyListFormat)[[decoder decodeObjectOfClass:[NSNumber class] forKey:NSStringFromSelector(@selector(format))] unsignedIntegerValue];
+    self.format = (NSPropertyListFormat)[[decoder decodeObjectOfClass:[NSNumber class] forKey:NSStringFromSelector(@selector((format)))] unsignedIntegerValue];
     self.writeOptions = [[decoder decodeObjectOfClass:[NSNumber class] forKey:NSStringFromSelector(@selector(writeOptions))] unsignedIntegerValue];
 
     return self;
@@ -1402,7 +1402,7 @@ typedef enum {
 - (void)encodeWithCoder:(NSCoder *)coder {
     [super encodeWithCoder:coder];
 
-    [coder encodeInteger:self.format forKey:NSStringFromSelector(@selector(format))];
+    [coder encodeInteger:self.format forKey:NSStringFromSelector(@selector((format)))];
     [coder encodeObject:@(self.writeOptions) forKey:NSStringFromSelector(@selector(writeOptions))];
 }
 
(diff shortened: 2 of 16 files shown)


boxcla:

BoxContentSDK/BoxContentSDK/BOXURLSessionManager.m
```diff
@@ -640,19 +683,32 @@ - (NSURLSessionTask *)backgroundTaskWithRequest:(NSURLRequest *)request orUpload
     return sessionTask;
 }
 
-- (BOOL)persistBackgroundSessionTaskWithSessionId:(NSString *)backgroundSessionId sessionTask:(NSURLSessionTask *)task taskDelegate:(id <BOXURLSessionDownloadTaskDelegate>)taskDelegate userId:(NSString *)userId associateId:(NSString *)associateId error:(NSError **)error
+- (BOOL)persistBackgroundSessionTaskWithSessionId:(NSString *)backgroundSessionId
+                                      sessionTask:(NSURLSessionTask *)task
+                                     taskDelegate:(id <BOXURLSessionTaskDelegate>)taskDelegate
```
adempsey, line 688: I changed the `taskDelegate` parameter type from `id <BOXURLSessionDownloadTaskDelegate>` to `id <BOXURLSessionTaskDelegate>`. It seems this was intended to use the generic protocol, I don't see any logic specific to downloads here. Let me know if I'm mistaken.

BoxContentSDK/BoxContentSDK/AppToApp/BOXAppToAppAnnotationBuilder.m
```diff
@@ -103,7 +103,7 @@ + (NSString *)stringByDestructivelyParsingInfo:(NSMutableDictionary *)info forKe
 
     if (result != nil)
     {
-        result = [result stringByReplacingPercentEscapesUsingEncoding:NSUTF8StringEncoding];
+        result = [result stringByRemovingPercentEncoding];
```
boxtcarpel, line 106: Thanks, the deprecation warnings were driving me crazy!

BoxContentSDK/BoxContentSDK/BOXURLRequestSerialization.m
```diff
@@ -1393,7 +1393,7 @@ - (instancetype)initWithCoder:(NSCoder *)decoder {
         return nil;
     }
 
-    self.format = (NSPropertyListFormat)[[decoder decodeObjectOfClass:[NSNumber class] forKey:NSStringFromSelector(@selector(format))] unsignedIntegerValue];
+    self.format = (NSPropertyListFormat)[[decoder decodeObjectOfClass:[NSNumber class] forKey:NSStringFromSelector(@selector((format)))] unsignedIntegerValue];
```
boxtcarpel, line 1396: Why is this needed? It isn't needed for 'writeOptions' on the next line for example...

BoxContentSDK/BoxContentSDK/BOXURLSessionManager.m
boxtcarpel: downloads and uploads can be background tasks, others cannot. I'm not sure taskDelegate is the intent here... @thuyBox do you remember?

BoxContentSDK/BoxContentSDK/Operations/BOXStreamOperation.m
```diff
@@ -22,8 +22,6 @@ @interface BOXStreamOperation () <BOXURLSessionDownloadTaskDelegate>
 
 @property (nonatomic, readwrite, assign) unsigned long long bytesReceived;
 
-- (void)writeDataToOutputStream;
```
boxtcarpel, line 25: unimplemented?

BoxContentSDK/BoxContentSDK/Operations/BOXStreamOperation.h
```diff
@@ -96,12 +89,6 @@ typedef void (^BOXAPIStreamProgressBlock)(NSData *data, long long expectedTotalB
 /** @name Overridden methods */
 
 /**
- * In addition to calling [super]([BOXAPIAuthenticatedOperation prepareAPIRequest]), schedule outputStream
- * in the current run loop and set `outputStream.delegate` to `self`.
- */
-- (void)prepareAPIRequest;
```
boxtcarpel, line 102: unimplemented? Or already defined in a superclass?

boxtcarpel (approved):

BoxContentSDK/BoxContentSDK/Operations/BOXStreamOperation.m
adempsey: Yes

BoxContentSDK/BoxContentSDK/Operations/BOXStreamOperation.h
adempsey: Already defined in superclass, it was redeclared here with no actual override.

BoxContentSDK/BoxContentSDK/BOXURLRequestSerialization.m
adempsey: Several classes expose methods named `format` with different return types, so just using `selector` here is ambiguous as to which method it's actually referring to. But in this case it seems as though we're just using it to get the property name, so I don't think this is causing any problems. The extra layer of parentheses tells the compiler to ignore the warning.

boxtcarpel (approved):

BoxContentSDK/BoxContentSDK/BOXURLSessionManager.m
thuyBox: correct, it should have been BOXURLSessionTaskDelegate

PR #392, opened 2017-05-09
Use BOXURLRequestSerialization to escape query parameters

adempsey:


boxcla:

boxtcarpel:
For getting the query dictionary if we're not ready to move to NSURLComponents, can use: NSURL+BOXURLHelper
- (NSDictionary *)box_queryDictionary;
+ (NSDictionary *)box_queryDictionaryWithString:(NSString *)queryString;

For escaping URLs without using NSURLComponents (if we're not ready to move to it on a specific component), you can use: BOXURLRequestSerialization BOXPercentEscapedStringFromString, it also has BOXQueryStringPair ...

boxtcarpel (approved):

scottliubox (approved):

PR #394, opened 2017-05-10
fix share extension not showing up as a share option after carthage b…

thuyBox:
…ranch merge


boxcla:

scottliubox (approved):

PR #395, opened 2017-05-12
Use deprecation attribute instead of Headerdoc comment for marker par…

adempsey:
…amter

The previous format was generating a warning. Using `__attribute__` allows for warnings in code completion.


boxcla:

jdizoglio (approved):

PR #400, opened 2017-05-18
A 403 error from a tokens request should trigger a logout.

RicoYao:


boxcla:

BoxContentSDK/BoxContentSDK/Operations/BOXAPIOAuth2ToJSONOperation.m
```diff
@@ -95,5 +96,19 @@ - (void)performCompletionCallback
     [[NSNotificationCenter defaultCenter] postNotificationName:BOXAuthOperationDidCompleteNotification object:self];
 }
 
+- (BOOL)shouldErrorTriggerLogout:(NSError *)error
```
adempsey, line 99: Can we add comments here to explain why we need to do this override?

BoxContentSDK/BoxContentSDKTests/BOXRequestTests.m
```diff
@@ -184,6 +186,43 @@ - (void)test_that_account_deactivated_400_error_triggers_logout_notification
...
+    
+    NSHTTPURLResponse *URLResponse = [self cannedURLResponseWithStatusCode:403 responseData:nil];
+    [self setCannedURLResponse:URLResponse cannedResponseData:nil forRequest:request];
+    request.operation = [[BOXAPIJSONOperation alloc] initWithURL:[request URLWithResource:nil ID:nil subresource:nil subID:nil] HTTPMethod:BOXAPIHTTPMethodGET body:nil queryParams:nil session:request.queueManager.session];
+
+    id operationMock = [OCMockObject partialMockForObject:request.operation];
+    [[operationMock reject] sendLogoutNotification];
+    
+    XCTestExpectation *expectation = [self expectationWithDescription:@"expectation"];
+    [request performRequestWithCompletion:^{
```
adempsey, line 219: shouldn't we check the status code of our response? Might just not know how to read this correctly, how does this verify that the request was successful?

BoxContentSDK/BoxContentSDK/Operations/BOXAPIOAuth2ToJSONOperation.m
RicoYao: Will do.

BoxContentSDK/BoxContentSDKTests/BOXRequestTests.m
RicoYao: That's a good point. Will add: 
XCTAssertEqual(403, request.operation.HTTPResponse.statusCode);

adempsey (approved):

boxtcarpel:
Nice! What about App Users, do we need handle an equivalent situation for BOXAPIAppUsersAuthOperation? @RicoYao

RicoYao:
@boxtcarpel I don't think we can make any such assumptions for app users. That contract is entirely controlled by the developer, and Box servers have no ability to get specific errors to the client since there is no tokens request.

PR #401, opened 2017-05-19
fix ContentSDK not found running sample app on device

thuyBox:


boxcla:

scottliubox (approved):

PR #402, opened 2017-05-19
cache app-level request information for Sample App's extension's back…

thuyBox:
…ground tasks to allow app to reconnect


boxcla:

PR #403, opened 2017-05-20
Certain files only partially downloaded

adempsey:


boxcla:

RicoYao:
@thuyBox I think you found and fixed this issue based on a different root cause? Can we just close this (without merging)?

PR #404, opened 2017-05-25
Fix an issue that could result in incomplete files

boxtcarpel:
Downloaded files could end up starting with a bunch of 0 bytes
instead of the actual data due to a timing issue. This change
to the correct order of operations resolves the problem.


boxcla:

thuyBox:
discussion offline

thuyBox:
+2

PR #405, opened 2017-06-01
Fix an issue with a method mismatch on checking cache client support

boxtcarpel:
The cache client method check was using the folder method instead
of the file variant.


boxcla:

scottliubox (approved):

PR #408, opened 2017-06-09
Support for early access to the operation and session task from a request

boxtcarpel:
Introducing a 'prepare' method for requests and operations, such that the
underlying session task could be accessed before an operation is enqueued
and contain the complete information.

### DIFF
--- a/BoxContentSDK/BoxContentSDK/Operations/BOXAPIOperation.m
+++ b/BoxContentSDK/BoxContentSDK/Operations/BOXAPIOperation.m
@@ -205,6 +205,29 @@ static BOOL BoxOperationStateTransitionIsValid(BOXAPIOperationState fromState, B
     BOXAbstract();
 }
 
+- (void)prepareOperation
+{
+    if (![self isCancelled]) {
+        if (self.sessionTask == nil) {
+            @synchronized(self.session)
+            {
+                //Note: if sessionTask exists, we cannot change its API request
+                //make sure you recreate sessionTask with the new API request if needed
+                [self prepareAPIRequest];
+                self.accessToken = self.session.accessToken;
+            }
+            NSError *error = nil;
+            self.sessionTask = [self createSessionTaskWithError:&error];
+            if (error != nil) {
+                BOXLog(@"BOXAPIOperation %@ failed to create session task to prepare to execute API request", self);
+                NSMutableDictionary *userInfo = [NSMutableDictionary dictionary];
+                [userInfo setObject:error forKey:NSUnderlyingErrorKey];
+                self.error = [NSError errorWithDomain:BOXContentSDKErrorDomain code:BOXContentSDKURLSessionFailToCreateSessionTask userInfo:userInfo];
+            }
+        }
+    }
+}
+
 #pragma mark - Process API call results
 - (void)processResponseData:(NSData *)data
 {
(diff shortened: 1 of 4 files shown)


boxcla:

BoxContentSDK/BoxContentSDK/Operations/BOXAPIOperation.m
```diff
@@ -205,6 +205,23 @@ - (void)prepareAPIRequest
...
+        if (self.sessionTask == nil) {
+            @synchronized(self.session)
+            {
+                //Note: if sessionTask exists, we cannot change its API request
+                //make sure you recreate sessionTask with the new API request if needed
+                [self prepareAPIRequest];
+                self.accessToken = self.session.accessToken;
+            }
+            NSError *error = nil;
+            self.sessionTask = [self createSessionTaskWithError:&error];
```
thuyBox, line 220: if there's an error creating session task, we should execute logic of line 374 to 377 to set self.error accordingly. We potentially have a valid self.sessionTask but there's other error setting up session task (in background tasks, we need to set up its cache before the task starts)

boxtcarpel: Done.

BoxContentSDK/BoxContentSDK/Operations/BOXAPIOperation.m
```diff
@@ -335,18 +358,18 @@ - (void)executeSessionTask
...
-    if (![self isCancelled])
-    {
-        @synchronized(self.session)
-        {
-            //Note: if sessionTask exists, we cannot change its API request
-            //make sure you recreate sessionTask with the new API request if needed
-            [self prepareAPIRequest];
-            self.accessToken = self.session.accessToken;
+    if (![self isCancelled]) {
+        if (self.sessionTask == nil) {
```
thuyBox, line 362: minor: not sure why we need this check explicitly?

BoxContentSDK/BoxContentSDK/Requests/BOXRequest.m
```diff
@@ -78,6 +78,13 @@ - (BOXAPIOperation *)createOperation
     return nil;
 }
 
+
+- (void)prepareOperation
+{
+    [self.operation.APIRequest setValue:[self userAgent] forHTTPHeaderField:@"User-Agent"];
+    [self.operation prepareOperation];
```
thuyBox, line 85: will we have a separate commit to expose sessionTask from BOXRequest?

thuyBox:
+2

BoxContentSDK/BoxContentSDK/Requests/BOXRequest.m
boxtcarpel: I don't think we need to... prepareOperation is exposed on request, and sessionTask is exposed on the operation that is exposed by the request.
Expected usage is like this:
           let request = BOXContentClient.default().fileUploadNewVersionRequestInBackground(withFileID:...)
            request?.requestAllFileFields = true
            request?.prepareOperation()
            if let task = request?.operation.sessionTask {
                NSFileProviderManager.default.register(task, ...

BoxContentSDK/BoxContentSDK/Operations/BOXAPIOperation.m
boxtcarpel: To avoid doing that work twice, now that it might actually get done on 'prepareOp'

thuyBox:
+2

PR #409, opened 2017-06-12
IOS-14136 - User gets logged out sometimes upon logging in.

irvingruan:
We should be returning NO when we want to use the redirect, otherwise, it will
trigger two auth calls and potentially re-use the same auth code twice.


boxcla:

boxtcarpel (approved):
+2 but since we did it together, lets make sure other folks approve the PR as well ;)

RicoYao:
+2

PR #413, opened 2017-06-20
Curly brackets need to be encoded as they are explicitly disallowed in URL (RFC 1738)

RicoYao:
Internal ticket: IOS-14153


boxcla:

boxtcarpel (approved):

PR #419, opened 2017-07-27
Lay down the groundwork for Auth Token Downscoping Requests

boxtcarpel:
Also some import cleanups

### DIFF
--- a/BoxContentSDK/BoxContentSDK/Requests/BOXBookmarkCommentsRequest.m
+++ b/BoxContentSDK/BoxContentSDK/Requests/BOXBookmarkCommentsRequest.m
@@ -8,6 +8,7 @@
 
 #import "BOXAPIJSONOperation.h"
 #import "BOXComment.h"
+#import "BOXDispatchHelper.h"
 
 @interface BOXBookmarkCommentsRequest ()
 
--- a/BoxContentSDK/BoxContentSDK/Requests/BOXCollectionItemsRequest.m
+++ b/BoxContentSDK/BoxContentSDK/Requests/BOXCollectionItemsRequest.m
@@ -6,6 +6,7 @@
 #import "BOXRequest_Private.h"
 #import "BOXCollectionItemsRequest.h"
 #import "BOXAPIJSONOperation.h"
+#import "BOXDispatchHelper.h"
 
 @interface BOXCollectionItemsRequest ()
 
--- a/BoxContentSDK/BoxContentSDK/Requests/BOXRequest_Private.h
+++ b/BoxContentSDK/BoxContentSDK/Requests/BOXRequest_Private.h
@@ -6,12 +6,13 @@
 #import "BOXAPIDataOperation.h"
 #import "BOXAPIJSONOperation.h"
 #import "BOXAPIJSONPatchOperation.h"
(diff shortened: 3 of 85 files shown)


boxcla:

BoxContentSDK/BoxContentSDK/Requests/BOXBookmarkCommentsRequest.m
```diff
@@ -8,6 +8,7 @@
 
 #import "BOXAPIJSONOperation.h"
 #import "BOXComment.h"
+#import "BOXDispatchHelper.h"
```
jdizoglio, line 11: Should this get imported in base class BoxRequest?

BoxContentSDK/BoxContentSDK/Requests/BOXCollectionItemsRequest.m
```diff
@@ -6,6 +6,7 @@
 #import "BOXRequest_Private.h"
 #import "BOXCollectionItemsRequest.h"
 #import "BOXAPIJSONOperation.h"
+#import "BOXDispatchHelper.h"
```
jdizoglio, line 9: Whey do we have these new class imports and no reference within the source?

BoxContentSDK/BoxContentSDK/Requests/BOXRequest_Private.h
```diff
@@ -71,6 +72,12 @@
                                  successBlock:(BOXAPIHeaderSuccessBlock)successBlock
                                  failureBlock:(BOXAPIHeaderFailureBlock)failureBlock;
 
+- (BOXAPIOAuth2ToJSONOperation *)authOperationWithURL:(NSURL *)URL
```
jdizoglio, line 75: If we're including in the contentSDK, does 'private' make sense? Maybe BOXRequest+Authentication?

BoxContentSDK/BoxContentSDK/Requests/BOXBookmarkCommentsRequest.m
boxtcarpel: We discussed offline, the base class previously imported this in the .h, even though it wasn't used or needed there. These subclasses who use it should import it (since some may not).

BoxContentSDK/BoxContentSDK/Requests/BOXCollectionItemsRequest.m
boxtcarpel: This is because it used to get brought it implicitly from the .h of the super. Now that it's being imported explicitly only when needed, everything else continues to work as-is.

BoxContentSDK/BoxContentSDK/Requests/BOXRequest_Private.h
boxtcarpel: Do you mean for this header? This is basically where a lot of the inner-working pieces get defined, which aren't generally needed (so not in the public .h) but may be used for advanced configuration, etc.
We can find an even better place for this stuff, but currently the convention is that these operation setup helpers all live in here. What do you think?

jdizoglio (approved):

PR #421, opened 2017-07-30
Allow to set specific list of fields to download in items requests

nikinapi:
Existing interface for requests like `get folder items` is not very convenient when a small subset of item fields is required. 
I suggest adding `fieldsToInclude` property to `BOXItemArrayRequest` as an opposite to conjunction of `requestAllItemFields` and `fieldsToExclude`. This property is used when `requestAllItemFields == NO` and acts similar to `fields` parameter in corresponding API requests.


boxcla:
Verified that @nikinapi has signed the CLA. Thanks for the pull request!

PR #422, opened 2017-08-08
Cache client not set for paginated full folder item requests

adempsey:


boxcla:

jlawton (approved):
👍

BoxContentSDK/BoxContentSDK/Requests/BOXFolderItemsRequest.m
```diff
@@ -179,6 +179,7 @@ - (void)performPaginatedRequestWithCached:(BOXItemArrayCompletionBlock)cacheBloc
                                   inRange:(NSRange)range
 {
     BOXFolderPaginatedItemsRequest *paginatedRequest = [[BOXFolderPaginatedItemsRequest alloc] initWithFolderID:self.folderID inRange:range];
+    paginatedRequest.cacheClient = self.cacheClient;
```
boxtcarpel, line 182: Is the behavior correct? Does it call the cache client with the expected sub range?

adempsey: Yes. `BOXFolderPaginatedItemsRequest` uses a different callback method than this class, `cacheFolderPaginatedItemsRequest:withItems:limit:offset:error:`

PR #424, opened 2017-08-12
IOS-14350 [BOXContentSDK] add support for Folder Info Request backgro…

jdizoglio:
…unded

### DIFF
--- a/BoxContentSDK/BoxContentSDK/Clients/BOXContentClient+Folder.m
+++ b/BoxContentSDK/BoxContentSDK/Clients/BOXContentClient+Folder.m
@@ -30,6 +30,18 @@
     return request;
 }
 
+- (BOXFolderRequest *)folderInfoRequestWithID:(NSString *)folderID
+                                  associateId:(NSString *)associateId
+{
+    BOXFolderRequest *request = nil;
+    request = [[BOXFolderRequest alloc] initWithFolderID:folderID associateId:associateId];
+    request.requestDirectoryPath = self.tempCacheDir;
+    
+    [self prepareRequest:request];
+    
+    return request;
+}
+
 - (BOXFolderCreateRequest *)folderCreateRequestWithName:(NSString *)folderName
                                          parentFolderID:(NSString *)parentFolderID
 {
@@ -45,6 +57,8 @@
     BOXFolderCreateRequest *request = [[BOXFolderCreateRequest alloc] initWithFolderName:folderName
                                                                           parentFolderID:parentFolderID
                                                                              associateId:associateId];
+    request.requestDirectoryPath = self.tempCacheDir;
+    
     [self prepareRequest:request];
     
     return request;
--- a/BoxContentSDK/BoxContentSDK/Requests/BOXFolderRequest.m
+++ b/BoxContentSDK/BoxContentSDK/Requests/BOXFolderRequest.m
@@ -17,22 +17,50 @@
 @property (nonatomic, readwrite, strong) NSString *folderID;
 @property (nonatomic, readwrite, assign) BOOL isTrashed;
 
+/// Properties related to Background tasks
+
+/**
+ Check if the request can executre on background. Requires valid associateId and requestDirectoryPath
+ 
+ @return BOOL Yes for can preform on background
+ */
+- (BOOL)shouldPerformBackgroundOperation;
+
+/**
+ Caller provided unique ID to execute the request as a NSURLSession background task
+ */
+@property (nonatomic, readwrite, copy) NSString *associateId;
+
 @end
 
 @implementation BOXFolderRequest
 
 - (instancetype)initWithFolderID:(NSString *)folderID
 {
-    return [self initWithFolderID:folderID isTrashed:NO];
+    return [self initWithFolderID:folderID associateId:nil];
+}
+
+- (instancetype)initWithFolderID:(NSString *)folderID
+                     associateId:(NSString *)associateId
+{
+    return [self initWithFolderID:folderID isTrashed:NO associateId: associateId];
 }
 
 - (instancetype)initWithFolderID:(NSString *)folderID isTrashed:(BOOL)isTrashed
 {
+    return [self initWithFolderID:folderID isTrashed:isTrashed associateId: nil];
+}
+
+- (instancetype)initWithFolderID:(NSString *)folderID
+                       isTrashed:(BOOL)isTrashed
+                     associateId:(NSString *)associateId
+{
     if (self = [super init]) {
         _folderID = folderID;
         _isTrashed = isTrashed;
+        _associateId = associateId;
     }
-
+    
     return self;
 }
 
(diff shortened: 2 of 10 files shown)


boxcla:

BoxContentSDK/BoxContentSDK/Requests/BOXFolderRequest.m
```diff
@@ -59,64 +80,136 @@ - (BOXAPIOperation *)createOperation
...
+                                                             HTTPMethod:BOXAPIHTTPMethodGET
+                                                  queryStringParameters:queryParameters
+                                                         bodyDictionary:nil
+                                                           successBlock:nil
+                                                           failureBlock:nil
+                                                            associateId:self.associateId];
+        
+        NSString *requestDirectory = [BOXContentClient defaultClient].tempCacheDir;
+        NSString *destinationPath = [requestDirectory stringByAppendingPathComponent:self.associateId];
+        dataOperation.destinationPath = destinationPath;
```
thuyBox: should set up if-non-match header similarly to the else block

BoxContentSDK/BoxContentSDK/Requests/BOXFolderRequest.m
```diff
@@ -59,64 +80,136 @@ - (BOXAPIOperation *)createOperation
...
+    if (self.associateId != nil) {
         BOOL isMainThread = [NSThread isMainThread];
-        BOXAPIJSONOperation *folderOperation = (BOXAPIJSONOperation *)self.operation;
-
-        folderOperation.success = ^(NSURLRequest *request, NSHTTPURLResponse *response, NSDictionary *JSONDictionary) {
-            BOXFolder *folder = [[BOXFolder alloc] initWithJSON:JSONDictionary];
-
-            [self.sharedLinkHeadersHelper storeHeadersFromAncestorsIfNecessaryForItemWithID:folder.modelID
+        
+        BOXAPIDataOperation *fileOperation = (BOXAPIDataOperation *)self.operation;
```
thuyBox: folderOperation?

BoxContentSDK/BoxContentSDK/Requests/BOXFolderRequest.m
```diff
@@ -59,64 +80,136 @@ - (BOXAPIOperation *)createOperation
...
-
-    [self addSharedLinkHeaderToRequest:JSONoperation.APIRequest];
-
-    return JSONoperation;
 }
 
 - (void)performRequestWithCompletion:(BOXFolderBlock)completionBlock
 {
-    if (completionBlock) {
+    if (self.associateId != nil) {
```
thuyBox: we should keep the logic that we don't execute if completionBlock is not valid

BoxContentSDK/BoxContentSDK/Requests/BOXFolderRequest.m
jdizoglio: That was a miss. I think I've been working to much, making simple mistakes.

BoxContentSDK/BoxContentSDK/Requests/BOXFolderRequest.m
jdizoglio: done

BoxContentSDK/BoxContentSDK/Requests/BOXFolderRequest.m
jdizoglio: done

BoxContentSDK/BoxContentSDK/Clients/BOXContentClient+Folder.m
```diff
@@ -30,6 +30,16 @@ - (BOXFolderRequest *)folderInfoRequestWithID:(NSString *)folderID
     return request;
 }
 
+- (BOXFolderRequest *)folderInfoRequestWithID:(NSString *)folderID
+                                  associateId:(NSString *)associateId
+{
+    BOXFolderRequest *request = nil;
+    request = [[BOXFolderRequest alloc] initWithFolderID:folderID];
```
thuyBox: set associateId?

BoxContentSDK/BoxContentSDK/Requests/BOXFolderRequest.m
```diff
@@ -17,22 +18,42 @@ @interface BOXFolderRequest ()
...
+
+- (instancetype)initWithFolderID:(NSString *)folderID
+                     associateId:(NSString *)associateId
+{
+    return [self initWithFolderID:folderID isTrashed:NO associateId: associateId];
 }
 
 - (instancetype)initWithFolderID:(NSString *)folderID isTrashed:(BOOL)isTrashed
 {
+    return [self initWithFolderID:folderID isTrashed:NO associateId: nil];
```
thuyBox: isTrashed instead of NO

jdizoglio: done

BoxContentSDK/BoxContentSDK/Clients/BOXContentClient+Folder.m
jdizoglio: done

BoxContentSDK/BoxContentSDK/Requests/BOXFolderRequest.m
```diff
@@ -3,6 +3,7 @@
 //  BoxContentSDK
 //
 
+#import "BOXContentClient_Private.h"
```
boxtcarpel: We should think twice before doing something like this.

It's not good for a class to work it's way upwards. I'm not sure we want to create this dependency. The request shouldn't have to know about the content client...
Is this really unavoidable?

BoxContentSDK/BoxContentSDK/Requests/BOXFolderRequest.m
```diff
@@ -59,64 +80,140 @@ - (BOXAPIOperation *)createOperation
...
+    if (self.associateId != nil) {
+        BOXAPIDataOperation *dataOperation = [self dataOperationWithURL:URL
+                                                             HTTPMethod:BOXAPIHTTPMethodGET
+                                                  queryStringParameters:queryParameters
+                                                         bodyDictionary:nil
+                                                           successBlock:nil
+                                                           failureBlock:nil
+                                                            associateId:self.associateId];
+        
+        NSString *requestDirectory = [BOXContentClient defaultClient].tempCacheDir;
```
boxtcarpel: Also, this makes an incorrect assumption that the default client is always the one used. This kind of indicator confirms to us that this is not the right thing and we should have the target path passed in instead.

BoxContentSDK/BoxContentSDK/Requests/BOXFolderRequest.m
```diff
@@ -59,64 +80,140 @@ - (BOXAPIOperation *)createOperation
...
+                    }
+                }
+                
+                if ([[NSFileManager defaultManager] fileExistsAtPath: weakFileOperation.destinationPath]) {
+                    NSError *error = nil;
+                    [[NSFileManager defaultManager] removeItemAtPath:weakFileOperation.destinationPath error:&error];
+                }
+                
+                [BOXDispatchHelper callCompletionBlock:^{
+                    completionBlock(folder, nil);
```
boxtcarpel: We wouldn't return an error under any scenario here? (successful network call, but somehow mangled output for example)

thuyBox: mangled output would be folder == nil? Yeah, we talked briefly. We could return a content sdk type of error for not being able to parse output or so.

BoxContentSDK/BoxContentSDK/Requests/BOXFolderRequest.m
```diff
@@ -59,64 +80,140 @@ - (BOXAPIOperation *)createOperation
...
-        folderOperation.success = ^(NSURLRequest *request, NSHTTPURLResponse *response, NSDictionary *JSONDictionary) {
-            BOXFolder *folder = [[BOXFolder alloc] initWithJSON:JSONDictionary];
-
-            [self.sharedLinkHeadersHelper storeHeadersFromAncestorsIfNecessaryForItemWithID:folder.modelID
+        if (self.associateId != nil) {
+            BOOL isMainThread = [NSThread isMainThread];
+            
+            BOXAPIDataOperation *folderOperation = (BOXAPIDataOperation *)self.operation;
+            
+            __weak BOXAPIDataOperation *weakFileOperation = folderOperation;
```
boxtcarpel: Name: weakFolderOperation

BoxContentSDK/BoxContentSDK/Requests/BOXFolderRequest.m
jdizoglio: Will open a ticket so that we can revisit across the requests.

BoxContentSDK/BoxContentSDK/Clients/BOXContentClient+Folder.m
```diff
@@ -30,6 +30,16 @@ - (BOXFolderRequest *)folderInfoRequestWithID:(NSString *)folderID
     return request;
 }
 
+- (BOXFolderRequest *)folderInfoRequestWithID:(NSString *)folderID
+                                  associateId:(NSString *)associateId
+{
+    BOXFolderRequest *request = nil;
+    request = [[BOXFolderRequest alloc] initWithFolderID:folderID associateId:associateId];
```
boxtcarpel, line 37: Set the path in here?

BoxContentSDK/BoxContentSDK/Requests/BOXFolderRequest.m
boxtcarpel: Can remove this now, right?

BoxContentSDK/BoxContentSDK/Clients/BOXContentClient+Folder.m
boxtcarpel: And do it for folder create, etc as well...

jdizoglio: done

boxtcarpel (approved):

PR #425, opened 2017-08-13
IOS-14392: fix crashes running UI code in background thread

thuyBox:


boxcla:

BoxContentSDK/BoxContentSDK/OAuth2/BOXAuthorizationViewController.m
```diff
@@ -559,7 +560,10 @@ - (void)URLSession:(NSURLSession *)session task:(NSURLSessionTask *)task didRece
             [alertController addAction:okAction];
             [self presentViewController:alertController animated:YES completion:nil];
         }
-        [self.activityIndicator stopAnimating];
+        BOOL isMainThread = [NSThread isMainThread];
```
jlawton: This *has* to happen on the main thread, so we don't have to know if that's where we're coming from.

Won’t this also crash if the alerts above or below this aren't on the main thread? This will only save us from a crash if we have no alerts to present. It looks like there's an assumption this method gets called on the main thread, but that's false.

thuyBox: updated

jlawton (approved):

BoxContentSDK/BoxContentSDK/OAuth2/BOXAuthorizationViewController.m
```diff
@@ -547,19 +548,24 @@ - (void)URLSession:(NSURLSession *)session task:(NSURLSessionTask *)task didRece
...
+                                                                   style:UIAlertActionStyleDefault
+                                                                 handler:^(UIAlertAction *action) {
+                                                                     [alertController dismissViewControllerAnimated:YES completion:nil];
+                                                                 }];
+                [alertController addAction:okAction];
+                [self presentViewController:alertController animated:YES completion:nil];
+            } onMainThread:YES];
         }
-        [self.activityIndicator stopAnimating];
+        [BOXDispatchHelper callCompletionBlock:^{
```
boxtcarpel, line 566: why is this using dispatch helper completion block? It's just a simple UI call, we should push it to main using a simple dispatch_async... no?

BoxContentSDK/BoxContentSDK/OAuth2/BOXAuthorizationViewController.m
```diff
@@ -706,12 +712,13 @@ - (void)URLSession:(NSURLSession *)session task:(NSURLSessionTask *)task didComp
...
         [self setWebViewCanBeUsedDirectly:YES forHost:task.currentRequest.URL.host];
         [self setWebViewCanBeUsedDirectly:YES forHost:[self.connectionResponse URL].host];
-        [(UIWebView *)self.view loadData:self.connectionData
-                                MIMEType:[self.connectionResponse MIMEType]
-                        textEncodingName:[self.connectionResponse textEncodingName]
-                                 baseURL:[self.connectionResponse URL]];
-
-        self.connectionResponse = nil;
+        [BOXDispatchHelper callCompletionBlock:^{
+            [(UIWebView *)self.view loadData:self.connectionData
```
boxtcarpel, line 716: same, this isn't a completion block call, it's just pushing some UI code to the main thread (always, regardless of whether things originally got called from main or not - which is what the dispatch helper is there to help with I believe)

BoxContentSDK/BoxContentSDK/OAuth2/BOXAuthorizationViewController.m
```diff
@@ -547,19 +548,24 @@ - (void)URLSession:(NSURLSession *)session task:(NSURLSessionTask *)task didRece
...
-
-            UIAlertAction *okAction = [UIAlertAction actionWithTitle:NSLocalizedString(@"OK", @"Label: Allow the user to accept the current condition, often used on buttons to dismiss alerts")
-                                                               style:UIAlertActionStyleDefault
-                                                             handler:^(UIAlertAction *action) {
-                                                                 [alertController dismissViewControllerAnimated:YES completion:nil];
-                                                             }];
-            [alertController addAction:okAction];
-            [self presentViewController:alertController animated:YES completion:nil];
+
+            [BOXDispatchHelper callCompletionBlock:^{
```
boxtcarpel, line 552: same

BoxContentSDK/BoxContentSDK/OAuth2/BOXAuthorizationViewController.m
```diff
@@ -16,6 +16,7 @@
 #import "BOXContentClient+Authentication.h"
 #import "BOXOAuth2Session.h"
 #import "BOXAppUserSession.h"
+#import "BOXDispatchHelper.h"
```
boxtcarpel, line 19: In fact, we shouldn't have a dependency on dispatch helper in here... Is there a reason to do it this way?

BoxContentSDK/BoxContentSDK/OAuth2/BOXAuthorizationViewController.m
thuyBox: done

PR #427, opened 2017-08-14
Add support for file info requests performing in the background

boxtcarpel:


boxcla:

boxtcarpel:
Similar to https://github.com/box/box-ios-sdk/pull/424

BoxContentSDK/BoxContentSDK/Requests/BOXFileRequest.m
```diff
@@ -65,74 +65,105 @@ - (BOXAPIOperation *)createOperation
...
+                                                           failureBlock:nil
+                                                            associateId:self.associateID];
+        NSString *requestDirectory = self.requestDirectoryPath;
+        if (requestDirectory.length == 0) {
+            // Emergency fallback for robustness. The request directory path is expected to be set at request configuration time.
+            requestDirectory = NSTemporaryDirectory();
+        }
+        //FIXME: I believe this should name the operation's destination file more uniquely (for example by
+        // pairing the associate ID with the type of operation and file ID), otherwise you can potentially
+        // get thrashing on the same file for various background actions.
```
boxtcarpel: Need your opinion on this FIXME. Would like to take care of it before pushing this, and I do think we should make the path more unique. What do you think?

jdizoglio: Our header usage states: "associateID Unique ID to associate with the operation to manage when in the background queue." Use of a uuid should take care of uniqueness. These are transient files that we don't allow access to and the number of requests is most likely limited.

The paring itself could also run into the same issue; if a developer calls the file info more than once with the same associate ID. If we want to help the request and enforce uniqueness, we may want to generate the associated ID in the content SDK and the return that to the caller.

BoxContentSDK/BoxContentSDK/Requests/BOXFileRequest.m
```diff
@@ -65,74 +65,105 @@ - (BOXAPIOperation *)createOperation
...
+        BOXAPIDataOperation *dataOperation = [self dataOperationWithURL:URL
+                                                             HTTPMethod:BOXAPIHTTPMethodGET
+                                                  queryStringParameters:queryParameters
+                                                         bodyDictionary:nil
+                                                           successBlock:nil
+                                                           failureBlock:nil
+                                                            associateId:self.associateID];
+        NSString *requestDirectory = self.requestDirectoryPath;
+        if (requestDirectory.length == 0) {
+            // Emergency fallback for robustness. The request directory path is expected to be set at request configuration time.
```
jdizoglio: instead of falling back, I ended up adding a check if self.requestDirectoryPath exists. If it does not exist, I allow the request to be made as non-background.

jdizoglio (approved):

PR #429, opened 2017-08-18
We need to mark the status property as nullable for swift support

jdizoglio:


boxcla:

boxtcarpel (approved):

BoxContentSDK/BoxContentSDK/Models/BOXItem.h
```diff
@@ -148,7 +148,7 @@
  *  Whether this item is deleted or not.
  *  Values include "active", "trashed" if the file has been moved to the trash, and "deleted" if the file has been permanently deleted
  */
-@property (nonatomic, readwrite, strong) NSString *status;
+@property (nonatomic, readwrite, copy, nullable) NSString *status;
```
boxtcarpel, line 151: Should we update the other properties as well?

thuyBox:
+2

BoxContentSDK/BoxContentSDK/Models/BOXItem.h
jdizoglio: Yes, I have a ticket open to review the entire class. I will commit to unblock the upstream caller.

PR #430, opened 2017-08-22
IOS-14409: expose sharedPaginatedRequestData of BOXFolderItemsRequest to link related paginated requests

thuyBox:
and add cache client protocol for completion of BOXFolderItemsRequest

### DIFF
--- a/BoxContentSDK/BoxContentSDK/Protocols/BOXContentCacheClientProtocol.h
+++ b/BoxContentSDK/BoxContentSDK/Protocols/BOXContentCacheClientProtocol.h
@@ -213,6 +213,8 @@
                               withFolder:(BOXFolder *)folder
                                    error:(NSError *)error;
 
+- (void)hasFinishedFolderItemsRequest:(BOXFolderItemsRequest *)request;
+
 #pragma mark - Comments
 
 - (void)cacheBookmarkCommentsRequest:(BOXBookmarkCommentsRequest *)request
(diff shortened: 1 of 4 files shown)


boxcla:

BoxContentSDK/BoxContentSDK/Protocols/BOXContentCacheClientProtocol.h
```diff
@@ -213,6 +213,9 @@
                               withFolder:(BOXFolder *)folder
                                    error:(NSError *)error;
 
+- (void)deleteChildrenOfFolderWithID:(NSString *)folderID
```
boxtcarpel: Should it be associated with a request? This feels odd of place, can we see if it should conform more to the existing pattern or whether it belongs somewhere else?

thuyBox: done

crousselle (approved):

BoxContentSDK/BoxContentSDK/Protocols/BOXContentCacheClientProtocol.h
```diff
@@ -213,6 +213,8 @@
                               withFolder:(BOXFolder *)folder
                                    error:(NSError *)error;
 
+- (void)finishFolderItemsRequest:(BOXFolderItemsRequest *)request;
```
boxtcarpel: Minor: naming, hasFinishedFolderItemsRequest? Just to make it clear that it isn't a call to action, but rather information that an event has occurred.

boxtcarpel (approved):

PR #435, opened 2017-08-28
IOS-14420: allow setting range step for BOXFolderItemsRequest

thuyBox:


boxcla:

BoxContentSDK/BoxContentSDK/Requests/BOXFolderItemsRequest.m
```diff
@@ -43,16 +45,24 @@ - (void)cancel
...
 {
-    NSUInteger rangeStep = 1000;
-    
-    // Root folder has better performance with a smaller page size
-    if ([self.folderID isEqualToString:BOXAPIFolderIDRoot]) {
-        rangeStep = 100;
+    if (_rangeStep == 0) {
+        NSUInteger rangeStep = 1000;
+
+        // Root folder has better performance with a smaller page size
```
boxtcarpel: This isn't unique to root, is it? I think it's just that large/complex folders have better performance with smaller page sizes?

boxtcarpel, line 59: Can mention that this is because it usually will be a larger and more complex folder (it's not just by virtue of being the root folder).

BoxContentSDK/BoxContentSDK/Requests/BOXFolderItemsRequest.m
```diff
@@ -24,6 +24,8 @@ @interface BOXFolderItemsRequest ()
 
 @implementation BOXFolderItemsRequest
 
+@synthesize rangeStep = _rangeStep;
+
 - (instancetype)initWithFolderID:(NSString *)folderID
 {
     if (self = [super init]) {
```
boxtcarpel, line 31: Should probably initialize range step to 0

BoxContentSDK/BoxContentSDK/Requests/BOXFolderItemsRequest.m
```diff
@@ -43,16 +45,24 @@ - (void)cancel
...
-    if ([self.folderID isEqualToString:BOXAPIFolderIDRoot]) {
-        rangeStep = 100;
+    if (_rangeStep == 0) {
+        NSUInteger rangeStep = 1000;
+
+        // Root folder has better performance with a smaller page size
+        if ([self.folderID isEqualToString:BOXAPIFolderIDRoot]) {
+            rangeStep = 100;
+        }
+        _rangeStep = rangeStep;
```
boxtcarpel: Shouldn't set the ivar in this case, I think we should just return the non-0 value we resolve to

boxtcarpel (approved):

BoxContentSDK/BoxContentSDK/Requests/BOXFolderItemsRequest.m
thuyBox: done

BoxContentSDK/BoxContentSDK/Requests/BOXFolderItemsRequest.m
thuyBox: done

PR #437, opened 2017-08-31
Adding an unranked default box value working around the expected unra…

jdizoglio:
…nked value issue


boxcla:

boxtcarpel (approved):

PR #438, opened 2017-08-31
IOS-14445: fix crash for UI activities not running on main thread in …

thuyBox:
…BOXAuthorizationViewController


boxcla:

BoxContentSDK/BoxContentSDK/OAuth2/BOXAuthorizationViewController.m
```diff
@@ -249,19 +249,28 @@ - (void)failAuthenticationWithURLSessionTask:(NSURLSessionTask *)task
...
-        }
-    }
+    BOOL isMainThread = [NSThread isMainThread];
+    dispatch_async(dispatch_get_main_queue(), ^{
+        //dispatch on main thread UI-related activities
+        if (self.view.window) {
+            [self handleConnectionErrorWithError:error message:message];
+            [self.activityIndicator stopAnimating];
+        } else {
+            [BOXDispatchHelper callCompletionBlock:^{
```
boxtcarpel: I don't think there's a need to pick up the dispatch helper here and create a new dependency. There should be no expectation on what thread 'prepareForDismissal' or the completion block get called on. They can be on the main thread and it would be the responsibility of the implementations of those to dispatch high-intensity work off to some worker queue.

boxtcarpel (approved):
Can be simplified by removing the use of the dispatch helper. This whole thing is a UI element (View Controller) after all, it's OK and expected for things to occur on the main queue.

PR #445, opened 2017-09-11
Remove incorrect location for handling temporary upload path cleanup

boxtcarpel:
The cache layer should not be cleaning up the upload asset location when it
is not deterministically the layer where it was created. It's the caller of
the Content Client that creates the upload file and thus owns the responsibility
for cleaning it up, etc.


boxcla:

crousselle (approved):
thx for fixing :/

PR #447, opened 2017-09-15
Fix how the extension gets calculated when accounting for empty filenames

boxtcarpel:


boxcla:

jlawton (approved):

PR #449, opened 2017-09-26
Add caching protocol for BOXUserAvatarRequest.

RicoYao:
- Removed "profile" pic_type option since it does not work.

### DIFF
--- a/BoxContentSDK/BoxContentSDK/Requests/BOXUserAvatarRequest.m
+++ b/BoxContentSDK/BoxContentSDK/Requests/BOXUserAvatarRequest.m
@@ -46,9 +46,6 @@
         case BOXAvatarTypeLarge:
             avatarTypeString = @"large";
             break;
-        case BOXAvatarTypeProfile:
-            avatarTypeString = @"profile";
-            break;
         case BOXAvatarTypeUnspecified:
         default:
             break;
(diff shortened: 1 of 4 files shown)


boxcla:

scottliubox (approved):

BoxContentSDK/BoxContentSDK/Requests/BOXUserAvatarRequest.m
```diff
@@ -70,30 +67,60 @@ - (BOXAPIOperation *)createOperation
...
+- (void)performRequestWithProgress:(BOXProgressBlock)progressBlock
+                            cached:(BOXImageBlock)cacheBlock
+                         refreshed:(BOXImageBlock)refreshBlock
+{
+    if (cacheBlock) {
+        if ([self.cacheClient respondsToSelector:@selector(retrieveCacheForUserAvatarRequest:completion:)]) {
+            [self.cacheClient retrieveCacheForUserAvatarRequest:self
+                                                     completion:cacheBlock];
+        } else {
+            cacheBlock(nil, nil);
```
boxtcarpel, line 79: should have an error for this case, no?

boxtcarpel (approved):

BoxContentSDK/BoxContentSDK/Requests/BOXUserAvatarRequest.m
RicoYao: @boxtcarpel I also found this curious, and agree with you, but it seems that other BOXRequest classes I looked at do not include an error in this case.

RicoYao: Actually I would suggest that the cacheBlock should not even be fired at all in this case. But again though, this convention appears to have been established.

boxtcarpel: Yeah, I think that's fair when the cache client doesn't even handle that case, but if any code was written with the assumption that the cache block always gets called then that would cause them problems, so I think you're right and we'll have to stick to this convention.

With this convention in place, I think we should move to having a particular error to represent 'no cache implementation' but for now it sounds like we have nil, nil. Thanks for looking into it!

boxtcarpel:
+2 LGTM

BoxContentSDK/BoxContentSDK/Requests/BOXUserAvatarRequest.m
scottliubox: Yes, originally it was decided that caching an error doesn't make sense, since we'd never load an error from cache when we fire off a request.

So I agree, we should just not even trigger the cacheBlock at all when we encounter the error. This is applicable to all requests as well.

PR #452, opened 2017-09-28
Make sure we do not immediately logout in case of refresh token being…

scottliubox:
… invalidated by race condition.

### DIFF
--- a/BoxContentSDK/BoxContentSDK/Operations/BOXAPIOAuth2ToJSONOperation.m
+++ b/BoxContentSDK/BoxContentSDK/Operations/BOXAPIOAuth2ToJSONOperation.m
@@ -12,6 +12,8 @@
 #import "BOXContentSDKErrors.h"
 #import "BOXURLRequestSerialization.h"
 #import "BOXAPIOperation_Private.h"
+#import "BOXOAuth2Session.h"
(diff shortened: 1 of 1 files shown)


boxcla:

BoxContentSDK/BoxContentSDK/Operations/BOXAPIOAuth2ToJSONOperation.m
```diff
@@ -100,11 +102,39 @@ - (BOOL)shouldErrorTriggerLogout:(NSError *)error
...
     BOOL shouldLogout = [super shouldErrorTriggerLogout:error];
     
+    BOOL isRefreshTokenInvalid = NO;
+    BOOL isErrorForbidden = NO;
+
+    NSDictionary *errorInfo = [error.userInfo objectForKey:BOXJSONErrorResponseKey];
+    NSString *errorType = [errorInfo objectForKey:BOXAuthURLParameterErrorCodeKey];
+    
+    if ([errorType isEqualToString:BOXAuthTokenRequestErrorInvalidGrant]) {
+        isRefreshTokenInvalid = YES;
```
jlawton: Minor: This could be written cleanly without the `if`. Same below.

jlawton (approved):

PR #453, opened 2017-09-28
Add BOXUserAvatarImageView

RicoYao:
A convenience UIImageView for showing user avatars. It will show a name-based avatar with the user's initials if it cannot fetch a true avatar, or if the user has not set one.

### DIFF
--- a/BoxContentSDK/BoxContentSDK/Views/BOXUserAvatarImageView.m
+++ b/BoxContentSDK/BoxContentSDK/Views/BOXUserAvatarImageView.m
@@ -1,0 +1,271 @@
+//
+//  BOXUserAvatarImageView.m
+//  BoxContentSDK
+//
+//  Created by Rico Yao on 9/27/17.
+//  Copyright © 2017 Box. All rights reserved.
+//
+
+#import "BOXUserAvatarImageView.h"
+#import "BOXUserAvatarRequest.h"
+#import "BOXContentClient+User.h"
+#import "BOXContentCacheClientProtocol.h"
+
+typedef enum {
+    BOXCircularAvatarBackgroundShadeSoftMagenta = 0,
+    BOXCircularAvatarBackgroundShadeSoftCyan,
+    BOXCircularAvatarBackgroundShadeLightRed,
+    BOXCircularAvatarBackgroundShadeModerateGreen,
+    BOXCircularAvatarBackgroundShadeVividViolet,
+    BOXCircularAvatarBackgroundShadeLightOrange,
+    BOXCircularAvatarBackgroundShadeBrightRed,
+    BOXCircularAvatarBackgroundShadeLimeGreen,
+    BOXCircularAvatarBackgroundShadeSoftPink,
+    BOXCircularAvatarBackgroundShadeModerateBlue,
+    BOXCircularAvatarBackgroundShadeDarkOrange,
+    BOXCircularAvatarBackgroundShadeDesaturatedBlue,
+    BOXCircularAvatarBackgroundShadeModerateLimeGreen,
+    BOXCircularAvatarBackgroundShadeBrightYellow,
+    BOXCircularAvatarBackgroundShadeSoftBlue,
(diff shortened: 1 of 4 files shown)


boxcla:

BoxContentSDK/BoxContentSDK/Views/BOXUserAvatarImageView.m
```diff
@@ -0,0 +1,265 @@
...
+
+- (void)renderImage:(UIImage *)image
+{
+    dispatch_async(dispatch_get_main_queue(), ^{
+        self.image = image;
+        [self setNeedsLayout];
+    });
+}
+
+#pragma mark - Helpers for name-based avatar
```
RicoYao, line 136: Sourced from https://git.dev.box.net/Mobile/box-ios-share-sdk-source/blob/dev/BoxShareSDK/BoxShareSDK/Categories/UIImage%2BBOXCircularAvatar.m

thuyBox: wonder if we could have this new class BOXUserAvatarImageView.m in ShareSDK instead? we want to avoid bringing UI components to content sdk, which would be imported as the base networking layer.

RicoYao: @boxtcarpel and I had some discussions about where to land this, and the problem with sticking it into something like the share SDK is that we'll need to make other components like the Comments SDK dependent on it. Long term, I think we'll merge all the UI SDKs, but until then this seemed like the best place to put it. Happy to discuss further though.

BoxContentSDK/BoxContentSDK/Views/BOXUserAvatarImageView.m
```diff
@@ -0,0 +1,267 @@
...
+        [self render];
+    }
+}
+
+#pragma mark - Private helpers
+
+- (void)render
+{
+    [self.request cancel];
+    self.request = [self.client userAvatarRequestWithID:self.user.modelID type:BOXAvatarTypeLarge];
```
venkatrj: 1. do you want to check for non null user and client objects and skip the flow if they are?
2. what happens if multiple instances of this class are requesting the user avatar for the same user? will it fire off multiple or have some way to check for a pending request and block the rest?

BoxContentSDK/BoxContentSDK/Views/BOXUserAvatarImageView.m
```diff
@@ -0,0 +1,267 @@
...
+    
+    return avatarNameAttributes;
+}
+
+// This function will take the First word and Last word in the name string as initials.
+// Incorrect initials will be shown when a user has a multi-part last name or a suffix
+// since there is no way of distinguishing whether a user has entered a middle name or falls
+// into the above case. If we instead use initials from the first 2 words, we have the issue of
+// showing incorrect initials when the user has entered a middle name.
+- (NSString *)initialsFromName:(NSString *)name
```
venkatrj, line 245: This seems like something that must be extracted into a utility method on the user class itself. isn't it?

RicoYao: Would normally agree, but we try to keep all the 'BOXModel' classes (such as BOXUser) really clean and free of these kinds of util methods. As much as possible we want those to just represent the API response from the server.

BoxContentSDK/BoxContentSDK/Views/BOXUserAvatarImageView.m
RicoYao: 1. Good point on nil-user check. I'll add that. The nil-client check feels a bit unnecessary and it's not something we check for in the rest of the SDK. For most users the client is effectively a singleton so it's never really nil.

2. The requests go into an operation queue where we've established reasonable concurrent execution limits. So it should handle many requests getting fired off pretty well.

PR #454, opened 2017-09-28
Delay the keychain read in the case of a race condition

jlawton:
This should give us enough room to recover from an inter-process token refresh race, without adding significant complexity to the code.

There is some anecdotal evidence that the previous fix will fail in a small fraction of cases, and this should guard against that, though I admit that the chosen delay is arbitrary.

### DIFF
--- a/BoxContentSDK/BoxContentSDK/Operations/BOXAPIOAuth2ToJSONOperation.m
+++ b/BoxContentSDK/BoxContentSDK/Operations/BOXAPIOAuth2ToJSONOperation.m
@@ -118,22 +118,41 @@
         BOXOAuth2Session *session = (BOXOAuth2Session *)self.session;
         
         if ([session isKindOfClass:[BOXOAuth2Session class]]) {
-            NSString *jsonString = [[BOXOAuth2Session keychainItemWrapperForUserWithID:session.user.modelID] objectForKey:(__bridge id)kSecValueData];
-            NSError *error = nil;
-            NSDictionary *dictionary = [NSJSONSerialization JSONObjectWithData:[jsonString dataUsingEncoding:NSUTF8StringEncoding]
-                                                                       options:0
-                                                                         error:&error];
-            NSString *keychainAccessToken = dictionary[keychainAccessTokenKey];
-            
-            // If our keychain token and token in memory are out of sync, we hold off on the logout.
-            if (keychainAccessToken.length > 0 && ![keychainAccessToken isEqualToString:self.accessToken]) {
-                [session restoreSessionWithKeyChainDictionary:dictionary];
+            // We're probably here because we're trying to refresh our access token right around the time it is being
+            // refreshed in another process, and we lost the race. Check the keychain, and only report refresh failure
+            // if it wasn't updated.
+            if ([self updateTokensFromKeychainIntoSession:session]) {
                 shouldLogout = NO;
+            } else {
+                // Pause briefly to allow time for the keychain to be updated (in case we have really unlucky timing).
+                // This only happens when racing to refresh our token, which is infrequent enough that this delay should
+                // be unnoticable to a user.
+                usleep(0.2 * USEC_PER_SEC);
+                if ([self updateTokensFromKeychainIntoSession:session]) {
+                    shouldLogout = NO;
+                }
             }
         }
     }
     
     return shouldLogout;
+}
+
+// Returns whether the session was actually updated by reading the keychain.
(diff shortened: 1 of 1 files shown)


boxcla:

BoxContentSDK/BoxContentSDK/Operations/BOXAPIOAuth2ToJSONOperation.m
```diff
@@ -136,4 +134,22 @@ - (BOOL)shouldErrorTriggerLogout:(NSError *)error
...
+- (BOOL)updateTokensFromKeychainIntoSession:(BOXOAuth2Session *)session
+{
+    NSString *jsonString = [[BOXOAuth2Session keychainItemWrapperForUserWithID:session.user.modelID] objectForKey:(__bridge id)kSecValueData];
+    NSError *error = nil;
+    NSDictionary *dictionary = [NSJSONSerialization JSONObjectWithData:[jsonString dataUsingEncoding:NSUTF8StringEncoding]
+                                                               options:0
+                                                                 error:&error];
+    NSString *keychainAccessToken = dictionary[keychainAccessTokenKey];
+
+    // If our keychain token and token in memory are out of sync, we hold off on the logout.
```
RicoYao: This comment could be improved now that it's been moved into this method.

BoxContentSDK/BoxContentSDK/Operations/BOXAPIOAuth2ToJSONOperation.m
```diff
@@ -118,16 +118,14 @@ - (BOOL)shouldErrorTriggerLogout:(NSError *)error
...
-            // If our keychain token and token in memory are out of sync, we hold off on the logout.
-            if (keychainAccessToken.length > 0 && ![keychainAccessToken isEqualToString:self.accessToken]) {
-                [session restoreSessionWithKeyChainDictionary:dictionary];
+            // We're probably here because we're trying to refresh our access token right around the time it is being
+            // refreshed in another process, and we lost the race. Check the keychain, and only report refresh failure
+            // if it wasn't updated.
+            // Pause briefly to allow time for the keychain to be updated (in case we have really unlucky timing).
+            // This only happens when racing to refresh our token, which is infrequent enough that this delay should
+            // be unnoticable to a user.
+            usleep(0.3 * USEC_PER_SEC);
```
RicoYao: Generally not a fan of the sleep, but if this is where we're headed, can we make it so that we only sleep if we discover that the tokens in memory match the keychain.

RicoYao: Another thing to consider is that If we have multiple calls to shouldErrorTriggerLogout all converging into one thread, these usleeps could stack up. And it could get pretty bad if these are firing on the main thread.

jlawton: Added initial keychain read. These can't stack up, and can't happen on the main thread. There is a dedicated concurrent queue for network operations, where this happens, and token refresh blocks the queue, so it only happens once per token expiration.

RicoYao:
+1

jlawton:
FYI Tested that the app logs out with this change, after a legitimate grant revocation. I don't think that the delay represents a security risk, as killing during the delay period is equivalent to just going offline before token refresh returns, and the window is small.

BoxContentSDK/BoxContentSDK/Operations/BOXAPIOAuth2ToJSONOperation.m
```diff
@@ -118,22 +118,42 @@ - (BOOL)shouldErrorTriggerLogout:(NSError *)error
...
+            // We're probably here because we're trying to refresh our access token right around the time it is being
+            // refreshed in another process, and we lost the race. Check the keychain, and only report refresh failure
+            // if it wasn't updated.
+            if ([self updateTokensFromKeychainIntoSession:session]) {
                 shouldLogout = NO;
+            } else {
+                // Pause briefly to allow time for the keychain to be updated (in case we have really unlucky timing).
+                // This only happens when racing to refresh our token, which is infrequent enough that this delay should
+                // be unnoticable to a user.
+                usleep(0.3 * USEC_PER_SEC);
```
boxtcarpel: any reason for this amount of time?

boxtcarpel: .2?

boxtcarpel (approved):

PR #459, opened 2017-10-16
Adding background tasks to delete and trash file requests

jdizoglio:


boxcla:

BoxContentSDK/BoxContentSDK/Clients/BOXContentClient+File.h
```diff
@@ -110,6 +110,17 @@
 - (BOXFileDeleteRequest *)fileDeleteRequestWithID:(NSString *)fileID;
 
 /**
+ *  Generate a background request to delete a file.
+ *
+ *  @param fileID File ID.
+ *  @param associateID   an Id to associate with this background upload task to reconnect to if needed
```
jlawton: I find this sentence a little awkward to parse. I think we could have fairly consistent language here and below. I see there are other comments along the same lines in the code, which I also find difficult, but we don't have to audit them all for this PR.

Maybe something like `A unique ID used to recover background tasks. Providing an ID associated with a running background task will cause the request to reconnect to that task, rather than making a new one.`

@thuyBox What do you think?

BoxContentSDK/BoxContentSDK/Requests/BOXFileDeleteRequest.m
```diff
@@ -42,52 +63,109 @@ - (BOXAPIOperation *)createOperation
...
-    };
-    [self performRequest];
+    if ([self shouldPerformBackgroundOperation] == YES) {
+        BOOL isMainThread = [NSThread isMainThread];
+        
+        BOXAPIDataOperation *fileOperation = (BOXAPIDataOperation *)self.operation;
+        __weak BOXAPIDataOperation *weakFileOperation = fileOperation;
+        
+        fileOperation.successBlock = ^(NSString *modelID, long long expectedTotalBytes) {
+            NSData *data = [NSData dataWithContentsOfFile:weakFileOperation.destinationPath];
```
jlawton: `data` is never used.

BoxContentSDK/BoxContentSDK/Requests/BOXFileDeleteRequest.m
```diff
@@ -42,52 +63,109 @@ - (BOXAPIOperation *)createOperation
...
+                    completionBlock(nil);
+                } onMainThread:isMainThread];
+            }
+        };
+        
+        fileOperation.failureBlock = ^(NSURLRequest *request, NSHTTPURLResponse *response, NSError *error) {
+            if ([self.cacheClient respondsToSelector:@selector(cacheFileDeleteRequest:error:)]) {
+                [self.cacheClient cacheFileDeleteRequest:self error:error];
+            }
+            if (completionBlock) {
```
jlawton, line 130: Do we not have to delete the destination file here? Is it not created if the server returns an error code, even if the server response has a body?

BoxContentSDK/BoxContentSDK/Requests/BOXFileDeleteRequest.m
```diff
@@ -42,52 +63,109 @@ - (BOXAPIOperation *)createOperation
...
-            [self.cacheClient cacheFileDeleteRequest:self error:error];
-        }
-        if (completionBlock) {
-            [BOXDispatchHelper callCompletionBlock:^{
-                completionBlock(error);
-            } onMainThread:isMainThread];
-        }
-    };
-    [self performRequest];
+    if ([self shouldPerformBackgroundOperation] == YES) {
```
jlawton, line 103: Seems like we could merge the `if` and the `else` blocks, if we just ensure that we're doing the right thing on `destinationPath == nil`.

BoxContentSDK/BoxContentSDK/Requests/BOXFileDeleteRequest.m
jdizoglio: done

BoxContentSDK/BoxContentSDK/Requests/BOXFileDeleteRequest.m
jdizoglio: We have different operation types. I think we can leave this as is and migrate all to a common background operation.

BoxContentSDK/BoxContentSDK/Requests/BOXFileDeleteRequest.m
jdizoglio: Ok, we have it right. No destination file is created when we have an api error. Only created for success.

BoxContentSDK/BoxContentSDK/Clients/BOXContentClient+File.h
```diff
@@ -274,6 +286,17 @@
 - (BOXFileDeleteRequest *)trashedFileDeleteFromTrashRequestWithID:(NSString *)fileID;
 
 /**
+ *  Generate a background to permanently delete a file in the trash.
+ *
+ *  @param fileID File ID.
+ *  @param associateID Unique ID to associate with the operation to manage when in the background queue.
```
jlawton: Maybe the same comment as above.

jlawton (approved):

PR #461, opened 2017-10-20
IOS-14788: do not run request serialization on main thread

thuyBox:


boxcla:

BoxContentSDK/BoxContentSDK/BOXURLRequestSerialization.m
```diff
@@ -471,9 +471,7 @@ - (NSMutableURLRequest *)requestWithMultipartFormRequest:(NSURLRequest *)request
         [inputStream close];
 
         if (handler) {
-            dispatch_async(dispatch_get_main_queue(), ^{
-                handler(error);
-            });
+            handler(error);
```
jlawton, line 471: How was this causing a deadlock? Are we calling `prepareAPIRequest` from the main thread, or is it internal?

BoxContentSDK/BoxContentSDK/BOXURLRequestSerialization.m
```diff
@@ -1210,13 +1208,6 @@ - (NSInteger)readData:(NSData *)data
 }
 
 - (BOOL)transitionToNextPhase {
-    if (![[NSThread currentThread] isMainThread]) {
-        dispatch_sync(dispatch_get_main_queue(), ^{
-            [self transitionToNextPhase];
-        });
-        return YES;
-    }
-
```
jlawton, line 1219: Can this ever be called in a context where there is not a current runloop, or the runloop is ephemeral? It seems that that might be the reason it's dispatching to the main thread, but maybe it's not a concern.

BoxContentSDK/BoxContentSDK/BOXURLRequestSerialization.m
thuyBox: more details are in the ticket. There's no reason for building multipart file to callback onto main thread internally to transfer to next phase. Also, doing this is not expected by the caller which usually calls prepareOperation synchronously, potentially on the main thread, which calls prepareAPIRequest internally, blocked on a background thread which is blocked on main thread, thus, results in deadlock.

BoxContentSDK/BoxContentSDK/BOXURLRequestSerialization.m
thuyBox: the way we're using it in content sdk, we won't actually use runloop. The inputstream we pass in is BOXMultipartBodyStream which overrides scheduleInRunLoop method to be no-op.

jlawton (approved):
Since this works, and is from 3rd party code, I'm happy to leave the `scheduleInRunLoop:` calls in there, even though they appear not to be used.

jlawton (approved):
Yes!

PR #463, opened 2017-10-25
[Major] Fix nullable warnings in BOXContentClient+Folder

scottliubox:


boxcla:

jlawton (approved):

jlawton:
As a rule, we should add the same annotations to the `.m` file.

PR #467, opened 2017-11-07
Unversioned base URL support

DigitalSolomon:


boxcla:
Hi @DigitalSolomon, thanks for the pull request. Before we can merge it, we need you to sign our Contributor License Agreement. You can do so electronically here: http://opensource.box.com/cla

DigitalSolomon:
@boxtcarpel I've updated the code to maintain the non-componentized base URL definitions.

BoxContentSDK/BoxContentSDK/Requests/BOXRequest_Private.h
```diff
@@ -27,6 +27,12 @@
                subresource:(NSString *)subresource
                      subID:(NSString *)subID;
 
+- (NSURL *) URLWithResource:(NSString *)resource
```
boxtcarpel: Minor: I'd move this above the other just to keep the expansion of parameters consistent for when we're reading through it.

DigitalSolomon:
@boxtcarpel Updated version-extracting code to cover more cases. Left comments near base URL definitions to remind about constraints.

BoxContentSDK/BoxContentSDK/Clients/BOXContentClient.m
```diff
@@ -25,6 +25,12 @@
 #import "BOXURLSessionManager.h"
 
 // Default API URLs
+/*
+ NOTE:  Currently when unversioned URLs are derived from the base URL,
+        it removes the last component of the URL path if and only if
+        the component begins with a number.
+        (e.g. "https://api.box.com/2.0" shortens to "https://api.box.com")
```
boxtcarpel, line 32: For clarity (I always assume future-me is dumber than now-me), maybe we can add something like "If we change the base URL to not end with the API version, the change the API version to begin with a non-numerical character - the 'base URL without version' implementation will need to be updated as well." ?
I'm sure you can word that better, but something like that?

BoxContentSDK/BoxContentSDK/Clients/BOXContentClient.m
```diff
@@ -486,6 +493,27 @@ + (NSString *)APIBaseURL
     return staticAPIBaseURL;
 }
 
+//NOTE: Right now we assume that version-portion of the URL string is specified by the last component in the URL path if it also starts with a number
```
boxtcarpel: "...URL path **and** it also starts with a number" ?

BoxContentSDK/BoxContentSDK/Requests/BOXRequest.m
```diff
@@ -48,6 +48,15 @@ - (NSString *)baseURL
     return _baseURL;
 }
 
+- (NSString *)baseURLWithoutVersion
```
boxtcarpel: I don't think this needs to be in the Request base class, right? It's not particularly common, we can just use it if/when we have a concrete request subclass that needs it.

BoxContentSDK/BoxContentSDK/Clients/BOXContentClient.m
```diff
@@ -486,6 +493,27 @@ + (NSString *)APIBaseURL
...
++ (NSString *)APIBaseURLWithoutVersion
+{
+    if (staticAPIBaseURLWithoutVersion.length == 0) {
+        NSString *versionedURL = [BOXContentClient APIBaseURL];
+        NSString *unversionedURL = nil;
+        NSURL *url = [NSURL URLWithString:versionedURL];
+        NSArray<NSString *> *components = [url pathComponents];
+        NSString *lastComponent = [components lastObject];
+        if( [lastComponent length] && isnumber([lastComponent characterAtIndex:0])) {
```
boxtcarpel: length > 0 ?

BoxContentSDK/BoxContentSDK/Clients/BOXContentClient.m
```diff
@@ -486,6 +493,27 @@ + (NSString *)APIBaseURL
...
+            NSString *removeString = [NSString stringWithFormat:@"/%@", lastComponent];
+            unversionedURL = [versionedURL stringByReplacingOccurrencesOfString:removeString withString:@""];
+        }
```
boxtcarpel: You can do this more simply actually. NSString has 'pathComponents' on it directly, as well as 'lastPathComponent'. And once you've verified that our last path component is a version, you can simply do 'stringByDeletingLastPathComponent'. That should work, right?

boxtcarpel (commented):
Mainly one comment about the version removal logic, and I think we can revert the change to BOXRequest.h/m - what do you think?

DigitalSolomon:
Feedback has been addressed, @boxtcarpel

BoxContentSDK/BoxContentSDK/Clients/BOXContentClient.m
```diff
@@ -486,6 +497,26 @@ + (NSString *)APIBaseURL
     return staticAPIBaseURL;
 }
 
+//NOTE: Right now we assume that version-portion of the URL string is specified by the last component in the URL path and it also starts with a number
++ (NSString *)APIBaseURLWithoutVersion
+{
+    if (staticAPIBaseURLWithoutVersion.length == 0) {
+        NSString *versionedURL = [BOXContentClient APIBaseURL];
+        NSString *unversionedURL = nil;
+        NSURL *url = [NSURL URLWithString:versionedURL];
```
boxtcarpel: so, we can skip these 3 lines and just use the NSString lastPathComponent, right?

boxtcarpel (approved):
1 comment

BoxContentSDK/BoxContentSDK/Clients/BOXContentClient.m
```diff
@@ -486,6 +497,24 @@ + (NSString *)APIBaseURL
...
 }
 
+//NOTE: Right now we assume that version-portion of the URL string is specified by the last component in the URL path and it also starts with a number
++ (NSString *)APIBaseURLWithoutVersion
+{
+    if (staticAPIBaseURLWithoutVersion.length == 0) {
+        NSString *versionedURL = [BOXContentClient APIBaseURL];
+        NSString *unversionedURL = nil;
+        NSString *lastComponent = [versionedURL lastPathComponent];
+        if (lastComponent.length > 0 && isnumber([lastComponent characterAtIndex:0])) {
```
Prithvijutur, line 507: wouldn't a regex to match lastComponent to a number work better?

Prithvijutur (approved):
LGMT

boxtcarpel:
test this please

PR #469, opened 2017-11-11
Be tolerant of null values for user created/modified dates.

RicoYao:
We would get those values from the API for special cases such as 'prior collboarator'
and 'anonomyous user'.


boxcla:

jdizoglio (approved):

PR #473, opened 2017-11-29
add new error codes for file info error domain

Prithvijutur:


boxcla:

PR #475, opened 2017-12-13
A hackfix for a race condtion where caching URLSession would fail due…

scottliubox:
… to directory not present.


boxcla:

BoxContentSDK/BoxContentSDK/BOXURLSessionCacheClient.m
```diff
@@ -701,11 +701,26 @@ - (BOOL)cacheBackgroundSessionId:(NSString *)backgroundSessionId
...
     if (success == YES && error == nil) {
         //persist data to onGoingSessionTasks/$backgroundSessionId/$sessionTaskId/$fileType
         NSString *path = [self filePathForBackgroundSessionId:backgroundSessionId sessionTaskId:sessionTaskId type:type];
+        
+        BOOL isDir = NO;
+        
         success = [self cacheAndAttemptToEncryptData:data atPath:path error:&error];
+        
+        if (!success) {
+            // For some reason we the directory that's supposed to be here is sometimes not, so we make sure to create it and try again.
```
thuyBox, line 712: do we have a scenario that trigger this issue? Since looks like we just create the dir at line 703 and we need to re-create the same dir here.

We should use createDirForBackgroundSessionId:sessionTaskId:error method instead of the steps below.

PR #487, opened 2018-01-31
Updating Sample application to add Carthage and a Makefile for convenience. 

jdizoglio:


boxcla:

BoxContentSDKSampleApp/Makefile
```diff
@@ -0,0 +1,30 @@
+CARTHAGE := $(shell command -v carthage 2> /dev/null)
+
+.PHONY: help
```
jlawton: All of these rules are technically .PHONY. Maybe we should mark that, too, to be clean.

BoxContentSDKSampleApp/Makefile
```diff
@@ -0,0 +1,30 @@
...
+
+help:
+	@grep -E '^[a-zA-Z_-]+:.*?## .*$$' $(MAKEFILE_LIST) | sort | awk 'BEGIN {FS = ":.*?## "}; {printf "\033[36m%-30s\033[0m %s\n", $$1, $$2}'
```
jlawton: Neat. That's a cool approach.

BoxContentSDKSampleApp/Makefile
```diff
@@ -0,0 +1,30 @@
...
+
+bootstrap: check_carthage ## Build the reference app dependency framework(s) to cartfile.resolved
+	# Make sure submodules handled by Carthage are up to date (as saved in Cartfile.resolved)
+	$(CARTHAGE) bootstrap --project-directory . --platform iOS
```
jlawton: No real need to have `--project-directory .` in any of these commands.

BoxContentSDKSampleApp/Cartfile
```diff
@@ -0,0 +1,2 @@
+# Box SDKs
+git "git@github.com:box/box-ios-sdk.git" ~> 1.0
```
jlawton: We could also use `github "box/box-ios-sdk" ~> 1.0`. I'm not sure if that affects whether Carthage tries to use binaries from Github releases, but maybe. Carthage would then use https by default, but I think that's fine for public repos.

BoxContentSDKSampleApp/Makefile
jdizoglio: done

BoxContentSDKSampleApp/Makefile
jdizoglio: done

BoxContentSDKSampleApp/Cartfile
jdizoglio: done

jlawton (approved):

BoxContentSDKSampleApp/Cartfile
```diff
@@ -0,0 +1,2 @@
+# Box SDKs
+git "box/box-ios-sdk" ~> 1.0
```
jlawton, line 2: `github`

PR #494, opened 2018-03-08
Modify BOXMetadata model to support new API models.

scottliubox:
Adding version and typeVersion fields to BOXMetadata, also added support for array<NSString> values for custom fields.

### DIFF
--- a/BoxContentSDK/BoxContentSDK/Models/BOXMetadata.h
+++ b/BoxContentSDK/BoxContentSDK/Models/BOXMetadata.h
@@ -29,6 +29,18 @@
 @property (nonatomic, readwrite, strong) NSString *scope;
 
 /**
+ * The current version of the metadata.
+ */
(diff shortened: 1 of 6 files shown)


boxcla:

BoxContentSDK/BoxContentSDK/Models/BOXMetadata.h
```diff
@@ -28,6 +28,18 @@
...
 @property (nonatomic, readwrite, strong) NSString *scope;
 
+/**
+ * The current version of the metadata.
+ */
+@property (nonatomic, readwrite, strong) NSNumber *version;
+
+
+/**
+ * The current type version of the metadata.
```
boxtcarpel, line 38: Can we better explain what a "type version" is? Is it the version of the type or something else?

boxtcarpel (approved):

PR #497, opened 2018-03-18
IOS-15063 Download the best available file format on demand

jdizoglio:
Changing our representations interface.

- Removing bool checks to add representation field to BoxFileRequest
- New mask options introduced
- Includes setRepresentationRequestOptions

### DIFF
--- a/BoxContentSDK/BoxContentSDK/BOXContentSDKConstants.h
+++ b/BoxContentSDK/BoxContentSDK/BOXContentSDKConstants.h
@@ -161,6 +161,7 @@ extern BOXRepresentationType *const BOXRepresentationTypeDASH;
 extern BOXRepresentationType *const BOXRepresentationTypeHLS;
 extern BOXRepresentationType *const BOXRepresentationTypeCrocodoc;
 extern BOXRepresentationType *const BOXRepresentationTypeDICOM;
+extern BOXRepresentationType *const BOXRepresentationTypeExtractedText;
 
 // Representations URL Template
 extern NSString *const BOXRepresentationTemplateKeyAccessPath;
@@ -175,12 +176,19 @@ extern BOXRepresentationStatus *const BOXRepresentationStatusPending;
 extern BOXRepresentationStatus *const BOXRepresentationStatusNone;
 extern BOXRepresentationStatus *const BOXRepresentationStatusError;
 
-// Representation dimensions
-typedef NSString BOXRepresentationDimensions;
-extern BOXRepresentationDimensions *const BOXRepresentationDimensionsThumbnail;
-extern BOXRepresentationDimensions *const BOXRepresentationDimensionsLargeThumbnail;
-extern BOXRepresentationDimensions *const BOXRepresentationDimensions1024x1024;
-extern BOXRepresentationDimensions *const BOXRepresentationDimensions2048x2048;
+// Representation Supported Image Formats
+// Formats greater than 1024 are rendered on demand, all others are rendered once the file is uploaded to Box.
+// JPG 2048, PNG 1024 and PNG 2048 representations not available for video file types
+// If the original file is not a square, images requested above 160x160 will retain the file's original aspect ratio.
+// JPG avaiable in all sizes, PNG available only in 1024 and 2048
+
+typedef NSString BOXRepresentationImageDimensions;
+extern BOXRepresentationImageDimensions *const BOXRepresentationImageDimensionsJPG32;
+extern BOXRepresentationImageDimensions *const BOXRepresentationImageDimensionsJPG94;
+extern BOXRepresentationImageDimensions *const BOXRepresentationImageDimensionsJPG160;
+extern BOXRepresentationImageDimensions *const BOXRepresentationImageDimensionsJPG320;
+extern BOXRepresentationImageDimensions *const BOXRepresentationImageDimensions1024;
+extern BOXRepresentationImageDimensions *const BOXRepresentationImageDimensions2048;
 
 // Folder upload email access level
 typedef NSString BOXFolderUploadEmailAccessLevel;
@@ -299,6 +307,7 @@ extern NSString *const BOXAPIObjectKeySyncState;
 extern NSString *const BOXAPIObjectKeyURL;
 extern NSString *const BOXAPIObjectKeyURLTemplate;
 extern NSString *const BOXAPIObjectKeyDownloadURL;
+extern NSString *const BOXAPIObjectKeyAuthenticatedDownloadURL;
 extern NSString *const BOXAPIObjectKeyVanityURL;
 extern NSString *const BOXAPIObjectKeyIsPasswordEnabled;
 extern NSString *const BOXAPIObjectKeyLogin;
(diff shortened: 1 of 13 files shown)


boxcla:

BoxContentSDK/BoxContentSDK/BOXContentSDKConstants.h
```diff
@@ -175,12 +176,19 @@ extern BOXRepresentationStatus *const BOXRepresentationStatusPending;
...
-extern BOXRepresentationDimensions *const BOXRepresentationDimensions1024x1024;
-extern BOXRepresentationDimensions *const BOXRepresentationDimensions2048x2048;
+// Representation Supported Image Formats
+// Formats greater than 1024 are rendered on demand, all others are rendered once the file is uploaded to Box.
+// JPG 2048, PNG 1024 and PNG 2048 representations not available for video file types
+// If the original file is not a square, images requested above 160x160 will retain the file's original aspect ratio.
+// JPG avaiable in all sizes, PNG available only in 1024 and 2048
+
+typedef NSString BOXRepresentationImageDimensions;
+extern BOXRepresentationImageDimensions *const BOXRepresentationImageDimensionsJPG32;
```
jlawton, line 186: Any reason to add `JPG` in these names?

BoxContentSDK/BoxContentSDK/Models/BOXFile.h
```diff
@@ -91,4 +91,11 @@
 
 @property (nonatomic, readwrite, strong) NSArray *representations;
 
+
+/**
+ *  Url available to download original content.
+ *  Warning: file exists only if canDownload is true
+ */
+@property (nonatomic, readwrite, strong) NSString *downloadUrl;
```
jlawton: Should this be an `NSURL`?

BoxContentSDK/BoxContentSDK/Requests/BOXFileRequest.h
```diff
@@ -5,20 +5,29 @@
 
 #import "BOXRequestWithSharedLinkHeader.h"
 
+typedef NS_OPTIONS(NSUInteger, BOXRepresentationRequestOptions) {
+    BOXRepresentationRequestNoOptions                          = 0,       // Default option if no options set or permission to file not granted
```
jlawton: Maybe we should just leave this out? `0` always behaves oddly in option sets, since you can't ask `x & 0 == 0` or anything like that.

BoxContentSDK/BoxContentSDK/BOXContentSDKConstants.h
jdizoglio: These dimensions are only available for jpeg only. The other 2 largest dimensions are both jpeg or png. See above comments.

BoxContentSDK/BoxContentSDK/Requests/BOXFileRequest.h
```diff
@@ -5,20 +5,29 @@
...
+    BOXRepresentationRequestOriginal                           = 1 << 0,  // Request original content url with file information request
```
jlawton: I think if we use `///` comments on the line above, I think that will get us short descriptions in the Xcode autocomplete box. Might be nice.

jdizoglio: done

BoxContentSDK/BoxContentSDK/Requests/BOXFileRequest.h
jdizoglio: done

BoxContentSDK/BoxContentSDK/Models/BOXFile.h
jdizoglio: done

BoxContentSDK/BoxContentSDK/BOXContentSDKConstants.m
```diff
@@ -278,6 +283,7 @@
 NSString *const BOXAPIObjectKeyURL = @"url";
 NSString *const BOXAPIObjectKeyURLTemplate = @"url_template";
 NSString *const BOXAPIObjectKeyDownloadURL = @"download_url";
+NSString *const BOXAPIObjectKeyAuthenticatedDownloadURL = @"authenticated_download_url";
```
Prithvijutur, line 286: Can't documentation on this key. Its not used yet, but is it something that is a tbd?

BoxContentSDK/BoxContentSDK/Models/BOXFile.h
```diff
@@ -91,4 +91,11 @@
 
 @property (nonatomic, readwrite, strong) NSArray *representations;
 
+
+/**
+ *  Url available to download original content.
+ *  Warning: file exists only if canDownload is true
+ */
+@property (nonatomic, readwrite, strong) NSURL *downloadUrl;
```
Prithvijutur, line 99: Not specific to this PR but throw in a nullable attribute here?

BoxContentSDK/BoxContentSDK/BOXContentSDKConstants.m
jdizoglio: Correct, we discovered from the reps team that original files are served either from 'authenticated_download_url' or 'download_url'. They are both ok to use, for due diligence, I added both. The difference is download_url supported caching for web via url path naming. I was considering changing ours to to authenticated_download_url. A simple path since we catalog our own cache. But will make this as a follow up Jira. I wanted to take time an understand if there is anything else to know about these fields before using the newer 'authenticated' field.

BoxContentSDK/BoxContentSDK/Requests/BOXFileRequest.m
```diff
@@ -241,4 +241,81 @@ - (BOOL)shouldPerformBackgroundOperation
...
+        va_end(argumentList);
+    }
+    
+    self.representationsRequested = [[NSMutableOrderedSet alloc] initWithArray:arguments];
+}
+
+- (NSString *)formatRepresentationRequestHeader
+{
+    // Only process valid API options.
+    if ([self.representationsRequested containsObject:@(BOXRepresentationRequestOriginal)]) {
```
Prithvijutur, line 264: This should already be removed by the time the options are set

Prithvijutur (approved):

BoxContentSDK/BoxContentSDK/Requests/BOXFileRequest.m
jdizoglio: I exposed formatRepresentationRequestHeader for testing. I added this check again to bullet proof reps processing. If a dev calls, I wanted to make sure this was not included in the x-reps-hints header.

BoxContentSDK/BoxContentSDK/Models/BOXFile.h
jdizoglio: Agree, chatted offline.  A good discussion for the team, to go back into ContentSDK and add Nullability attributes.

technicallyerik:
This PR appears to have caused breaking changes in the latest Preview SDK.

PR #505, opened 2018-04-06
Fix Tests

MattKiazyk:
I'm adding another feature in another branch, but one thing that bugged me was that the tests were always failing. This is an attempt to fix them.

1. Added a UI`TestApp` to fix keychain tests.
as per https://forums.developer.apple.com/thread/60617 and others, in order to proper test keychain you need an app, and more importantly a keychain entitlement file. 

2. I changed `BOXContentClient` to use `NSException` instead of an Assert when checking for multiple user sessions. A test was failing (was checking for NSException but code was asserting). I figured this was an ok change, as even in development, if that scenario happens, we can stop the app.

Hopefully this fixes them - they run locally now at least :)

### DIFF
--- a/BoxContentSDK/BoxContentSDK.xcodeproj/project.pbxproj
+++ b/BoxContentSDK/BoxContentSDK.xcodeproj/project.pbxproj
@@ -7,6 +7,12 @@
 	objects = {
 
 /* Begin PBXBuildFile section */
+		00D124202077E1B70026AF5A /* AppDelegate.m in Sources */ = {isa = PBXBuildFile; fileRef = 00D1241F2077E1B70026AF5A /* AppDelegate.m */; };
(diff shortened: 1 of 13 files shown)


boxcla:
Verified that @MattKiazyk has signed the CLA. Thanks for the pull request!

MattKiazyk:
So the one test that is failing in Travis is from a commit that isn't even part of this branch - so not sure what Travis is picking up?

```
BOXFileRequestTests

  test_that_representation_options_are_set_correctly_when_request_is_performed, [31m((expectedHeaderXRepsHint) equal to (actualXRepsHint)) failed: ("[hls]") is not equal to ("[mp4]")[0m

  [36m/Users/travis/build/box/box-ios-sdk/BoxContentSDK/BoxContentSDKTests/BOXFileRequestTests.m:184[0m

  ```

jlawton:
Thank you for contributing! I'll take a look shortly.

jlawton (approved):
This looks great. The tests all pass for me locally and on CI. Thank you, Matt.

PR #507, opened 2018-04-12
Unified metadata

MattKiazyk:
One "feature" of the rest api is the ability to pass up a unified metadata field name that allows the api to return a boxItem's metadata values. This PR adds that ability.

I tried to keep the same format and existing options as is. I've added a new 

```
- (BOXSearchRequest *)searchMetadataRequestWithTemplateKey:(NSString *)templateKey scope:(NSString *)scope filters:(NSArray *)filters inRange:(NSRange)range unifiedMetadataKeys:(NSArray *)unifiedMetadataKeys
```
that adds the ability to pass up an array of unified metadata strings

An example would be 
```
//create searchFilters above as per normal

BOXContentClient.default().searchMetadataRequest(withTemplateKey: "searchTemplateKey", scope: "enterprise, filters: searchFilters, in: NSRange(location: 0, length: maxItems), unifiedMetadataKeys: ["metadata.enterprise.project"]
```

This call returns a `metadata` property inside the search results which the sdk parses and adds a new `metadata` array object onto a BOXItem.

We are successfully using this in an app with no issues. 

Please let me know if anything needs to be changed!

-Matt


boxcla:
Verified that @MattKiazyk has signed the CLA. Thanks for the pull request!

MattKiazyk:
Hi,

There's been no response/comment on this PR for 1.5 months. 

Is this something that could be added? It would be beneficial for a lot of other devs using the sdk.

Thanks

-Matt

scottliubox:
@MattKiazyk Thank you for the PR, we are sorry for not getting back to you earlier. The functionality you added is great and we will be reviewing it very soon. Thanks.

jdizoglio (approved):

PR #508, opened 2018-04-16
add representation info request and model

Prithvijutur:


boxcla:

BoxContentSDK/BoxContentSDK/Models/BOXRepresentationInfo.m
```diff
@@ -0,0 +1,59 @@
...
+
+@implementation BOXRepresentationInfo
+
+- (instancetype)initWithJSON:(NSDictionary *)JSONData
+{
+    if (self = [super init]) {
+
+        _JSONData = JSONData;
+
+        _type = [NSJSONSerialization box_ensureObjectForKey:BOXAPIObjectKeyRepresentation
```
jdizoglio: We should use the current representation model.

BoxContentSDK/BoxContentSDK/Requests/BOXRepresentationInfoRequest.m
```diff
@@ -0,0 +1,63 @@
...
+    self = [super init];
+    if (self != nil) {
+        _representation = representation;
+    }
+    return self;
+}
+
+- (BOXAPIOperation *)createOperation
+{
+    return [self JSONOperationWithURL:self.representation.infoURL
```
jdizoglio, line 32: Maybe evaluate self.representation.infoURL for nil

BoxContentSDK/BoxContentSDK/Requests/BOXRequest.h
```diff
@@ -67,6 +68,8 @@ typedef void (^BOXMetadatasBlock)(NSArray <BOXMetadata *> *metadatas, NSError *e
 
 typedef void (^BOXMetadataTemplatesBlock) (NSArray <BOXMetadataTemplate *>*metadataTemplates, NSError *error);
 
+typedef void (^BOXRepresentationInfoBlock)(BOXRepresentationInfo *info, NSError *error);
```
jdizoglio: BoxRepresentations model should be fine.

BoxContentSDK/BoxContentSDK/Requests/BOXRepresentationInfoRequest.m
Prithvijutur: What should the result for nil infoUrl be? Looking at precedents and it looks like we just do a BOXAssert when createOperation fails

jdizoglio: confirmed. Chatted offline. We do call the request and error. We can discuss with team to add incomplete operation error handling instead of making the request with an invalid url.

PR #509, opened 2018-04-18
IOS-15205: part2 make sure token request's associated access token is …

thuyBox:
…set correctly

### DIFF
--- a/BoxContentSDK/BoxContentSDK/OAuth2/BOXOAuth2Session.m
+++ b/BoxContentSDK/BoxContentSDK/OAuth2/BOXOAuth2Session.m
@@ -106,7 +106,7 @@
                                                                                       BOXAuthTokenRequestClientSecretKey : self.clientSecret,
                                                                                       BOXAuthTokenRequestRedirectURIKey : self.redirectURIString,
                                                                                       }];
-    
+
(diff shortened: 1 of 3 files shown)


boxcla:

BoxContentSDK/BoxContentSDK/OAuth2/BOXOAuth2Session.m
```diff
@@ -236,6 +247,9 @@ - (void)performRefreshTokenGrant:(NSString *)expiredAccessToken
                                                                                          body:POSTParams
                                                                                   queryParams:nil
                                                                                       session:self];
+    //Note: record the expired access token that initiated this refresh operation
+    //to detect if this refresh operation fails because we have acquired a newer access token
+    operation.accessToken = expiredAccessToken;
```
jlawton, line 253: Setting this doesn't cause us to send the `Authorization:` header in the request?

thuyBox: we'll still have the expired access token in the auth header as before, but since it's token request, that header won't matter.

jlawton (approved):

jlawton (approved):

boxtcarpel (commented):
+2

PR #510, opened 2018-04-20
Add background task to fix potential SSO failure

jlawton:


boxcla:

BoxContentSDK/BoxContentSDK/OAuth2/BOXAuthorizationViewController.m
```diff
@@ -163,6 +167,40 @@ - (void)viewDidAppear:(BOOL)animated
...
+    [super viewWillDisappear:animated];
+}
+
+// Fixes a bug starting iOS 11.3, where backgrounding the app to respond to
+// 2FA could cause the login to fail.
+- (void)applicationWillResignActive:(NSNotification *)note
+{
+    UIApplication *app = [UIApplication sharedApplication];
+    [app beginBackgroundTaskWithExpirationHandler:^{
+        [app endBackgroundTask:self.backgroundTaskID];
```
boxtcarpel, line 193: weak self?

BoxContentSDK/BoxContentSDK/OAuth2/BOXAuthorizationViewController.m
```diff
@@ -154,6 +156,8 @@ - (void)viewDidLoad
             me.completionBlock(me, user, error);
         }];
     }
+    
+    self.backgroundTaskID = UIBackgroundTaskInvalid;
```
boxtcarpel: any reason not to initialize this at init time?

boxtcarpel (approved):

BoxContentSDK/BoxContentSDK/OAuth2/BOXAuthorizationViewController.m
jlawton: Shouldn't need it. There should be no way to deallocate self before this block is released.

BoxContentSDK/BoxContentSDK/OAuth2/BOXAuthorizationViewController.m
jlawton: No. In fact, I think that Invalid == 0, and it's automatically initialized, but I'll move it.

PR #516, opened 2018-04-25
IOS-15205: fix random logout by specifying request's access token correctly to correctly detect if there are newer access tokens

thuyBox:


boxcla:

jlawton (approved):

PR #518, opened 2018-04-26
IOS-15606: part 2 to fix sso login issue if app goes into background …

thuyBox:
…(#517)


boxcla:

jlawton (approved):

PR #520, opened 2018-05-01
IOS-15606: part 3 to register background/foreground notification inst…

thuyBox:
…ead of resign/active

### DIFF
--- a/BoxContentSDK/BoxContentSDK/OAuth2/BOXAuthorizationViewController.m
+++ b/BoxContentSDK/BoxContentSDK/OAuth2/BOXAuthorizationViewController.m
@@ -168,35 +168,44 @@ typedef void (^BOXAuthCancelBlock)(BOXAuthorizationViewController *authorization
                                      message:self.connectionErrorMessage];
     }
 
-    [[NSNotificationCenter defaultCenter] addObserver:self selector:@selector(applicationWillResignActive:) name:UIApplicationWillResignActiveNotification object:nil];
-    [[NSNotificationCenter defaultCenter] addObserver:self selector:@selector(applicationDidBecomeActive:) name:UIApplicationDidBecomeActiveNotification object:nil];
+    [[NSNotificationCenter defaultCenter] addObserver:self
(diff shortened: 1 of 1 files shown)


boxcla:

BoxContentSDK/BoxContentSDK/OAuth2/BOXAuthorizationViewController.m
```diff
@@ -168,35 +168,44 @@ - (void)viewDidAppear:(BOOL)animated
                                      message:self.connectionErrorMessage];
     }
 
-    [[NSNotificationCenter defaultCenter] addObserver:self selector:@selector(applicationWillResignActive:) name:UIApplicationWillResignActiveNotification object:nil];
-    [[NSNotificationCenter defaultCenter] addObserver:self selector:@selector(applicationDidBecomeActive:) name:UIApplicationDidBecomeActiveNotification object:nil];
+    [[NSNotificationCenter defaultCenter] addObserver:self
+                                             selector:@selector(didReceiveBackgroundNotification:)
+                                                 name:UIApplicationDidEnterBackgroundNotification
```
boxtcarpel, line 173: This is not too late?

boxtcarpel (approved):
'will resign active' vs. 'did enter background'? How do we know that 'did enter background' isn't/won't be too late for the web view losing it's connection...

BoxContentSDK/BoxContentSDK/OAuth2/BOXAuthorizationViewController.m
thuyBox: we don't get UIApplicationDidBecomeActiveNotification at all. Verified this works on iOS 11.3 device.

jlawton: Apple docs suggest it's okay. https://developer.apple.com/documentation/uikit/uiapplicationdelegate/1622997-applicationdidenterbackground?language=objc

PR #521, opened 2018-05-01
IOS-15606: part 3 to register background/foreground notification inst…

thuyBox:
…ead of resign/active


boxcla:

jlawton (approved):

PR #524, opened 2018-05-02
IOS-15408: allow query for associateIds of specific background sessio…

thuyBox:
…n id and user id


boxcla:

jlawton (approved):

PR #526, opened 2018-05-09
enfore data integrity check on all uploads

Prithvijutur:
- Verified sample app can still upload.


boxcla:

PR #529, opened 2018-05-11
add content-md5 header for data uploads

Prithvijutur:
Add content-md5 for data uploads only

TBD 
   - Convert upload tasks to background upload (deprecate foreground tasks)
   - Calculate sha1 on copy at background temporary filepath
   - Add to header on completion handler

### DIFF
--- a/BoxContentSDK/BoxContentSDK/Clients/BOXContentClient+File.h
+++ b/BoxContentSDK/BoxContentSDK/Clients/BOXContentClient+File.h
@@ -131,6 +131,7 @@
  */
 - (BOXFileUploadRequest *)fileUploadRequestToFolderWithID:(NSString *)folderID
                                         fromLocalFilePath:(NSString *)localFilePath;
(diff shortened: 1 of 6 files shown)


boxcla:

BoxContentSDK/BoxContentSDK/Clients/BOXContentClient+File.h
```diff
@@ -170,6 +171,7 @@
  */
 - (BOXFileUploadNewVersionRequest *)fileUploadNewVersionRequestWithID:(NSString *)fileID
                                                     fromLocalFilePath:(NSString *)localFilePath;
+__attribute__((deprecated("Foreground fileuploads will be removed. Please use fileUploadNewVersionRequestInBackgroundWithFileID")));
```
Prithvijutur, line 174: Anyone else, please comment on if there is a reason against disabling foregroup uploads

PR #530, opened 2018-05-12
Add content-md5 header for multipart uploads

Prithvijutur:


boxcla:

BoxContentSDK/BoxContentSDK/BOXURLRequestSerialization.h
```diff
@@ -12,10 +12,11 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
+@protocol BOXStreamingDataProcessor;
 @interface BOXMultipartBodyStream : NSInputStream <NSStreamDelegate>
 
 @property (readonly, nonatomic, assign) unsigned long long contentLength;
-
+@property (nonatomic, readwrite, weak) id<BOXStreamingDataProcessor> contentProcessor;
```
jlawton, line 19: It's a little odd to me that this is weak.

BoxContentSDK/BoxContentSDK/External/HashHelper/BOXStreamingHashHelper.m
```diff
@@ -0,0 +1,48 @@
...
+
+#import "BOXHashHelper.h"
+
+#include <CommonCrypto/CommonDigest.h>
+#include <CoreFoundation/CoreFoundation.h>
+#import "BOXStreamingHashHelper.h"
+
+@interface BOXStreamingHashHelper()
+@property (nonatomic, readwrite, strong) NSString *calcSha1;
+@property (nonatomic, assign) CC_SHA1_CTX *context;
```
jlawton: This could possibly just be allocated inline, to avoid the `malloc` and `free`?

BoxContentSDK/BoxContentSDK/External/HashHelper/BOXStreamingHashHelper.m
```diff
@@ -0,0 +1,48 @@
...
+    CC_SHA1_Init(_context);
+}
+
+- (void) processData:(NSData* ) buffer{
+    CC_SHA1_Update(_context, [buffer bytes], (CC_LONG)buffer.length);
+}
+
+- (NSString *) close {
+    unsigned char digest[CC_SHA1_DIGEST_LENGTH];
+    CC_SHA1_Final(&digest[0], _context);
```
jlawton: I think this could just be `digest` rather than `&digest[0]`.

BoxContentSDK/BoxContentSDK/External/HashHelper/BOXStreamingHashHelper.m
```diff
@@ -0,0 +1,48 @@
...
+
+    char hash[2 * sizeof(digest) + 1];
```
jlawton: I think it might be necessary to initialize this with `= ""` or `= {0}` in order to guarantee that it's zero-filled (the only exception, I believe, is for `static` variables). It's likely to be zeros by accident, so this would only show as a bug rarely, but could cause a EXC_BAD_ACCESS on the NSString initialization below.

BoxContentSDK/BoxContentSDK/BOXURLRequestSerialization.m
```diff
@@ -1166,6 +1179,10 @@ - (NSInteger)read:(uint8_t *)buffer
         NSInteger numberOfBytesRead = 0;
 
         numberOfBytesRead = [self.inputStream read:&buffer[totalNumberOfBytesRead] maxLength:(length - (NSUInteger)totalNumberOfBytesRead)];
+        if(self.contentProcessor) {
+            NSData *bodyData = [NSData dataWithBytes:(const void *)(&buffer[0] + totalNumberOfBytesRead) length:numberOfBytesRead];
```
jlawton: Could use `buffer` rather than `&buffer[0]`

jlawton: If we know that `processData:` completes synchronously, we should be able to use the non-copying NSData constructor here, and avoid a little work.

BoxContentSDK/BoxContentSDK/BOXURLRequestSerialization.m
```diff
@@ -431,23 +432,27 @@ - (BOXStreamingMultipartFormData *)formDataWithRequest:(NSMutableURLRequest *)re
...
+        BOXMultipartBodyStream *bodyStream = (BOXMultipartBodyStream*) inputStream;
+        bodyStream.contentProcessor = streamingHashHelper;
+        [streamingHashHelper open];
 
         while ([inputStream hasBytesAvailable] && [outputStream hasSpaceAvailable]) {
-            uint8_t buffer[1024];
+            uint8_t buffer[(1024 * 8)];
 
-            NSInteger bytesRead = [inputStream read:buffer maxLength:1024];
+            NSInteger bytesRead = [inputStream read:buffer maxLength:(1024 * 8)];
```
jlawton: Maybe we should use `sizeof(buffer)` here to avoid the repetition.

BoxContentSDK/BoxContentSDK/External/HashHelper/BOXStreamingHashHelper.m
```diff
@@ -0,0 +1,48 @@
...
+@end
+
+@implementation BOXStreamingHashHelper
+- (instancetype) init {
+    _context = (CC_SHA1_CTX*) malloc(sizeof(CC_SHA1_CTX));
+    return self;
+}
+
+- (void) open {
+    CC_SHA1_Init(_context);
```
jlawton: Consider calling `open` in `init` or keeping track of the open state, to avoid mis-use?

BoxContentSDK/BoxContentSDK/External/HashHelper/BOXStreamingHashHelper.h
```diff
@@ -0,0 +1,18 @@
...
+//  BOXStreamingHashHelper.h
+//  BoxContentSDK
+//
+//  Created by Prithvi Jutur on 5/21/18.
+//  Copyright © 2018 Box. All rights reserved.
+//
+
+#import <Foundation/Foundation.h>
+
+@protocol BOXStreamingDataProcessor <NSObject>
```
jlawton: Probably doesn't need to inherit from `NSObject`, with no optional methods.

BoxContentSDK/BoxContentSDK/External/HashHelper/BOXStreamingHashHelper.h
```diff
@@ -0,0 +1,18 @@
+//
+//  BOXStreamingHashHelper.h
+//  BoxContentSDK
+//
+//  Created by Prithvi Jutur on 5/21/18.
+//  Copyright © 2018 Box. All rights reserved.
+//
+
+#import <Foundation/Foundation.h>
```
jlawton: `@import`

jlawton: `NS_ASSUME_NONNULL`

BoxContentSDK/BoxContentSDK/External/HashHelper/BOXStreamingHashHelper.h
jlawton: Maybe a short doc string. Otherwise I have to look elsewhere to see why `close` returns `NSString *`

BoxContentSDK/BoxContentSDK/External/HashHelper/BOXStreamingHashHelper.m
```diff
@@ -0,0 +1,48 @@
...
+//
+
+#import "BOXHashHelper.h"
+
+#include <CommonCrypto/CommonDigest.h>
+#include <CoreFoundation/CoreFoundation.h>
+#import "BOXStreamingHashHelper.h"
+
+@interface BOXStreamingHashHelper()
+@property (nonatomic, readwrite, strong) NSString *calcSha1;
```
jlawton: I think this is never used.

BoxContentSDK/BoxContentSDK/BOXURLRequestSerialization.h
```diff
@@ -271,7 +272,7 @@ forHTTPHeaderField:(NSString *)field;
  */
 - (NSMutableURLRequest *)requestWithMultipartFormRequest:(NSURLRequest *)request
                              writingStreamContentsToFile:(NSURL *)fileURL
-                                       completionHandler:(nullable void (^)(NSError * _Nullable error))handler;
+                                       completionHandler:(nullable void (^)(NSString * digest, NSError * _Nullable error))handler;
```
jlawton: Minor: Could `digest` be `nil` or invalid on error?

BoxContentSDK/BoxContentSDK/BOXURLRequestSerialization.m
```diff
@@ -643,6 +649,8 @@ @interface BOXHTTPBodyPart : NSObject
 
 - (NSInteger)read:(uint8_t *)buffer
         maxLength:(NSUInteger)length;
+
+@property (nonatomic, readwrite, weak) id<BOXStreamingDataProcessor> contentProcessor;
```
jlawton, line 653: Weak seems odd to me.

jlawton (commented):
Mostly nitpicking.

BoxContentSDK/BoxContentSDK/BOXURLRequestSerialization.h
Prithvijutur: No point updating the content processor if the owner no longer owns the reference?

BoxContentSDK/BoxContentSDK/BOXURLRequestSerialization.m
```diff
@@ -1166,6 +1179,10 @@ - (NSInteger)read:(uint8_t *)buffer
         NSInteger numberOfBytesRead = 0;
 
         numberOfBytesRead = [self.inputStream read:&buffer[totalNumberOfBytesRead] maxLength:(length - (NSUInteger)totalNumberOfBytesRead)];
+        if(self.contentProcessor) {
+            NSData *bodyData = [NSData dataWithBytes:(const void *)(buffer + totalNumberOfBytesRead) length:numberOfBytesRead];
```
jlawton: Consider non-copying constructor.

BoxContentSDK/BoxContentSDK/BOXURLRequestSerialization.m
```diff
@@ -1166,6 +1179,10 @@ - (NSInteger)read:(uint8_t *)buffer
         NSInteger numberOfBytesRead = 0;
 
         numberOfBytesRead = [self.inputStream read:&buffer[totalNumberOfBytesRead] maxLength:(length - (NSUInteger)totalNumberOfBytesRead)];
+        if(self.contentProcessor) {
```
jlawton: Put contentProcessor in a strong local reference, since it's accessed twice?

BoxContentSDK/BoxContentSDK/BOXURLRequestSerialization.m
Prithvijutur: According to the docs(https://developer.apple.com/documentation/foundation/nsdata/1547229-datawithbytesnocopy?language=objc), Non-copying constructor only works on blocks allocated from heap using malloc since it takes ownership of the buffer.

BoxContentSDK/BoxContentSDK/BOXURLRequestSerialization.m
Prithvijutur: I can pass raw pointers without converting to NSData to save on the copy but I thought passing around the chunk as NSData would be a better interface for the contentProcessor

jlawton: I was thinking about `+[NSData dataWithBytesNoCopy:length:]`

jlawton (approved):

PR #533, opened 2018-05-25
Update travisCI to use xcode 9.4

Prithvijutur:
- Also fixes couple of warnings


boxcla:

boxtcarpel:
Should we update this to use Xcode 9.4?

boxtcarpel (approved):

PR #534, opened 2018-05-30
Prefer hls video playback representations for iOS 11

jdizoglio:


boxcla:

jlawton (approved):

PR #541, opened 2018-06-13
Removing the set representation option method that used va_list param…

jdizoglio:
…eters to support the swift bridge, changing the method to an NSSet parameter. For consitentcy we should follow up migrating the objc set represetation option method as well.


boxcla:

Prithvijutur (approved):

BoxContentSDK/BoxContentSDK/Requests/BOXFileRequest.h
```diff
@@ -86,6 +86,8 @@ typedef NS_OPTIONS(NSUInteger, BOXRepresentationRequestOptions) {
 
 - (void)setRepresentationRequestOptions:(BOXRepresentationRequestOptions)representationOptions args:(va_list)args;
```
scottliubox: Can we remove the header of the replaced method?

jdizoglio: yes, it was a miss. good catch

scottliubox (approved):

PR #542, opened 2018-06-13
Removing the set representation option method that used va_list param…

jdizoglio:
…eters to support the swift bridge, changing the method to an NSSet parameter. For consitentcy we should follow up migrating the objc set represetation option method as well. (#541)


boxcla:

boxtcarpel (approved):

Prithvijutur (approved):

PR #544, opened 2018-06-18
change sha1 calculation to happen async to completing file download

Prithvijutur:
After running time profiler on sha1 calculation, it appears that there is a significant slowdown in marking large downloads as complete. While tolerable, we want to still delegate handling these rare cases to SDK consumers to deal with it as needed. (Either invalidating the cache or recording metrics). The tradeoff here is between adding increased download times for everyone Vs catching the rare data integrity error

### DIFF
--- a/BoxContentSDK/BoxContentSDK/Requests/BOXFileRepresentationDownloadRequest.m
+++ b/BoxContentSDK/BoxContentSDK/Requests/BOXFileRepresentationDownloadRequest.m
@@ -116,16 +116,22 @@
         }
         
         fileOperation.successBlock = ^(NSString *modelID, long long expectedTotalBytes) {
-            NSError *dataIntegrityErrorIfAny = nil;
-            if([self.sha1Hash length] && ![self.sha1Hash isEqualToString: [BOXHashHelper sha1HashOfFileAtPath:self.destinationPath]]) {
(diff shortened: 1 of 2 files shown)


boxcla:

BoxContentSDK/BoxContentSDK/Requests/BOXFileRepresentationDownloadRequest.m
```diff
@@ -116,16 +116,19 @@ - (void)performRequestWithProgress:(BOXProgressBlock)progressBlock completion:(B
...
-                if(dataIntegrityErrorIfAny) {
-                    [[NSNotificationCenter defaultCenter] postNotificationName:BOXFileDownloadCorruptedNotification object:self];
-                }
-            }
             [BOXDispatchHelper callCompletionBlock:^{
-                completionBlock(dataIntegrityErrorIfAny);
+                completionBlock(nil);
             } onMainThread:isMainThread];
+
+            NSString *calculatedSha1 = [BOXHashHelper sha1HashOfFileAtPath:self.destinationPath];
```
RLEE0731: Are we only doing the sha1 check on the original item? Or does it check all representation types?

Prithvijutur: We do the check if the string exists. However, currently, the BOX API only returns the sha1 for the original content.

PR #545, opened 2018-06-20
Undo previous commit deprecating foreground uploads

Prithvijutur:
Due to performance implications on background uploads, we want to keep supporting foreground uploads for use-cases that dont need to be backgrounded


boxcla:

DigitalSolomon (approved):
Discussed offline, LGTM

PR #546, opened 2018-06-21
Update reps interface in content SDK remove ns_options to enum and re…

jdizoglio:
…move variadic list methods to use array list


boxcla:

jlawton (approved):

PR #547, opened 2018-06-22
Update original file url reference to use authenticated_download_url

jdizoglio:


boxcla:

thuyBox (approved):

PR #549, opened 2018-06-25
Updating sample app compatibility

RLEE0731:
With the recent updates to the SDK, we're now enforcing SHA1 verification in file upload request header, instead of having it be optional.


boxcla:
Verified that @RLEE0731 has signed the CLA. Thanks for the pull request!

thuyBox (approved):

PR #556, opened 2018-08-07
Improved app user support

cburnette:
Box iOS team, please review.  Thanks!


boxcla:
Verified that @cburnette has signed the CLA. Thanks for the pull request!

BoxContentSDK/BoxContentSDK/Clients/BoxContentClientServerAuthWrapper.h
```diff
@@ -0,0 +1,18 @@
+#import "BOXContentClient.h"
+#import "BOXAPIAccessTokenDelegate.h"
+
+typedef void (^ServerAuthFetchTokenBlock)(NSString *userId, NSDictionary *userInfo, void (^)(NSString *, NSDate *, NSError *));
```
jlawton: Suggestion: This is a little hard to read, especially without names on the callback parameters. I might consider this or similar as an option:

```
typedef void (^ServerAuthFetchTokenBlock)(NSString *userId, NSDictionary *userInfo, TokenFetchCompletion completion);

// Either provide a token and its expiration, or an error.
typedef void (^TokenFetchCompletion)(NSString * __nullable accessToken, NSDate * __nullable tokenExpiration, NSError * __nullable error);
```

BoxContentSDK/BoxContentSDK/Clients/BoxContentClientServerAuthWrapper.h
```diff
@@ -0,0 +1,18 @@
...
+
+@interface BoxContentClientServerAuthWrapper : NSObject <BOXAPIAccessTokenDelegate>
+
+@property (nonatomic, nonnull, readonly, strong) BOXContentClient *boxClient;
+@property (nonatomic, nullable, readonly, strong) NSString *userId;
+@property (nonatomic, nullable, readwrite, strong) NSDictionary *userInfo;
```
jlawton: NSString and NSDictionary should be copied, lest they be modified under us.

BoxContentSDK/BoxContentSDK/Clients/BoxContentClientServerAuthWrapper.h
```diff
@@ -0,0 +1,18 @@
+#import "BOXContentClient.h"
+#import "BOXAPIAccessTokenDelegate.h"
+
```
jlawton: Wrap body of file in `NS_ASSUME_NONNULL_BEGIN` and `NS_ASSUME_NONNULL_END` and we only need to call out `nullable` elsewhere.

BoxContentSDK/BoxContentSDK/Clients/BoxContentClientServerAuthWrapper.m
```diff
@@ -0,0 +1,36 @@
+#import "BoxContentClientServerAuthWrapper.h"
+#import "BoxAbstractSession.h"
+
+@interface BoxContentClientServerAuthWrapper ()
+
+@property (nonnull, nonatomic, readonly, strong) ServerAuthFetchTokenBlock fetchTokenBlock;
```
jlawton: Blocks should be copied or we could be given a stack allocated block and crash.

BoxContentSDK/BoxContentSDK/Clients/BoxContentClientServerAuthWrapper.m
```diff
@@ -0,0 +1,36 @@
+#import "BoxContentClientServerAuthWrapper.h"
+#import "BoxAbstractSession.h"
+
```
jlawton: Also assume nonnull in here.

BoxContentSDK/BoxContentSDK/Clients/BoxContentClientServerAuthWrapper.h
```diff
@@ -0,0 +1,18 @@
...
+typedef void (^ServerAuthFetchTokenBlock)(NSString *userId, NSDictionary *userInfo, void (^)(NSString *, NSDate *, NSError *));
+
+@interface BoxContentClientServerAuthWrapper : NSObject <BOXAPIAccessTokenDelegate>
+
+@property (nonatomic, nonnull, readonly, strong) BOXContentClient *boxClient;
+@property (nonatomic, nullable, readonly, strong) NSString *userId;
+@property (nonatomic, nullable, readwrite, strong) NSDictionary *userInfo;
+
+
+- (instancetype)initWithToken:(nullable NSString*)token
```
jlawton: Is it okay to take a token without the ability to take an expiration date?

doc/BoxPlatform.md
```diff
@@ -0,0 +1,74 @@
...
+
+In order to support this application architecture, there is a wrapper class included in the SDK to handle the mechanics of retrieving tokens for a user from the remote server. It includes the ability to specify an initial token that you may have already retrieved, and allows you to specify a code block that will be called by the SDK when it detects an expired token.  This code block is where you call your secure server with whatever information you need in order to retrieve a new token for the user.
+
+<b>PLEASE NOTE:</b> It is up to you as the app developer to ensure that the process of retrieving tokens is secured.  Typically this is done by passing a JSON Web Token in the Authorization header of the request to your remote server that is signed by your authentication system.  That way you can verify that the incoming request is indeed originating from an authorized app and device.
```
jlawton, line 10: I know that this is more of a side note, but I'm curious how you securely get a JWT onto the user's device, which is what I think I'm reading here.

BoxContentSDK/BoxContentSDK/Clients/BoxContentClientServerAuthWrapper.m
```diff
@@ -0,0 +1,36 @@
...
+@property (nonnull, nonatomic, readonly, strong) ServerAuthFetchTokenBlock fetchTokenBlock;
+
+@end
+
+@implementation BoxContentClientServerAuthWrapper
+
+- (instancetype)initWithToken:(nullable NSString *)token
+                       userId:(nullable NSString *)userId
+                     userInfo:(nullable NSDictionary *)userInfo
```
jlawton: Why have a `userInfo` when the block could just close around what it needs? I suppose because it's a readwrite property? If that's the rationale, it should definitely be atomic, too.

doc/BoxPlatform.md
```diff
@@ -0,0 +1,74 @@
...
+
+Below is an example of configuring the <b>BoxContentClientServerAuthWrapper</b> instance to work with a specific App User Id, including example code of how to retrieve a new token when the SDK detects a token expiration. Note that you are not require to specify an initial token as the SDK will simply invoke your fetchTokenBlock. However, in some cases you may already have a token (for example, a downscoped token) that you wish to initialize the wrapper with.  The userInfo parameter is a dictionary for specifying arbitrary information you may need in your fetchTokenBlock code.
+
+```objectivec
```
jlawton, line 14: We get a lot of complaints that the SDK is Objective-C, and users want Swift. There are good reasons not to write the SDK in Swift yet, but maybe we do want our sample code to be Swift? I can see arguments that we wouldn't want that, but thought it should be considered.

BoxContentSDK/BoxContentSDK/Clients/BoxContentClientServerAuthWrapper.m
```diff
@@ -0,0 +1,36 @@
...
+        
+        [_boxClient session].accessToken = token;
+    }
+    
+    return self;
+}
+
+- (void)fetchAccessTokenWithCompletion:(void (^)(NSString *, NSDate *, NSError *))completion
+{
+    _fetchTokenBlock(_userId, _userInfo, completion);
```
jlawton: I don't see a lot of code directly accessing ivars these days. Might want to use `self.fetchTokenBlock`

BoxContentSDK/BoxContentSDK/OAuth2/BOXAppUserSession.m
```diff
@@ -24,25 +24,9 @@ - (void)performAuthorizationCodeGrantWithReceivedURL:(NSURL *)URL withCompletion
         weakSelf.accessToken = accessToken;
         weakSelf.accessTokenExpiration = accessTokenExpiration;
         
-        BOXUserRequest *userRequest = [[BOXUserRequest alloc] init];
-        [weakSelf prepareRequest:userRequest];
-        
-        [userRequest performRequestWithCompletion:^(BOXUser *user, NSError *error) {
```
jlawton, line 30: Removing this is not a backwards-compatible change for the non-app-user case, I believe. How do we at least get our automatic keychain storage to work for the normal Box user case?

jlawton (commented):
Sorry that these comments are at a couple of levels of abstraction. At a high level, I think this direction is getting there, but I worry that this change doesn't maintain backwards compatibility. For instance, I know that it breaks assumptions in the official Box iOS app.

BoxContentSDK/BoxContentSDK/OAuth2/BOXAppUserSession.m
cburnette: @jlawton  so actually it is (I believe) as that session is only used for app user sessions.  The other sessions are un-modified.  Anyway, I just checked in a big change where I realized I can eliminate the wrapper and just move this functionality into BOXContentClient with a new initializer class method.  Have a look and let me know what you think.  thanks!

cburnette: @jlawton  I don't believe I've changed anything with the existing user models

jlawton: Ah. Great. I'll take another look.

BoxContentSDK/BoxContentSDK/Clients/BOXContentClient.h
```diff
@@ -74,6 +77,18 @@ extern NSString *const BOXContentClientBackgroundTempFolder;
  */
 @property (nonatomic, readwrite, weak) id<BOXAPIAccessTokenDelegate> accessTokenDelegate;
 
+/**
+ * When using a server-based auth client (App Users or downscoped tokens) this property store the box user id of the associated user.
+ * Passed into your fetchTokenBlock delegate method when a new token is required.
+ */
+@property (nonatomic, nullable, readwrite, strong) NSString *userId;
```
jlawton: I see a lot of potential for using this accidentally, or setting it at the wrong time. It should probably only be in the implementation file, and if the developer wants to switch users, they should be forced to create a fresh client. (Should also be a `copy` property.)

BoxContentSDK/BoxContentSDK/Clients/BOXContentClient.h
```diff
@@ -74,6 +77,18 @@ extern NSString *const BOXContentClientBackgroundTempFolder;
...
+
+/**
+ * This property is for storing any relevant information needed when your fetchTokenBlock delegate method is invoked
+ * to retrieve a new token.  Passed into your fetchTokenBlock delegate method when a new token is required.
+ */
+@property (nonatomic, nullable, readwrite, strong) NSDictionary *userInfo;
```
jlawton: Should also be an `atomic` and `copy` property. I wonder if the name should be more specific.

BoxContentSDK/BoxContentSDK/Clients/BOXContentClient.m
```diff
@@ -46,6 +46,8 @@ @interface BOXContentClient ()
 
 @property (nonatomic, readwrite, strong) BOXSharedLinkHeadersHelper *sharedLinksHeaderHelper;
 
+@property (nonnull, nonatomic, readwrite, strong) ServerAuthFetchTokenBlock fetchTokenBlock;
```
jlawton: This should be a `copy` property, or we're inviting crashes.

jlawton (commented):
I like this version better than the previous one.

BoxContentSDK/BoxContentSDK/Clients/BOXContentClient.h
cburnette: good point, i'll make that change.

BoxContentSDK/BoxContentSDK/Clients/BOXContentClient.h
cburnette: so on the atomic vs. nonatomic thing, i looked through the existing code and it seems like most things should be atomic but are nonatomic so I went with that.  I can change to atomic.  as for userInfo I went with that name as that's the idiomatic name for a dictionary of random information to be delivered in a callback, correct?

BoxContentSDK/BoxContentSDK/Clients/BOXContentClient.m
cburnette: ok cool, didn't realize you could copy a block.  Thanks!

BoxContentSDK/BoxContentSDK/Clients/BOXContentClient.h
cburnette: and yes, my intention is that if you want a client for a different app user for some reason you would create a new one.

BoxContentSDK/BoxContentSDK/Clients/BOXContentClient.h
jlawton: It's no doubt true that we're wrong elsewhere. This is presumably expected to be written by the user, and also used in a callback on an internal queue, so it seems we should get the semantics right for this. Most of our readwrite properties are really only expected to be written in some sort of setup phase. If that were the case here, the user could just capture the value in their block.

I'm going back and forth on the name. `userInfo` is idiomatic, as you say. The trouble I'm having is that calling this particular block I wouldn't really consider the main responsibility of the class, so the generic name just feels odd. It's probably just me, though.

cburnette: As far as the name, I'm open to suggestions.  In terms of capturing this info in the block, keep in mind that typically the way this is set up there is one block that contains the logic to call the backend given a few parameters, like a user id and any extra info like a scope list.  So you could instantiate multiple BOXContentClients that all have the same fetchTokenBlock but have different user ids and extra user info.

cburnette: what about fetchTokenBlockInfo?

cburnette:
Please don't anyone merge this yet as I'm working through a nuance when using this with the Preview SDK.  Thanks!

cburnette:
reopened with adjusted design

BoxContentSDK/BoxContentSDK/Clients/BOXContentClient.h
```diff
@@ -74,6 +79,12 @@ extern NSString *const BOXContentClientBackgroundTempFolder;
  */
 @property (nonatomic, readwrite, weak) id<BOXAPIAccessTokenDelegate> accessTokenDelegate;
 
+/**
+ * This property is for storing any relevant information needed when your fetchTokenBlock delegate method is invoked
+ * to retrieve a new token.  Passed into your fetchTokenBlock delegate method when a new token is required.
+ */
+@property (nonatomic, nullable, readwrite, strong) NSDictionary *fetchTokenBlockInfo;
```
jlawton: This should be `copy`, rather than `strong`

BoxContentSDK/BoxContentSDK/Clients/BOXContentClient.m
```diff
@@ -126,22 +133,32 @@ + (BOXContentClient *)clientForUser:(BOXUserMini *)user
...
-            [[[self class] SDKClients] setObject:client forKey:user.modelID];
+            [[[self class] SDKClients] setObject:client forKey:user.uniqueId];
         }
         
         return client;
     }
 }
 
-+ (BOXContentClient *)clientForNewSession
++ (BOXContentClient *)clientForServerAuthUser:(nonnull ServerAuthUser*)serverAuthUser
```
jlawton: Nit: Pointer spacing should be `Type *`

BoxContentSDK/BoxContentSDK/Clients/BOXContentClient.m
```diff
@@ -409,6 +434,7 @@ - (NSString*)tempCacheDir
 }
 
 - (void)setAccessTokenDelegate:(id<BOXAPIAccessTokenDelegate>)accessTokenDelegate
+                serverAuthUser:(ServerAuthUser*)serverAuthUser
```
jlawton: Nit: Pointer spacing.

BoxContentSDK/BoxContentSDK/Clients/BOXContentClient.h
```diff
@@ -118,6 +129,21 @@ extern NSString *const BOXContentClientBackgroundTempFolder;
...
+ * Get a BOXContentClient that supports server-based auth (App Users or downscoped tokens).
+ *
+ * @param serverAuthUser A required instance of ServerAuthUser that will be passed to the fetchTokenBlock when a new token is required from the remote server
+ * @param token An optional token that you have already retrieved from your remote server; if nil your fetchTokenBlock will be invoked to retrieve a new token
+ * @param fetchTokenBlockInfo An optional dictionary of relevant information you may need when your fetchTokenBlock is invoked
+ * @param fetchTokenBlock A required code block that will be invoked whenever an expired token is detected.  This is where you will call your secure remote server to generate a new token
+ *
+ * @return BOXContentClient that is configured for server-based auth
+ */
++ (BOXContentClient *)clientForServerAuthUser:(nonnull ServerAuthUser*)serverAuthUser
```
jlawton: Nit: Spacing.

BoxContentSDK/BoxContentSDK/Models/BOXUser.h
```diff
@@ -7,12 +7,43 @@
 #import "BOXModel.h"
 #import "BOXEnterprise.h"
 
+@protocol UniqueSDKUser
+
+@property (nonnull,  nonatomic, readonly, strong) NSString *uniqueId;
+@property (nullable, nonatomic, readonly, strong) NSString *name;
+@property (nullable, nonatomic, readonly, strong) NSString *login;
```
jlawton: Maybe add doc comments to the protocol, as well as, or rather than, the ServerAuthUser.

jlawton: These properties (and in ServerAuthUser) should be `copy` rather than `strong`.

BoxContentSDK/BoxContentSDK/Models/BOXUser.m
```diff
@@ -44,6 +68,10 @@ - (instancetype)initWithUserID:(NSString *)userID name:(NSString *)name login:(N
     return self;
 }
 
+- (NSString *)uniqueId{
```
jlawton: Nit: Space before `{`

BoxContentSDK/BoxContentSDK/OAuth2/BOXAbstractSession.m
```diff
@@ -258,8 +258,8 @@ + (NSArray *)usersInKeychain
     }
     
     NSArray *sortedUsers = [users sortedArrayUsingComparator:^NSComparisonResult(id a, id b) {
-        BOXUserMini *userA = (BOXUserMini *) a;
-        BOXUserMini *userB = (BOXUserMini *) b;
+        id<UniqueSDKUser> userA = (id<UniqueSDKUser>) a;
```
jlawton: You can declare `a` and `b` as `id<UniqueSDKUser>` in the block parameters, and you don't need the casts.

jlawton: I wonder if this sort is even necessary, but we don't have to address that here.

cburnette:
@jlawton Hey James, just checked in changes to address your comments. Thanks!

doc/BoxPlatform.md
cburnette: As an example, in the app I put together for Boxworks that uses AWS Mobile Hub for authentication, there is a secure connection established between the app and AWS using a secret (and their SDK), and when a user provides username/password and its validated, AWS generates a signed JWT and returns it to the app.  That JWT is send in the Authentication header for all the other calls to other AWS services, like Lambda, where the Box token generation happens.

doc/BoxPlatform.md
cburnette: good point, I have Swift example code I can include

jdizoglio (approved):

PR #558, opened 2018-08-27
Refactor representations 

Prithvijutur:
Move reps into helper class that requests can include via composition instead of being tied to base class.

### DIFF
--- a/BoxContentSDK/BoxContentSDK/BOXContentSDKConstants.h
+++ b/BoxContentSDK/BoxContentSDK/BOXContentSDKConstants.h
@@ -515,3 +515,33 @@ typedef NS_ENUM(NSUInteger, BOXAvatarType) {
     BOXAvatarTypeLarge
 };
 
+/**
+ Representations are digital assets created for files stored in Box. To request representations information for
+ a file, set the options for each type.
+ 
+ - BOXRepresentationRequestOriginal:                        Request original content url with file information request
+ Reference the reponse field 'authenticated_download_url' to download original content when file
+ has download permissions are owner, co-owner, editor, viewer uploader, viewer
+ - BOXRepresentationRequestAllRepresentations:              If permissions allow retrieve representations for thumbnails, preview representations for the given file
+ - BOXRepresentationRequestHighDefinitionVideo:             Request video representions if available for the given file
+ - BOXRepresentationRequestThumbnailRepresentation:         Request thumbnail representions if available for the given file
+ - BOXRepresentationRequestLargeThumbnailRepresentation:    Request large thumbnail representions are available for the given file
+ - BOXRepresentationRequestPDFRepresentation:               Request PDF representions if available for the given file
+ - BOXRepresentationRequestJPGRepresentation:               Request JPG representions if available for the given file
+ - BOXRepresentationRequestMP3Representation:               Request PNG representions if available for the given file
+ - BOXRepresentationRequestMP4Representation:               Request MP4 representions if available for the given file
+ - BOXRepresentationRequesteExtractedTextRepresentation:    Request extracted text if unformatted text is contained in a document (non-image)
+ */
+typedef NS_ENUM(NSUInteger, BOXRepresentationRequestOptions) {
+    BOXRepresentationRequestOriginal,
+    BOXRepresentationRequestAllRepresentations,
+    BOXRepresentationRequestHighDefinitionVideo,
+    BOXRepresentationRequestThumbnailRepresentation,
+    BOXRepresentationRequestLargeThumbnailRepresentation,
+    BOXRepresentationRequestPDFRepresentation,
+    BOXRepresentationRequestJPGRepresentation,
+    BOXRepresentationRequestMP3Representation,
+    BOXRepresentationRequestMP4Representation,
+    BOXRepresentationRequesteExtractedTextRepresentation
+} NS_ENUM_AVAILABLE_IOS(10_0);
+
--- a/BoxContentSDK/BoxContentSDK/Helper/BOXRepresentationsHelper.h
+++ b/BoxContentSDK/BoxContentSDK/Helper/BOXRepresentationsHelper.h
@@ -1,0 +1,22 @@
+//
+//  BOXRepresentationsHelper.h
+//  BoxContentSDK
+//
+//  Created by Prithvi Jutur on 8/24/18.
+//  Copyright © 2018 Box. All rights reserved.
+//
+
+#import <Foundation/Foundation.h>
+
+@interface BOXRepresentationsHelper : NSObject
+/**
+ Setting a representation or list option will include availability of file with request representation information
+ */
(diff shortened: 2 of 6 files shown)


boxcla:

BoxContentSDK/BoxContentSDK/Helper/BOXRepresentationsHelper.h
```diff
@@ -0,0 +1,22 @@
...
+
+@interface BOXRepresentationsHelper : NSObject
+/**
+ Setting a representation or list option will include availability of file with request representation information
+ */
+@property (nonatomic, readwrite, strong) NSMutableOrderedSet *representationsRequested;
+
+/**
+ Setting a representation or Array of options will include availability for that file when request representation information
+ */
```
ric0liva, line 19: This comment and the previous one are the same. Is this intended?

BoxContentSDK/BoxContentSDK/Helper/BOXRepresentationsHelper.m
```diff
@@ -0,0 +1,99 @@
...
+#import "BOXRepresentationsHelper.h"
+#import "BOXContentSDKConstants.h"
+
+@interface BOXRepresentationsHelper()
+
+///**
+// Setting a representation or list option will include availability of file with request representation information
+// */
+//@property (nonatomic, readwrite, strong) NSMutableOrderedSet *representationsRequested;
+
```
ric0liva, line 18: Do we need this commented out property? If not please remove.

BoxContentSDK/BoxContentSDK/Helper/BOXRepresentationsHelper.m
```diff
@@ -0,0 +1,99 @@
...
+/**
+ Ordered list of representation types, returns first representation type supported by the specified file type
+ */
```
ric0liva, line 21: The property is a boolean. but the comment says "ordered lists" which is confusing. We should improve the clarity of this comment.

BoxContentSDK/BoxContentSDK/Requests/BOXFileRequest.m
```diff
@@ -224,6 +221,14 @@ - (void)performRequestWithCached:(BOXFileBlock)cacheBlock
...
 }
 
+- (void)setRepresentationRequestOptions:(NSArray *)representationOptions {
+    [self.repsHelpers setRepresentationRequestOptions:representationOptions];
+}
+
+- (NSString *)formatRepresentationRequestHeader {
+    return [self.repsHelpers formatRepresentationRequestHeader];
+}
+
```
ric0liva, line 231: Do we need these functions or can we invoke the helper directly.  If we need them, perhaps they ought to be in some protocol.

BoxContentSDK/BoxContentSDK/BOXContentSDKConstants.h
```diff
@@ -515,3 +515,33 @@ typedef NS_ENUM(NSUInteger, BOXAvatarType) {
...
+    BOXRepresentationRequestAllRepresentations,
+    BOXRepresentationRequestHighDefinitionVideo,
+    BOXRepresentationRequestThumbnailRepresentation,
+    BOXRepresentationRequestLargeThumbnailRepresentation,
+    BOXRepresentationRequestPDFRepresentation,
+    BOXRepresentationRequestJPGRepresentation,
+    BOXRepresentationRequestMP3Representation,
+    BOXRepresentationRequestMP4Representation,
+    BOXRepresentationRequesteExtractedTextRepresentation
+} NS_ENUM_AVAILABLE_IOS(10_0);
```
jlawton, line 546: Probably don't need the availability attribute.

BoxContentSDK/BoxContentSDK/Helper/BOXRepresentationsHelper.h
```diff
@@ -0,0 +1,22 @@
...
+//  Copyright © 2018 Box. All rights reserved.
+//
+
+#import <Foundation/Foundation.h>
+
+@interface BOXRepresentationsHelper : NSObject
+/**
+ Setting a representation or list option will include availability of file with request representation information
+ */
+@property (nonatomic, readwrite, strong) NSMutableOrderedSet *representationsRequested;
```
jlawton, line 15: Seems like this could be `readonly`, since it's mutable. It strikes me as an anti-pattern exposing it as mutable, too.

jlawton: Of what is this a set?

BoxContentSDK/BoxContentSDK/Helper/BOXRepresentationsHelper.h
```diff
@@ -0,0 +1,22 @@
...
+
+/**
+ Setting a representation or Array of options will include availability for that file when request representation information
+ */
+- (void)setRepresentationRequestOptions:(NSArray *)representationOptions;
```
jlawton, line 20: Of what is this an array?

BoxContentSDK/BoxContentSDK/Helper/BOXRepresentationsHelper.m
```diff
@@ -0,0 +1,99 @@
...
+@property (nonatomic, readwrite, assign) BOOL matchSupportedRepresentation;
```
jlawton, line 22: What does this do?

BoxContentSDK/BoxContentSDK/Helper/BOXRepresentationsHelper.m
```diff
@@ -0,0 +1,99 @@
...
+    }
+    
+    representationFields = [representationFields stringByAppendingString:@"["];
+    
+    __block NSString *delimiter = @"],[";
+    if (self.matchSupportedRepresentation == YES) {
+        delimiter = @",";
+    }
+    
+    [self.representationsRequested enumerateObjectsUsingBlock:^(id obj, NSUInteger idx, BOOL *stop) {
```
jlawton, line 60: This whole thing is crazy. Can we just map the set of numbers to an array of strings and `componentsJoinedByString:`?

jlawton (commented):
So, this helper holds a mutable set, which it doesn't really "own" in any real sense, and it has a function `(Set, Bool) -> String`. Maybe it could not manage the set and just be a pure function, or it could mediate the interactions with the set, and be a real stateful object?

BoxContentSDK/BoxContentSDK/Helper/BOXRepresentationsHelper.m
Prithvijutur: This isn't new, it just moved around

BoxContentSDK/BoxContentSDK/Helper/BOXRepresentationsHelper.h
Prithvijutur: Again, None of this is new, just moved around.

PR #561, opened 2018-09-05
Add new cache function that does not depend on a box-ios-sdk request

Prithvijutur:
We have other requests that return files and it makes sense to provide a caching interface that is not tied into a request type


boxcla:

ric0liva (approved):

Prithvijutur:
Also reviewed by jLawton

PR #563, opened 2018-09-13
Rep hints fix

jlawton:
I expect this will get factored away soon, but we may as well be correct.


boxcla:

Prithvijutur (approved):

PR #565, opened 2018-09-25
Use representationHints, remove representation options

jlawton:


boxcla:

Prithvijutur (approved):
LGTM, You are going to make the corresponding changes to unbreak the app simultaneously?

BoxContentSDK/BoxContentSDK/Requests/BOXFileRequest.m
```diff
@@ -60,10 +58,9 @@ - (BOXAPIOperation *)createOperation
     if (self.requestAllFileFields) {
         fieldString = [self fullFileFieldsParameterString];
     }
-    
-    NSString *repsFieldString = [self.repsHelper getRepresentationsFieldString];
-    if(repsFieldString.length > 0) {
-        fieldString = [fieldString stringByAppendingFormat:@",%@", repsFieldString];
+
+    if(self.representationHints.length > 0) {
+        fieldString = [fieldString stringByAppendingFormat:@",%@,%@", BOXAPIObjectKeyRepresentations, BOXAPIObjectKeyAuthenticatedDownloadURL];
```
Prithvijutur, line 63: Why do we always want to request authenticated_downloadURL?

jlawton: We used to do that in `getRepresentationsFieldString` based on the contents of the representationOptions. We no longer have that signal. The idiomatic way to handle this would be to add authenticatedDownloadURL to allFields, and just get it all the time, but this was a more conservative change.

BoxContentSDK/BoxContentSDK/BOXContentSDKConstants.m
```diff
@@ -467,3 +467,9 @@
 NSString *const BOXURLSessionTaskCacheResponseData = @"responseData";
 NSString *const BOXURLSessionTaskCacheError = @"error";
 NSString *const BOXURLSessionTaskCacheUserIdAndAssociateId = @"userIdAndAssociateId";
+
+const BOXRepresentationHints BOXRepresentationHintsDefaultThumbnails = @"[jpg?dimensions=320x320][jpg?dimensions=1024x1024&paged=false]";
```
thuyBox: 320x320&paged=false?

BoxContentSDK/BoxContentSDK/BOXContentSDKConstants.h
```diff
@@ -515,33 +515,19 @@ typedef NS_ENUM(NSUInteger, BOXAvatarType) {
...
+///
+/// The exact definition of these values may change between SDK versions, but
+/// the semantics will remain.
+///
+/// For further information, refer to https://developer.box.com/v2.0/reference
+typedef NSString *BOXRepresentationHints;
+/// Large and small thumbnails (currently JPG thumbnails up to 1024px and 320px
+/// on a side, if available)
+extern const BOXRepresentationHints BOXRepresentationHintsDefaultThumbnails;
+/// A good preview for most file types (currently at least the first available of
```
thuyBox: should it be HLS if available and the first available of pdf,mp4,mp3,jpg?

jlawton: I don't want to guarantee that, and so I'd rather not explain it. What I've written is correct, if incomplete.

BoxContentSDK/BoxContentSDK/BOXContentSDKConstants.m
jlawton: 320 is never paged.

BoxContentSDK/BoxContentSDK/BOXContentSDKConstants.h
thuyBox: I think this suggests user only gets 1 representation, but that's not always the case. So maybe instead of "A", should be "At least one available preview ...."?

jlawton: "at least" covers the potential to get more than one, in the parens.

jlawton: Changed the comment a bit. I don't want to say "At least" in the main sentence, because no reps are really guaranteed. We might return zero. I know that the header is different from the returned reps, but people will equate them a bit.

BoxContentSDK/BoxContentSDK/BOXContentSDKConstants.m
```diff
@@ -467,3 +467,9 @@
...
+// We separate HLS here because HLS is only suitable for streaming, and there will be cases
+// we want to fall back on MP4
+const BOXRepresentationHints BOXRepresentationHintsDefaultPreview = @"[hls][pdf,mp4,mp3,jpg?paged=false]";
+const BOXRepresentationHints BOXRepresentationHintsDefaultThumbnailsandPreview = @"[jpg?dimensions=320x320][jpg?dimensions=1024x1024&paged=false][hls][pdf,mp4,mp3,jpg?paged=false]";
```
ahmedcharles: Are you sure that `jpg?paged=false` is what you want? Have you tested this alone and found that it returns the asset you want?

jlawton: Our preview SDK (with which this is generally used) doesn't yet handle paged reps, so this seems to be the right call for now. I'll test, but I'd expect this probably to get an unpaged jpg 2048 if available.

ahmedcharles: Ah, that rep doesn't exist, hence why I suggested the testing.

BoxContentSDK/BoxContentSDK/BOXContentSDKConstants.m
thuyBox: Box API Reference does not mention 320 is never paged. Even though that might be the case currently, if conversion decides to provide paging for 320 later, they could do that without considering that is a breaking change, and our app won't be able to handle it. We should use paged=false if we expect non-paged rep.

ahmedcharles: As I said to @jlawton earlier, we don't document the exact way that defaults are specified or the algorithm used when resolving the filter (hints) against specific representations. But we do guarantee (explicitly or implicitly) that the resolution is backwards compatible. This discussion reminded me to add documentation about this to help ensure that it remains a consideration going forward. So, if you request `jpg?dimensions=320x320` today and get back the same thing as giving `jpg?dimensions=320x320&paged=false`, then that is guaranteed to remain that way indefinitely.

That said, I also recommended that users always specify all properties because it's clearer about what they want and what they will get, so I ultimately agree, though for different reasons. The current code is right even if it is not as clear as it could be.

jlawton: Added explicit `paged=false` for 320 case.

thuyBox (approved):

PR #576, opened 2018-10-27
Avoid crash in release mode due to malformed URL

jlawton:


boxcla:

BoxContentSDK/BoxContentSDK/Categories/NSURL+BOXURLHelper.m
```diff
@@ -26,8 +26,9 @@ + (NSDictionary *)box_queryDictionaryWithString:(NSString *)queryString
     {
         keyValuePairComponents = [keyValuePair componentsSeparatedByString:@"="];
         BOXAssert(keyValuePairComponents.count == 2, @"Inconsistent information in keyValuePair=%@", keyValuePair);
-        [params setValue:[[keyValuePairComponents objectAtIndex:1] stringByRemovingPercentEncoding]
-                  forKey:[keyValuePairComponents objectAtIndex:0]];
+        if (keyValuePairComponents.count == 2) {
```
boxtcarpel, line 28: should we just add an else clause and put a BOXAssertFail in it?
Both patterns are fine, just asking to compare.

boxtcarpel (approved):

BoxContentSDK/BoxContentSDK/Categories/NSURL+BOXURLHelper.m
jlawton: That's an interesting point. Tempted to suggest that BOXAssert in SDKs is the wrong pattern, and BOXAssertFail is the better one, because then we're less likely to rely on it catching the error, which in release mode it doesn't.

jlawton: Or maybe even to call it BOXDebugAbort or something, but that's a more involved discussion.

PR #577, opened 2018-11-01
IOS-15844: allow encoding/decoding of BOXRepresentation

thuyBox:


boxcla:

BoxContentSDK/BoxContentSDK/Models/BOXRepresentation.h
```diff
@@ -9,7 +9,7 @@
 #import <Foundation/Foundation.h>
 #import "BOXContentSDKConstants.h"
 
-@interface BOXRepresentation : NSObject
+@interface BOXRepresentation : NSObject <NSCoding>
```
jlawton: Could we use `NSSecureCoding`? It only really requires using `decodeObjectOfClass:forKey:` below and implementing `+ (BOOL)supportsSecureCoding { return YES; }`.

BoxContentSDK/BoxContentSDK/Models/BOXRepresentation.m
```diff
@@ -87,4 +87,31 @@ - (instancetype)initWithJSON:(NSDictionary *)JSONResponse
     return self;
 }
 
+- (id)initWithCoder:(NSCoder *)coder
+{
+    self = [super init];
+    if (self) {
+        _type = [coder decodeObjectForKey:@"type"];
```
jlawton: Maybe also use `NSStringFromSelector` here to match below.

BoxContentSDK/BoxContentSDK/Models/BOXRepresentation.h
thuyBox: done

BoxContentSDK/BoxContentSDK/Models/BOXRepresentation.m
thuyBox: done

BoxContentSDK/BoxContentSDK/Models/BOXFile.h
```diff
@@ -92,7 +93,7 @@
 /**
  *  Array of available file conversion formats for original content
  */
-@property (nonatomic, readwrite, strong) NSArray *representations;
+@property (nonatomic, readwrite, strong) NSArray<BOXRepresentation *> *representations;
```
jlawton: Thanks :)

BoxContentSDK/BoxContentSDK/Models/BOXRepresentation.m
```diff
@@ -87,4 +87,36 @@ - (instancetype)initWithJSON:(NSDictionary *)JSONResponse
     return self;
 }
 
+- (id)initWithCoder:(NSCoder *)coder
+{
+    self = [super init];
+    if (self) {
+        _type = [coder decodeObjectOfClass:[BOXRepresentation class] forKey:NSStringFromSelector(@selector(type))];
```
jlawton: I think these need to be the classes of the things actually being decoded. So, `NSString.class` in this case.

thuyBox: done

BoxContentSDK/BoxContentSDK/Models/BOXRepresentation.m
```diff
@@ -87,4 +87,36 @@ - (instancetype)initWithJSON:(NSDictionary *)JSONResponse
...
+    self = [super init];
+    if (self) {
+        _type = [coder decodeObjectOfClass:[BOXRepresentationType class] forKey:NSStringFromSelector(@selector(type))];
+        _status = [coder decodeObjectOfClass:[BOXRepresentationStatus class] forKey:NSStringFromSelector(@selector(status))];
+        _statusCode = [coder decodeObjectOfClass:[NSString class] forKey:NSStringFromSelector(@selector((statusCode)))];
+        _infoURL = [coder decodeObjectOfClass:[NSURL class] forKey:NSStringFromSelector(@selector(infoURL))];
+        _contentURL = [coder decodeObjectOfClass:[NSURL class] forKey:NSStringFromSelector(@selector(contentURL))];
+        _details = [coder decodeObjectOfClass:[NSDictionary class] forKey:NSStringFromSelector(@selector(details))];
+        _dimensions = [coder decodeObjectOfClass:[BOXRepresentationImageDimensions class] forKey:NSStringFromSelector(@selector(dimensions))];
+        _JSONData = [coder decodeObjectOfClass:[NSDictionary class] forKey:NSStringFromSelector(@selector(JSONData))];
```
jlawton: The dictionaries might need the element classes also added to be successful.

thuyBox: don't think we can know for sure the subclasses in JSONDate or details dictionary. Without subclasses, that mean we can rely on Apple to decode to the best extend they can?

jlawton: I think we have to provide all the possibilities. We should test it. My guess is we want something like:

```
NSSet *jsonClasses =  [NSSet setFromArray:@[NSDictionary.class, NSArray.class, NSString.class, NSNumber.class, NSNull.class]];
[coder decodeObjectOfClasses:jsonClasses forKey:NSStringFromSelector(@selector(JSONData))];
```

jlawton:
I just want to make sure we're considered if we will ever want to be able to read the serialized version of the representations, and if so, JSON would be a better fit.

jlawton (approved):
Aside from the potential decision to use JSON, this looks good.

BoxContentSDK/BoxContentSDK/Models/BOXRepresentation.h
```diff
@@ -39,4 +39,11 @@
  */
 - (instancetype)initWithJSON:(NSDictionary *)JSONResponse;
 
+/**
+ *  Reconstruct JSONData from BOXRepresentation's properties
+ *
+ *  @return NSDictionary the reconstructed JSONData of this BOXRepresentation
+ */
+- (NSDictionary *)composeJSONData;
```
jlawton: How do you feel about saying `JSONData` when we actually mean a dictionary which can be serialized to JSON? I've always found that naming a bit confusing.

BoxContentSDK/BoxContentSDK/Models/BOXFile.h
jlawton: It should really be `copy`, too.

BoxContentSDK/BoxContentSDK/Models/BOXRepresentation.h
jlawton: Also, do you think it's important to differentiate between JSON we stored and JSON we construct? I'd rather move towards just making `JSONData` always construct. I think it's reasonable to distinguish it given where we are, but I think the current semantics of `JSONData` are not great.

BoxContentSDK/BoxContentSDK/Models/BOXRepresentation.m
```diff
@@ -87,4 +87,41 @@ - (instancetype)initWithJSON:(NSDictionary *)JSONResponse
...
+        JSONData[BOXAPIObjectKeyRepresentation] = self.type;
+    }
+    if (self.dimensions) {
+        JSONData[BOXAPIObjectKeyProperties] = @{BOXAPIObjectKeyDimensions: self.dimensions};
+    }
+    if (self.status || self.statusCode) {
+
+        NSMutableDictionary *statusDict = [NSMutableDictionary new];
+        if (self.status) {
+            statusDict[BOXAPIObjectKeyState] = self.status;
```
jlawton: I don't think we need any of these conditionals, since `x[y] = nil` is okay.

BoxContentSDK/BoxContentSDK/Models/BOXRepresentation.m
```diff
@@ -87,4 +87,41 @@ - (instancetype)initWithJSON:(NSDictionary *)JSONResponse
...
+        }
+        JSONData[BOXAPIObjectKeyStatus] = statusDict;
+    }
+    if (self.details) {
+        JSONData[BOXAPIObjectKeyDetails] = self.details;
+    }
+
+    NSString *contentURLString = self.contentURL.path;
+    if (contentURLString && [self.type isEqualToString:BOXRepresentationTypeHLS]) {
+        contentURLString = [contentURLString stringByReplacingOccurrencesOfString:BOXRepresentationTemplateValueHLSManifest withString:BOXRepresentationTemplateKeyAccessPath];
```
jlawton: Haha. Maybe we should just store the template in the first place, as well?

BoxContentSDK/BoxContentSDK/Models/BOXRepresentation.h
```diff
@@ -9,7 +9,7 @@
 #import <Foundation/Foundation.h>
 #import "BOXContentSDKConstants.h"
 
-@interface BOXRepresentation : NSObject
+@interface BOXRepresentation : NSObject <NSSecureCoding>
```
jlawton: We don't conform any more.

thuyBox: done

BoxContentSDK/BoxContentSDK/Models/BOXRepresentation.h
thuyBox: done

BoxContentSDK/BoxContentSDK/Models/BOXRepresentation.m
thuyBox: done. Keeping the check for cases like "@{BOXAPIObjectKeyURLTemplate: contentURLString};" which requires non-nil

BoxContentSDK/BoxContentSDK/Models/BOXRepresentation.m
thuyBox: want to preserve current behavior in this scope

jlawton: I contend that it wouldn't change any observable behavior.

BoxContentSDK/BoxContentSDKTests/BOXFileRequestTests.m
```diff
@@ -112,77 +112,4 @@ - (void)test_that_expected_file_is_returned_when_request_is_performed
     [self waitForExpectationsWithTimeout:5.0 handler:nil];
 }
 
-- (void)test_that_representation_options_are_set_correctly_when_request_is_performed
```
thuyBox, line 115: method setRepresentationRequestOptions no longer exists

BoxContentSDK/BoxContentSDK/Models/BOXRepresentation.m
```diff
@@ -87,4 +87,36 @@ - (instancetype)initWithJSON:(NSDictionary *)JSONResponse
...
+    JSONData[BOXAPIObjectKeyDetails] = self.details;
+
+    NSString *contentURLString = self.contentURL.absoluteString;
+    if (contentURLString && [self.type isEqualToString:BOXRepresentationTypeHLS]) {
+        contentURLString = [contentURLString stringByReplacingOccurrencesOfString:BOXRepresentationTemplateValueHLSManifest withString:BOXRepresentationTemplateKeyAccessPath];
+    }
+    if (contentURLString) {
+        JSONData[BOXAPIObjectKeyContent] = @{BOXAPIObjectKeyURLTemplate: contentURLString};
+    }
+    if (self.infoURL.path) {
```
jlawton: Odd to check for path, then set absoluteString.

BoxContentSDK/BoxContentSDK/Models/BOXFile.h
thuyBox: done

BoxContentSDK/BoxContentSDK/Models/BOXRepresentation.m
thuyBox: updated

BoxContentSDK/BoxContentSDK/Models/BOXRepresentation.m
thuyBox: updated

jlawton (approved):

jlawton (approved):

PR #584, opened 2019-01-29
Update App Auth documentation

mattwiller:
- Fixed Objective-C sample code
- Removed non-SDK-related code from sample
- Added equivalent Swift sample code

### DIFF
--- a/doc/BoxPlatform.md
+++ b/doc/BoxPlatform.md
@@ -38,14 +38,13 @@ ServerAuthFetchTokenBlock fetchTokens = ^(NSString *uniqueId, NSDictionary *fetc
     NSString *fullUrl = [NSString stringWithFormat:@"https://YOUR_SECURE_SERVER.example.com/token?id=%@", uniqueId];
     NSURL *url = [NSURL URLWithString:fullUrl];
 
-    __block NSString *token;
     NSURLSessionDataTask *task = [[NSURLSession sharedSession] dataTaskWithURL:url
             completionHandler:^(NSData *data, NSURLResponse *response, NSError *error) {
                 if (error) {
                     NSLog(@"Failed to retrieve new access token from server");
                     completion(nil, nil, error);
                 } else {
-                    token = [[NSString alloc] initWithData:data encoding:NSUTF8StringEncoding];
+                    NSString *token = [[NSString alloc] initWithData:data encoding:NSUTF8StringEncoding];
                     NSLog(@"Succeeded in retrieving new access token: %@", token);
 
                     // Pass token and expiration time back to the client
@@ -57,8 +56,8 @@ ServerAuthFetchTokenBlock fetchTokens = ^(NSString *uniqueId, NSDictionary *fetc
 
 // Create a client for the authenticated user; the client is keyed to that user's unique ID for caching purposes
 // The unique ID can be any identifier for the user, or a dummy value
-ServerAuthUser *user = [[ServerAuthUser alloc] initWithUniqueID:@"UNIQUE_ID"];
-BOXContentClient *boxClient = [BOXContentClient clientForServerAuthUser:user
+ServerAuthUser *authUser = [[ServerAuthUser alloc] initWithUniqueID:@"UNIQUE_ID"];
+BOXContentClient *boxClient = [BOXContentClient clientForServerAuthUser:authUser
                                                 initialToken:nil
                                                 fetchTokenBlockInfo:nil
                                                 fetchTokenBlock:fetchTokens];
@@ -89,7 +88,7 @@ let fetchToken: ServerAuthFetchTokenBlock = { (
     let url = URL(string: "https://YOUR_SECURE_SERVER.example.com/token?id=\(uniqueID ?? "")")!
     let task = URLSession.shared.dataTask(with: url) { data, response, error in
         
-        if error != nil {
+        guard error == nil else {
             NSLog("Failed to retrieve new access token from server")
             completion(nil, nil, error)
             return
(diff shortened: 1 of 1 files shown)


boxcla:
Verified that @mattwiller has signed the CLA. Thanks for the pull request!

doc/BoxPlatform.md
```diff
@@ -1,71 +1,133 @@
...
+    // NOTE: This code is specific to your app; it is responsible for retrieving new tokens from a secure server
+    NSLog(@"Attempting to retrieve new access token from server for uniqueId: %@", uniqueId);
 
-@interface AppDelegate ()
+    // Make a request to the secure sever for a token
+    NSString *fullUrl = [NSString stringWithFormat:@"https://YOUR_SECURE_SERVER.example.com/token?id=%@", uniqueId];
+    NSURL *url = [NSURL URLWithString:fullUrl];
 
-@property (nonatomic, readonly, strong) BOXContentClient *boxClient;
+    __block NSString *token;
```
jlawton: This can just be local inside the block. No need for a `__block` variable.

doc/BoxPlatform.md
```diff
@@ -1,71 +1,133 @@
...
+// The unique ID can be any identifier for the user, or a dummy value
+ServerAuthUser *user = [[ServerAuthUser alloc] initWithUniqueID:@"UNIQUE_ID"];
+BOXContentClient *boxClient = [BOXContentClient clientForServerAuthUser:user
+                                                initialToken:nil
+                                                fetchTokenBlockInfo:nil
+                                                fetchTokenBlock:fetchTokens];
 
-- (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(NSDictionary *)launchOptions {
+// Example of making an API call
+BOXUserRequest *request = [boxClient currentUserRequest];
```
jlawton, line 66: Do app users generally have a valid response to `/users/me`? I would think not. I know it's just an example.

doc/BoxPlatform.md
```diff
@@ -1,71 +1,133 @@
...
+```
 
-    _boxClient = [BOXContentClient clientForServerAuthUniqueId:@"A_UNIQUE_ID"
-                                                  initialToken:nil
-                                           fetchTokenBlockInfo:nil
-                                               fetchTokenBlock:fetchTokenBlock];                                                          
+__Swift:__
+```swift
+let fetchToken: ServerAuthFetchTokenBlock = { (
+        uniqueID: String?,
```
jlawton, line 79: My guess is that this shouldn't be Optional, but maybe that is indeed how it's being passed to Swift, if we never specified nullability.

doc/BoxPlatform.md
```diff
@@ -1,71 +1,133 @@
...
+        fetchTokenBlockInfo: [AnyHashable: Any]?,
+        completion: ((String?, Date?, Error?) -> Void)!
```
jlawton, line 81: Same here. I would think this shouldn't be an IUO, but again, it might be, if we didn't specify nullability.

doc/BoxPlatform.md
```diff
@@ -1,71 +1,133 @@
...
-        if (error){
-            NSLog(@"Error getting user info");
+    // NOTE: This code is specific to your app; it is responsible for retrieving new tokens from a secure server
+    NSLog("Attempting to retrieve new access token from server for unique ID: \(uniqueID)");
+    
+    // Make a request to the secure sever for a token
+    let url = URL(string: "https://YOUR_SECURE_SERVER.example.com/token?id=\(uniqueID ?? "")")!
+    let task = URLSession.shared.dataTask(with: url) { data, response, error in
+        
+        if error != nil {
```
jlawton: This is correct, but `if let error = error {  }` or `guard error == nil else {  }` might be more idiomatic. The first unwraps the error from the Optional. The second enforces that there's some form of return in the block. I guess I'd go with the second.

doc/BoxPlatform.md
```diff
@@ -1,71 +1,133 @@
...
+        guard let httpResponse = response as? HTTPURLResponse, httpResponse.statusCode == 200 else {
+            NSLog("Server returned incorrect response")
+            completion(nil, nil, Error("Token sever returned incorrect response"))
+            return
         }
-    }];
-
-    return YES;
+        
+        let token = String(data: data!, encoding: .utf8)!
```
jlawton, line 103: I suppose this is fine for sample code, but these `!` are invitations to crash, and I wouldn't do it in production.

doc/BoxPlatform.md
```diff
@@ -1,71 +1,133 @@
...
 }
 
-ServerAuthFetchTokenBlock myFetchTokenBlock = ^(NSString *uniqueId, NSDictionary *fetchTokenBlockInfo, void (^completion)(NSString *, NSDate *, NSError *))
-{
-    //this code is specific to your app; it is responsible for retrieving new tokens from a secure server
-    NSLog(@"Attempting to retrieve new access token from server for uniqueId: %@", unique);
+// Create a client for the authenticated user; the client is keyed to that user's unique ID for caching purposes
+// The unique ID can be any identifier for the user, or a dummy value
+let user = ServerAuthUser(uniqueID: "UNIQUE_ID")!
+let client = BOXContentClient.init(for: user, initialToken: nil, fetchTokenBlockInfo: nil, fetchTokenBlock: fetchToken)!
```
jlawton: Can delete the `.init` here.

doc/BoxPlatform.md
```diff
@@ -1,71 +1,133 @@
...
+
+// Example of making an API call
+client.currentUserRequest().perform { (user: BOXUser?, error: Error?) in
     
-    NSString *fullUrl = [NSString stringWithFormat:@"https://your.secure.server/token?id=%@", uniqueId];
-    NSURL *url = [NSURL URLWithString:fullUrl];
+    if error != nil {
```
jlawton: Same comment about using `guard` here.

doc/BoxPlatform.md
```diff
@@ -1,71 +1,133 @@
...
+    task.resume()
 }
 
-ServerAuthFetchTokenBlock myFetchTokenBlock = ^(NSString *uniqueId, NSDictionary *fetchTokenBlockInfo, void (^completion)(NSString *, NSDate *, NSError *))
-{
-    //this code is specific to your app; it is responsible for retrieving new tokens from a secure server
-    NSLog(@"Attempting to retrieve new access token from server for uniqueId: %@", unique);
+// Create a client for the authenticated user; the client is keyed to that user's unique ID for caching purposes
+// The unique ID can be any identifier for the user, or a dummy value
+let user = ServerAuthUser(uniqueID: "UNIQUE_ID")!
```
jlawton: It's probably clear, but maybe call this `authUser` or something, as `user` gets shadowed in the block below.

doc/BoxPlatform.md
mattwiller: App Users should have a valid response to `GET /users/me`, provided that their token isn't downscoped

doc/BoxPlatform.md
mattwiller: The compiler doesn't seem to like it if I make it non-optional, so my guess is that's how it's defined

jlawton (approved):

PR #585, opened 2019-02-04
Fix up App Users interface changes

mattwiller:
Some small changes to improve quality of life after the breaking changes around App Users:
* Added nullability annotations to fetch tokens block parameters
* Made `client.accessTokenDelegate` readonly, since it can no longer be set directly
* Updated App Users documentation code samples


boxcla:
Verified that @mattwiller has signed the CLA. Thanks for the pull request!

BoxContentSDK/BoxContentSDK/Clients/BOXContentClient.h
```diff
@@ -19,7 +19,7 @@
 @protocol BOXURLSessionManagerDelegate;
 @protocol UniqueSDKUser;
 
-typedef void (^ServerAuthFetchTokenBlock)(NSString *, NSDictionary *, void (^)(NSString *, NSDate *, NSError *));
+typedef void (^ServerAuthFetchTokenBlock)(NSString * _Nonnull, NSDictionary * _Nullable, void (^ _Nonnull)(NSString * _Nullable, NSDate * _Nullable, NSError * _Nullable));
```
jlawton: We could even write: `typedef void (^ServerAuthFetchTokenBlock)(NSString * _Nonnull uniqueID, NSDictionary * _Nullable userInfo, void (^ _Nonnull completion)(NSString * _Nullable, NSDate * _Nullable, NSError * _Nullable));` and users will get the default parameter names autocompleted.

jlawton: Right now, I find this declaration inscrutable.

jlawton: That's true of the completion block, too, and we could add names there, but I don't think users will generally see them.

mattwiller:
@jlawton Thanks for the review!  I added some names to those parameters — can you verify that those are in line with Objective-C naming conventions?  Also, I just wanted to check with you whether adding these nullability annotations is considered backwards compatible — is it safe to put those in at this point?

mattwiller:
Also, any idea why the CI checks are erroring?  Is there something I should do to get those to run?

jlawton:
The names look good.

As for backwards compatibility, it's not technically compatible (this is one reason we've never fixed the nullability across the board). In Objective-C, consuming code doesn't change, but in Swift, as you've done in the sample, you have to change the uses of uniqueID. I think this is small enough, and actually simplifies use, and people mostly won't mind, but that's a call to be made. Requiring the block should be fine, because it doesn't change syntax, and just enforces the contract up front, rather than getting a bug later.

CI seems not to be able to find the right simulator to run. Maybe updating `.travis.yml` to use one of the simulators from the error message might get it to work, though.

BoxContentSDK/BoxContentSDK/Clients/BOXContentClient.h
```diff
@@ -199,6 +199,15 @@ extern NSString *const BOXContentClientBackgroundTempFolder;
  **/ 
 - (void)setSharedLinkStorageDelegate:(id <BOXSharedLinkStorageProtocol>)delegate;
 
+/**
+ * Sets the access token delegate for an App Auth client.
+ *
+ * @param accessTokenDelegate  The delegate responsible for fetching new tokens when necessary.
+ * @param serverAuthUser       The user for which this client should be authenticated.
```
jdizoglio: Maybe update description:
"The user for which this client is authenticated" or "Currently authenticated user for this client"

jdizoglio (approved):
See minor comment

PR #587, opened 2019-02-07
IOS-16730: allow caching item responses from shared link creation req…

thuyBox:
…uests

### DIFF
--- a/BoxContentSDK/BoxContentSDK/Requests/BOXBookmarkShareRequest.h
+++ b/BoxContentSDK/BoxContentSDK/Requests/BOXBookmarkShareRequest.h
@@ -7,7 +7,7 @@
 
 @interface BOXBookmarkShareRequest : BOXItemShareRequest
 
-@property (nonatomic, readonly, strong) NSString *bookmarkID;
+@property (nonatomic, readonly, copy) NSString *bookmarkID;
 
 - (instancetype)initWithBookmarkID:(NSString *)bookmarkID;
 - (void)performRequestWithCompletion:(BOXBookmarkBlock)completionBlock;
--- a/BoxContentSDK/BoxContentSDK/Requests/BOXBookmarkShareRequest.h
+++ b/BoxContentSDK/BoxContentSDK/Requests/BOXBookmarkShareRequest.h
@@ -7,6 +7,8 @@
 
 @interface BOXBookmarkShareRequest : BOXItemShareRequest
 
+@property (nonatomic, readonly, strong) NSString *bookmarkID;
+
 - (instancetype)initWithBookmarkID:(NSString *)bookmarkID;
 - (void)performRequestWithCompletion:(BOXBookmarkBlock)completionBlock;
 
--- a/BoxContentSDK/BoxContentSDK/Requests/BOXBookmarkUnshareRequest.m
+++ b/BoxContentSDK/BoxContentSDK/Requests/BOXBookmarkUnshareRequest.m
@@ -12,7 +12,7 @@
 
 @interface BOXBookmarkUnshareRequest ()
 
-@property (nonatomic, readwrite, strong) NSString *bookmarkID;
+@property (nonatomic, readwrite, copy) NSString *bookmarkID;
 
 @end
 
--- a/BoxContentSDK/BoxContentSDK/Requests/BOXBookmarkUnshareRequest.m
+++ b/BoxContentSDK/BoxContentSDK/Requests/BOXBookmarkUnshareRequest.m
@@ -60,14 +60,24 @@
     BOXAPIJSONOperation *bookmarkOperation = (BOXAPIJSONOperation *)self.operation;
 
     bookmarkOperation.success = ^(NSURLRequest *request, NSHTTPURLResponse *response, NSDictionary *JSONDictionary) {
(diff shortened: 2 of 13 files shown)


boxcla:

BoxContentSDK/BoxContentSDK/Requests/BOXBookmarkShareRequest.h
```diff
@@ -7,6 +7,8 @@
 
 @interface BOXBookmarkShareRequest : BOXItemShareRequest
 
+@property (nonatomic, readonly, strong) NSString *bookmarkID;
```
jlawton: All these strings should really be copied.

thuyBox: done

jlawton (approved):

BoxContentSDK/BoxContentSDK/Requests/BOXBookmarkUnshareRequest.m
```diff
@@ -60,14 +60,24 @@ - (void)performRequestWithCompletion:(BOXBookmarkBlock)completionBlock
...
                 completionBlock(bookmark, nil);
             } onMainThread:isMainThread];
         }
     };
     bookmarkOperation.failure = ^(NSURLRequest *request, NSHTTPURLResponse *response, NSError *error, NSDictionary *JSONDictionary) {
+        if ([self.cacheClient respondsToSelector:@selector(cacheBookmarkUnshareRequest:withBookmark:error:)]) {
+            [self.cacheClient cacheBookmarkUnshareRequest:self
+                                             withBookmark:nil
+                                                    error:error];
+        }
```
potrebic, line 80: It seems like in general the code is very duplicative. L76-89 are just like L63-68, if a `bookmark` and `error` ivars were initialized to `nil` or parameterized in a helper method.

What leads to these sorts of patterns? Is it the nature of the language (Obj-C) or that nature of the iOS frameworks? Or is it the patterns we (Box) have developed?

BoxContentSDK/BoxContentSDK/Requests/BOXFileShareRequest.m
```diff
@@ -110,14 +110,24 @@ - (void)performRequestWithCompletion:(BOXFileBlock)completionBlock
...
                 completionBlock(file, nil);
             } onMainThread:isMainThread];
         }
     };
     fileOperation.failure = ^(NSURLRequest *request, NSHTTPURLResponse *response, NSError *error, NSDictionary *JSONDictionary) {
+        if ([self.cacheClient respondsToSelector:@selector(cacheFileShareRequest:withFile:error:)]) {
+            [self.cacheClient cacheFileShareRequest:self
+                                           withFile:nil
+                                              error:error];
+        }
```
potrebic, line 130: idem. More duplication. I understand that duplication already existed before this change. This change is just following the existing patterns.

BoxContentSDK/BoxContentSDK/Requests/BOXFolderUnshareRequest.m
```diff
@@ -72,6 +77,11 @@ - (void)performRequestWithCompletion:(BOXFolderBlock)completionBlock
         }
     };
     folderOperation.failure = ^(NSURLRequest *request, NSHTTPURLResponse *response, NSError *error, NSDictionary *JSONDictionary) {
+        if ([self.cacheClient respondsToSelector:@selector(cacheFolderUnshareRequest:withFolder:error:)]) {
+            [self.cacheClient cacheFolderUnshareRequest:self
+                                             withFolder:nil
+                                                  error:error];
+        }
```
potrebic, line 84: idem

BoxContentSDK/BoxContentSDK/Requests/BOXFolderShareRequest.m
```diff
@@ -108,14 +108,24 @@ - (void)performRequestWithCompletion:(BOXFolderBlock)completionBlock
...
                 completionBlock(folder, nil);
             } onMainThread:isMainThread];
         }
     };
     folderOperation.failure = ^(NSURLRequest *request, NSHTTPURLResponse *response, NSError *error, NSDictionary *JSONDictionary) {
+        if ([self.cacheClient respondsToSelector:@selector(cacheFolderShareRequest:withFolder:error:)]) {
+            [self.cacheClient cacheFolderShareRequest:self
+                                           withFolder:nil
+                                                error:error];
+        }
```
potrebic, line 128: idem

BoxContentSDK/BoxContentSDK/Requests/BOXBookmarkUnshareRequest.m
jlawton: I think it mainly stems from wanting different params to success and failure blocks, because Objective-C can't express sum types. I'd love to see a different pattern in non-objective-c-supporting Swift. I agree that this could be abstracted internally here to reduce duplication.

thuyBox: I think the main reason is for clarity. There are some cases we would need different logic in success and failure block. Maybe cleaner to have the shared block if we need to extend the shared logic further.

PR #588, opened 2019-02-08
Request a BOXUser object by user ID

jdizoglio:


boxcla:

BoxContentSDK/BoxContentSDK/BOXContentSDKConstants.m
```diff
@@ -107,6 +107,8 @@
 BOXSharedLinkAccessLevel *const BOXSharedLinkAccessLevelOpen = @"open";
 BOXSharedLinkAccessLevel *const BOXSharedLinkAccessLevelCompany = @"company";
 BOXSharedLinkAccessLevel *const BOXSharedLinkAccessLevelCollaborators = @"collaborators";
+BOXSharedLinkAccessLevel *const BOXSharedLinkExpireLinkFileAccess = @"expireLinkFiles";
```
thuyBox: don't seem like we use them?

BoxContentSDK/BoxContentSDK/BOXContentSDKConstants.h
```diff
@@ -117,6 +117,8 @@ typedef NSString BOXSharedLinkAccessLevel;
 extern BOXSharedLinkAccessLevel *const BOXSharedLinkAccessLevelOpen;
 extern BOXSharedLinkAccessLevel *const BOXSharedLinkAccessLevelCompany;
 extern BOXSharedLinkAccessLevel *const BOXSharedLinkAccessLevelCollaborators;
+extern BOXSharedLinkAccessLevel *const BOXSharedLinkExpireLinkFileAvailable;
```
boxtcarpel: These aren't quite shared link access levels...
Should these just be separate constants specific to the expiration date?

boxtcarpel: In fact, these actually aren't used anywhere either, can we remove them?

BoxContentSDK/BoxContentSDK/BOXContentSDKConstants.m
boxtcarpel: This constant name doesn't actually match the definition above, and is not used anywhere. Can we just remove these?

BoxContentSDK/BoxContentSDK/Requests/BOXUserRequest.h
```diff
@@ -14,6 +14,10 @@
  */
 @property (nonatomic, readwrite, assign) BOOL requestAllUserFields;
 
+@property (nonatomic, readonly, strong) NSString *userID;
```
boxtcarpel, line 17: Should we leave a comment that this property only reports a value if a specific user is requested, and that in the default case where no user is specified the current user is used (but this value will be nil or an empty string)?

boxtcarpel (approved):

BoxContentSDK/BoxContentSDK/BOXContentSDKConstants.m
jdizoglio: removed

BoxContentSDK/BoxContentSDK/BOXContentSDKConstants.h
jdizoglio: removed

BoxContentSDK/BoxContentSDK/BOXContentSDKConstants.m
jdizoglio: removed

BoxContentSDK/BoxContentSDK/Requests/BOXUserRequest.h
jdizoglio: Good point, for a manual construction of the request, setting the userID will have the same impact as the initWithUserID

PR #598, opened 2019-03-20
IOS-17005 - fix linking of the ContentSDKTestFramework

potrebic:
Needed to add the -ObjC linking option. I learned then when attempting
to use the ContentSDKTestFramework from another target (BoxToolkitTests)
and a test hit a runtime error when trying to use OCMock. And the runtime
error actually gave a super useful error syaing to try to -ObjC error!


boxcla:

jlawton (approved):

PR #603, opened 2019-03-27
add unit test for the just fixed default invitee role bug

potrebic:


boxcla:

jlawton (approved):

potrebic:
Of note is that I confirmed the new test fails prior to James fix.

PR #612, opened 2019-06-02
IOS-17254: don't crash if API response contains unexpected null.

Jeff-Meadows:
When we request an item from the Box API, we request the
collections and collection_memberships fields. The collections
field is an array of collection objects, and the memberships
field is an array of collection membership objects, which have a
collection property which is a collection object.

Recently, a backend bug caused some API responses to contain
null instead of a collection object, either in the collections
array or as the collection property of a collection memberhip.

Most of our JSON parsing code deals with dictionaries, and we have
code designed to handle unexpected nulls, but in these cases,
we weren't handling them, so our app crashed when it received
such a response.

This commit skips trying to parse any null collection objects,
ensuring that if we get a malformed response like that in the
future, we won't crash.

### DIFF
--- a/BoxContentSDK/BoxContentSDK/Models/BOXItem.m
+++ b/BoxContentSDK/BoxContentSDK/Models/BOXItem.m
@@ -242,10 +242,11 @@
                                                     nullAllowed:NO];
         
         // Parse common collections into a BOXCollection
-        NSArray *collectionsJSONArray = [NSJSONSerialization box_ensureObjectForKey:BOXAPIObjectKeyCollections 
-                                                                       inDictionary:JSONResponse
-                                                                    hasExpectedType:[NSArray class]
-                                                                        nullAllowed:YES];
+        NSArray *collectionsJSONArray = [NSJSONSerialization box_ensureObjectForKey:BOXAPIObjectKeyCollections
+                                                                        inDictionary:JSONResponse
+                                                                     hasExpectedType:[NSArray class]
+                                                                         nullAllowed:YES
+                                                                   suppressNullAsNil:YES];
         NSMutableArray *collections = [NSMutableArray arrayWithCapacity:collectionsJSONArray.count];
         for (NSDictionary *dict in collectionsJSONArray) {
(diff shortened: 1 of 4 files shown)


boxcla:
Verified that @Jeff-Meadows has signed the CLA. Thanks for the pull request!

mattwiller:
+1 from me, but I'll defer to someone more familiar with the codebase to approve

BoxContentSDK/BoxContentSDK/Models/BOXItem.m
```diff
@@ -248,7 +248,9 @@ - (instancetype)initWithJSON:(NSDictionary *)JSONResponse
                                                                         nullAllowed:YES];
         NSMutableArray *collections = [NSMutableArray arrayWithCapacity:collectionsJSONArray.count];
         for (NSDictionary *dict in collectionsJSONArray) {
-            [collections addObject:[[BOXCollection alloc] initWithJSON:dict]];
+            if (![dict isEqual:[NSNull null]]) {
+                [collections addObject:[[BOXCollection alloc] initWithJSON:dict]];
```
jlawton, line 253: Do you think, while we're at it, we might want to check that we're not getting `nil` from `initWithJSON:`? Because adding that to an array would also crash.

Jeff-Meadows: As far as I can tell, that can't happen unless we change our code to make it happen, which would be caught by unit tests. Is that understanding inaccurate?

BoxContentSDK/BoxContentSDK/Models/BOXItem.m
```diff
@@ -248,7 +248,9 @@ - (instancetype)initWithJSON:(NSDictionary *)JSONResponse
                                                                         nullAllowed:YES];
```
boxtcarpel: we're doing "nullAllowed" here, however we're not handling the scenario where we pull in an NSNull value instead of a valid array. Should we handle that? or "supressNullAsNil"?

boxtcarpel (approved):

PR #622, opened 2019-07-30
Release memory each time we're finished with a chunk

jlawton:
We're using an autoreleased data object as part of preparing the file for upload, so we need to clean it up each time round the loop, or we end up loading the entire contents of the file into memory at once.


boxcla:

thuyBox (approved):

PR #630, opened 2019-08-05
IOS-17410 removed unusued md5HashOfFileAtPath method

MashfiqueA:
Removed a method which was flagged by Data Theorem for using md5.


boxcla:
Hi @anwar722, thanks for the pull request. Before we can merge it, we need you to sign our Contributor License Agreement. You can do so electronically here: http://opensource.box.com/cla

mattwiller:
@anwar722 This change doesn't look like it's backwards-compatible; could we deprecate this method instead of removing it?  If users are calling this method in their applications, we don't want to break them.

PR #632, opened 2019-08-07
IOS-17416 updated NSCoding conforming classes to NSSecureCoding

MashfiqueA:
Classes conforming to NSCoding were flagged by Data Theorem as unsafe.

### DIFF
--- a/BoxContentSDK/BoxContentSDK/AppToApp/BOXAppToAppMessage.m
+++ b/BoxContentSDK/BoxContentSDK/AppToApp/BOXAppToAppMessage.m
@@ -333,10 +333,10 @@
     if (self = [self init])
     {
         _messageType = [aDecoder decodeInt32ForKey:@"messageType"];
-        _senderApplication = [aDecoder decodeObjectForKey:@"senderApplication"];
-        _receiverApplication = [aDecoder decodeObjectForKey:@"receiverApplication"];
-        _action = [aDecoder decodeObjectForKey:@"action"];
-        _fileMetadata = [aDecoder decodeObjectForKey:@"metadata"];
+        _senderApplication = [aDecoder decodeObjectOfClass:[BOXAppToAppApplication class] forKey:@"senderApplication"];
+        _receiverApplication = [aDecoder decodeObjectOfClass:[BOXAppToAppApplication class] forKey:@"receiverApplication"];
+        _action = [aDecoder decodeObjectOfClass:[NSString class] forKey:@"action"];
+        _fileMetadata = [aDecoder decodeObjectOfClass:[BOXAppToAppFileMetadata class] forKey:@"metadata"];
     }
 
     return self;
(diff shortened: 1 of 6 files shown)


boxcla:
Verified that @anwar722 has signed the CLA. Thanks for the pull request!

mattwiller:
Just to verify, since `NSSecureCoding` inherits from `NSCoding`, this should be backwards-compatible, right?

jlawton (commented):
Matt, yes it will be backwards compatible. @anwar722 I think we also need to change the decode methods to use secure decoding.

BoxContentSDK/BoxContentSDK/AppToApp/BOXAppToAppMessage.m
```diff
@@ -333,13 +333,17 @@ - (id)initWithCoder:(NSCoder *)aDecoder
...
     {
         _messageType = [aDecoder decodeInt32ForKey:@"messageType"];
-        _senderApplication = [aDecoder decodeObjectForKey:@"senderApplication"];
-        _receiverApplication = [aDecoder decodeObjectForKey:@"receiverApplication"];
-        _action = [aDecoder decodeObjectForKey:@"action"];
-        _fileMetadata = [aDecoder decodeObjectForKey:@"metadata"];
+        _senderApplication = [aDecoder decodeObjectOfClass:[BOXAppToAppApplication class] forKey:@"senderApplication"];
+        _receiverApplication = [aDecoder decodeObjectOfClass:[BOXAppToAppApplication class] forKey:@"receiverApplication"];
+        _action = [aDecoder decodeObjectOfClass:[NSString class] forKey:@"action"];
+        _fileMetadata = [aDecoder decodeObjectOfClass:[BOXAppToAppFileMetadata class] forKey:@"metadata"];
```
jlawton, line 339: Does this class support secure coding?

MashfiqueA: Not sure what you mean, can you please clarify a little more?

jlawton: if `BOXAppToAppFileMetadata` doesn't correctly implement secure decoding, then this class can't possibly be secure in its decoding.

MashfiqueA: `BOXAppToAppFileMetadata` contains some NSString properties which are put into a dictionary using the two `appToAppFileMetadataWithFileName` methods to form the metadata. This metadata dictionary is what's encoded/decoded inside that class.

So I guess we can safely say `BOXAppToAppFileMetadata` is safely decoded or implements safe decoded?

jlawton: It must conform to `NSSecureCoding` and use the appropriate decoding methods, or it's not secure.

MashfiqueA: Caught up offline. `BOXAppToAppFileMetadata` conforms to `NSSecureCoding` and implements the `supportsSecureCoding` method.

jlawton (approved):

PR #634, opened 2019-08-09
added method to generate sha512 hash from NSString

MashfiqueA:


boxcla:
Verified that @anwar722 has signed the CLA. Thanks for the pull request!

jlawton (approved):

PR #639, opened 2019-08-19
IOS-17522 - remove unnecessary code, which elims a race condition

potrebic:
The `outputStreamHasSpaceAvailable` state was an optimization to
avoid acquring a mutex and calling NSOutputStream.hasSpaceAvailable.
Logic using `outputStreamHasSpaceAvailable` wasn't properly contained
within a mutex, so it was open to race conditions. I don't suspect
this race condition was actually triggering this bug (that will
come in another commit). But this change simplifies the code and
removes a race condition.

### DIFF
--- a/BoxContentSDK/BoxContentSDK/Operations/BOXAPIDataOperation.m
+++ b/BoxContentSDK/BoxContentSDK/Operations/BOXAPIDataOperation.m
@@ -21,11 +21,6 @@
 // output stream from this NSData when space becomes availble
 @property (nonatomic, readwrite, strong) NSMutableData *receivedDataBuffer;
 
-// The output stream may trigger the has space available callback when no data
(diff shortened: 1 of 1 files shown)


boxcla:

MashfiqueA (approved):

jlawton (approved):

mattwiller:
@potrebic Is this change expected to be backwards compatible?  Could a third-party developer have been relying on the property that's being removed?

jlawton:
That property was private, so we are okay in compatibility.

thuyBox:
@potrebic could we verify this change with:
- regular downloads from app (preview+offline)
- background downloads from File Provider

potrebic:
@thuyBox I tried the first 2 things you suggested. I'll do the file provider test now.

thuyBox:
thanks, @potrebic

PR #640, opened 2019-08-19
IOS-17522 - clear receivedDataBuffer when outputStream is cleared

potrebic:
IOS-17522 shows writeDataToOutputStream accessing freed memory. This is an
attempt to avoid that. This fix is speculative in that I haven't been able
to repro the same crash. I have confirmed that writeDataToOutputStream
will gracefully fail if it executed with receivedDataBuffer==nil.

### DIFF
--- a/BoxContentSDK/BoxContentSDK/Operations/BOXAPIDataOperation.m
+++ b/BoxContentSDK/BoxContentSDK/Operations/BOXAPIDataOperation.m
@@ -212,6 +212,7 @@
 
 - (void)dealloc
 {
+    self.receivedDataBuffer = nil;
     if (self.outputStream) {
         self.outputStream.delegate = nil;
         [self.outputStream close];


boxcla:

MashfiqueA (approved):

BoxContentSDK/BoxContentSDK/Operations/BOXAPIDataOperation.m
```diff
@@ -212,6 +212,7 @@ - (void)finish
 
 - (void)dealloc
 {
+    self.receivedDataBuffer = nil;
```
jlawton, line 215: This isn't harmful, but should be unnecessary. Did this appear to fix something?

potrebic: As I noted... I could never repro the exact issue.

jlawton (approved):

BoxContentSDK/BoxContentSDK/Operations/BOXAPIDataOperation.m
potrebic: James and I chatted offline. And I shared my sketchy theories.  James agreed that it that scenario this might help (if some other code is doing something unsafe). So this change wouldn't fix *that* problem, but could mitigate the outcome by avoiding a crash here.

PR #641, opened 2019-08-19
Cherry pick of the 2 commits addressing IOS-17522

potrebic:


boxcla:

BoxContentSDK/BoxContentSDK/Operations/BOXAPIDataOperation.m
```diff
@@ -303,15 +297,7 @@ - (void)sessionTask:(NSURLSessionTask *)sessionTask processIntermediateData:(NSD
             [self.receivedDataBuffer appendData:data];
         }
 
-        // If the output stream does have space available, trigger the writeDataToOutputStream
```
thuyBox, line 306: I have concern that removing this check completely will increase the number of times writeDataToOutputStream being called, thus, [self.outputStream hasSpaceAvailable] will be called many times unnecessarily (not sure what the cost of this check is), instead of called once we know that there are space available through stream:handleEvent of NSStreamEventHasSpaceAvailable.

potrebic:
Flaky tests! Argh:
test_that_expected_folder_is_returned_when_request_is_performed
test_that_forbidden_403_error_does_not_trigger_logout_notification_for_non_token_requests

BoxContentSDK/BoxContentSDK/Operations/BOXAPIDataOperation.m
potrebic: As I noted in the commit message in the original change, this is trading off "code complexity & race condition" for some perceived (but not documented?) optimization. In my experiments with my change I found the following:

A download is triggered. `processIntermediateData` is called N times. *EACH* of those N times `hasSpaceAvailable` on the OutputStream is true. Thus there was really nothing to optimize. Example - for a 159KB PDF I saw `process` being called 39 times. And `hasSpaceAvailable` was true for each call. For a 4Mb jpeg it was 139/139.

So in fact... to me it seems that optimization was actually counter-productive. 

Further, `hasSpaceAvailable` is a read-only BOOL property. The documentation doesn't say anything about this being a complex calculation. Having a complex calculation behind a read-only BOOL property would be a bad design. In the commits that added the complexity of `outputStreamHasSpaceAvailable` (the code being removed now) don't mention any specific perf issues... I suspect that those old changes were a case of premature/misguided optimization.  This is why commit messages are so important and should document the "why" and not the "what".

If you have more specific information/data that there is some perf issue please share.

potrebic:
flaky test: test_that_expected_file_is_returned_when_request_is_performed

jlawton (approved):

thuyBox:
+1

PR #645, opened 2019-08-23
Fix a flaky test - test_that_forbidden_403_error_does_not_trigger_log…

potrebic:
…out_notification_for_non_token_requests

This test was flaky due to a subtle (and still not well understood)
interaction between OCMock's 'reject' method and the BOXAPIOperation
finish method. This resulted in random crashes in the finish method.

The solution was to use a BOXAPIOperation `spy` implementation allowing
a test to block until the finish method is *completely* finished. This
technique eliminates the random crashing.

Comments in the code have more details.

### DIFF
--- a/BoxContentSDK/BoxContentSDKTests/BOXRequestTests.m
+++ b/BoxContentSDK/BoxContentSDKTests/BOXRequestTests.m
@@ -50,6 +50,34 @@
 
 @end
 
+// For use by tests that need to track when [BOXAPIOperation finish] has
+// completed. There are subtle (and really not-well-understood) situations
+// where a test completes (the test method 'returns') and after that
+// the 'finish' method is still executing. Most tests are naturally keyed
+// of the call `[self performCompletionCallback];` in BOXAPIOperation,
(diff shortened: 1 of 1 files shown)


boxcla:

BoxContentSDK/BoxContentSDKTests/BOXRequestTests.m
```diff
@@ -222,8 +262,6 @@ - (void)test_that_forbidden_403_error_does_not_trigger_logout_notification_for_n
         [expectation fulfill];
     }];
     [self waitForExpectationsWithTimeout:2.0 handler:nil];
-
-    [operationMock verify];
```
potrebic, line 226: NOTE - this was removed because it was unnecessary. The `reject` call above sets up a fail-fast scenario, when the test is failed immediately if `sendLogoutNotification` is called. And there are no other OCMock expectations... thus calling verify is a no-op. So I removed. All this is unrelated to fixing the flakiness of this test.

jlawton (approved):

MashfiqueA (approved):

PR #648, opened 2019-09-05
IOS-17361: support immediate API request execution

thuyBox:
Pre-requisite: PR https://github.com/box/box-ios-sdk/pull/649

Currently, executing request will enqueue an API Operation to a network queue for execution.

Adding support for immediate API request execution allows the network request to be executed immediately without waiting on the queue. This is useful for cases where the caller needs control of timing of network request execution.

- Add BOXRequest's shouldPerformRequestImmediately option: to allow API network request to start immediately instead of being put onto a network queue
- Add BoxContentClient's shouldBackgroundRequestsPerformImmediately option: to support marking all background requests' API network requests to be executed immediately instead of being put onto a network queue

### DIFF
--- a/BoxContentSDK/BoxContentSDK/Clients/BOXContentClient.h
+++ b/BoxContentSDK/BoxContentSDK/Clients/BOXContentClient.h
@@ -86,6 +86,12 @@ extern NSString *const BOXContentClientBackgroundTempFolder;
 @property (nonatomic, nullable, readwrite, copy) NSDictionary *fetchTokenBlockInfo;
 
 /**
+ *  Used to set background requests' shouldPerformRequestImmediately provided those requests prepared
+ *  using BOXContentClient. E.g.: BOXFileRequest returned by fileInfoRequestWithID:associateID:
+ */
+@property (nonatomic, readwrite, assign) BOOL shouldBackgroundRequestsPerformImmediately;
+
+/**
  *  The list of Box users that have established a session through the SDK.
  *
  *  @return array of BOXUserMini model objects
--- a/BoxContentSDK/BoxContentSDK/Clients/BOXContentClient.m
+++ b/BoxContentSDK/BoxContentSDK/Clients/BOXContentClient.m
@@ -23,6 +23,7 @@
 #import "BOXUserRequest.h"
 #import "BOXContentClient+User.h"
 #import "BOXURLSessionManager.h"
+#import "BOXRequest.h"
 
 // Default API URLs
 /*
@@ -219,6 +220,7 @@ static BOXContentClient *defaultInstance = nil;
                                                                queueManager:_queueManager
                                                           urlSessionManager:_urlSessionManager];
         _queueManager.session = self.session;
+        _shouldBackgroundRequestsPerformImmediately = NO;
         
         if (staticRedirectURIString.length > 0) {
             _OAuth2Session.redirectURIString = staticRedirectURIString;
@@ -495,10 +497,22 @@ static BOXContentClient *defaultInstance = nil;
         BOXSharedItemRequest *shareItemRequest = (BOXSharedItemRequest *)request;
         shareItemRequest.sharedLinkHeadersHelper = self.sharedLinksHeaderHelper;
     }
-    
+
+    // If the request is currently not set to perform immediately,
+    // Check if we should set it to perform immediately or not
+    if (request.shouldPerformRequestImmediately == NO) {
+       request.shouldPerformRequestImmediately = self.shouldBackgroundRequestsPerformImmediately
+        && [self isBackgroundRequest:request];
+    }
     if (self.OAuth2Session.refreshToken && (self.OAuth2Session.clientID.length == 0 || self.OAuth2Session.clientSecret.length == 0)) {
         [NSException raise:@"Set client ID and client secret first." format:@"You must set a client ID and client secret first."];
     }
+}
+
+- (BOOL)isBackgroundRequest:(BOXRequest *)request
+{
+    return [request conformsToProtocol:@protocol(BOXBackgroundRequestProtocol)]
+        && [(id<BOXBackgroundRequestProtocol>)request associateID] != nil;
 }
 
 + (NSMutableDictionary *)SDKClients
--- a/BoxContentSDK/BoxContentSDK/Operations/BOXAPIOperation.m
+++ b/BoxContentSDK/BoxContentSDK/Operations/BOXAPIOperation.m
@@ -113,6 +113,7 @@ static BOOL BoxOperationStateTransitionIsValid(BOXAPIOperationState fromState, B
         _body = [POSTParams copy];
         _queryStringParameters = queryParams;
         _session = session;
+        _shouldStartImmediately = NO;
         // delay setting up the session task as long as possible so the authentication credentials remain fresh
 
         _APIRequest = nil;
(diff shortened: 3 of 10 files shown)


boxcla:

BoxContentSDK/BoxContentSDK/Clients/BOXContentClient.m
```diff
@@ -495,7 +497,10 @@ - (void)prepareRequest:(BOXRequest *)request
         BOXSharedItemRequest *shareItemRequest = (BOXSharedItemRequest *)request;
         shareItemRequest.sharedLinkHeadersHelper = self.sharedLinksHeaderHelper;
     }
-    
+    if ([request conformsToProtocol:@protocol(BOXBackgroundRequestProtocol)]) {
```
jlawton: I think this should maybe check for `self.shouldBackgroundRequestsPerformImmediately` before even checking the request, and probably shouldn't override if the user has set `request. shouldPerformRequestImmediately` some other way.

BoxContentSDK/BoxContentSDK/Operations/BOXAPIOperation.m
```diff
@@ -297,23 +298,32 @@ - (BOOL)isFinished
...
+            // they grab the lock and are adding dependencies.
+            self.state = BOXAPIOperationStateExecuting;
 
-        [self performSelector:@selector(executeOperation) onThread:[[self class] globalAPIOperationNetworkThread] withObject:nil waitUntilDone:NO];
-    }
-    else
-    {
-        BOXAssertFail(@"Operation was not ready but start was called");
-    }
+            [self performSelector:@selector(executeOperation) onThread:[[self class] globalAPIOperationNetworkThread] withObject:nil waitUntilDone:NO];
```
jlawton, line 320: I'm curious what `globalAPIOperationNetworkThread` gives us. I'm pretty sure it's a thread because it was written in a time before GCD. It's probably used for serialization, that we may or may not need for correctness. If we're sure it's safe to skip it, a comment about why might be useful.

BoxContentSDK/BoxContentSDK/Requests/BOXRequest.h
```diff
@@ -70,6 +70,15 @@ typedef void (^BOXMetadataTemplatesBlock) (NSArray <BOXMetadataTemplate *>*metad
 
 typedef void (^BOXRepresentationInfoBlock)(BOXRepresentation *representation, NSError *error);
 
+@protocol BOXBackgroundRequestProtocol <NSObject>
```
jlawton: Minor: I don't think this needs to conform to `NSObject`

BoxContentSDK/BoxContentSDK/Requests/BOXRequest.h
```diff
@@ -70,6 +70,15 @@ typedef void (^BOXMetadataTemplatesBlock) (NSArray <BOXMetadataTemplate *>*metad
...
+/**
+Caller provided unique ID to execute the request as a NSURLSession background task.
+This is a required value for performing the request in the background.
+*/
+@property (nonatomic, readwrite, copy) NSString *associateID;
```
jlawton: I like this name change!

BoxContentSDK/BoxContentSDK/Clients/BOXContentClient.m
thuyBox: done

BoxContentSDK/BoxContentSDK/Requests/BOXRequest.h
thuyBox: done

BoxContentSDK/BoxContentSDK/Operations/BOXAPIOperation.m
thuyBox: I'm not sure if there are other intention for globalAPIOperationNetworkThread besides it's done before GCD time. The thread entry point does wrap the runloop inside autoreleasepool, so we'll need to make sure we add autoreleasepool around operation execution. Updated

BoxContentSDK/BoxContentSDK/Clients/BOXContentClient.m
```diff
@@ -495,7 +497,17 @@ - (void)prepareRequest:(BOXRequest *)request
...
     }
-    
+
+    // If the request is currently not set to perform immediately,
+    // Check if we should set it to perform immediately or not
+    if (request.shouldPerformRequestImmediately == NO) {
+       request.shouldPerformRequestImmediately =
+            //should background requests perform immediately
+            self.shouldBackgroundRequestsPerformImmediately
+            //and is this request a background request
```
roliva-box: instead of this comment I would rather see a method: `[request isBackgroundRequest];` or [self isBackgroundRequest:request];`

BoxContentSDK/BoxContentSDK/Clients/BOXContentClient.m
```diff
@@ -495,7 +497,17 @@ - (void)prepareRequest:(BOXRequest *)request
         BOXSharedItemRequest *shareItemRequest = (BOXSharedItemRequest *)request;
         shareItemRequest.sharedLinkHeadersHelper = self.sharedLinksHeaderHelper;
     }
-    
+
+    // If the request is currently not set to perform immediately,
+    // Check if we should set it to perform immediately or not
+    if (request.shouldPerformRequestImmediately == NO) {
+       request.shouldPerformRequestImmediately =
+            //should background requests perform immediately
```
roliva-box: nit: this comment seem redundant since it reads same as the line of code

BoxContentSDK/BoxContentSDK/Requests/BOXFileUploadNewVersionRequest.m
```diff
@@ -78,7 +79,7 @@ - (BOXAPIOperation *)createOperation
         }
 
         operation.uploadMultipartCopyFilePath = self.uploadMultipartCopyFilePath;
-        operation.associateId = self.associateId;
+        operation.associateId = self.associateID;
```
roliva-box: There are many many of these rename changes.  I suggest that (if it's possible) you put a prelim PR for the rename so the bug-fix PR is more focused.

BoxContentSDK/BoxContentSDK/Requests/BOXRequest.m
```diff
@@ -65,7 +65,12 @@ - (NSURLRequest *)urlRequest
 - (void)performRequest
 {
     [self.operation.APIRequest setValue:[self userAgent] forHTTPHeaderField:@"User-Agent"];
-    [self.queueManager enqueueOperation:self.operation];
+    if (self.shouldPerformRequestImmediately == YES) {
+        self.operation.shouldStartImmediately = YES;
```
roliva-box, line 69: Interesting we have to tell the operation it can start immediately right before we tell it to start.  Shouldn't the latter be enough?

BoxContentSDK/BoxContentSDK/Clients/BOXContentClient.m
thuyBox: done

BoxContentSDK/BoxContentSDK/Clients/BOXContentClient.m
thuyBox: done

BoxContentSDK/BoxContentSDK/Requests/BOXRequest.m
thuyBox: operation's start does not start the operation immediately, it puts operation execution on a global thread.

BoxContentSDK/BoxContentSDK/Requests/BOXFileUploadNewVersionRequest.m
thuyBox: added https://github.com/box/box-ios-sdk/pull/649

this PR uses the above PR as the first commit.
Please refer to additional changes starting from 2nd commit https://github.com/box/box-ios-sdk/pull/648/commits

BoxContentSDK/BoxContentSDK/Clients/BOXContentClient.h
```diff
@@ -85,6 +85,12 @@ extern NSString *const BOXContentClientBackgroundTempFolder;
  */
 @property (nonatomic, nullable, readwrite, copy) NSDictionary *fetchTokenBlockInfo;
 
+/**
+ *  Used to set background requests' shouldPerformRequestImmediately provided those requests prepared
+ *  using BOXContentClient. E.g.: BOXFileRequest returned by fileInfoRequestWithID:associateID:
```
roliva-box, line 90: This is a bit unclear... `...those request  prepared using`.  
Did you mean `... those request are built in BOXContentClient.`  ?  
I'm not getting the requirement from this comment.  
How can we enforce where the request comes from or how is it prepared?

roliva-box (approved):

BoxContentSDK/BoxContentSDK/Clients/BOXContentClient.h
thuyBox: "Did you mean ... those request are built in BOXContentClient. ?" Yes

In normal cases, users use BOXContentClient categories e.g. BOXContentClient+File's - (BOXFileRequest *)fileInfoRequestWithID:(NSString *)fileID to create BOXFileRequest.

User can choose to instantiate BOXRequest directly e.g. [[BOXFileRequest alloc] initWithFileID:fileID] and configure other necessary properties for it to work similarly to how the above method works. This case should be rare. In this case, the request won't be affected by this property.

PR #649, opened 2019-09-10
IOS-17361: (part 2) add BOXBackgroundRequestProtocol to be conformed …

thuyBox:
…by BOXRequest which supports background mode

Changes affect BOXRequest and its subclasses
- Renaming of `associateId` property to `associateID`
- Making `associateID` a readonly property from the interface
- Adding initializer interfaces which takes `associateID`


boxcla:

roliva-box (approved):

PR #650, opened 2019-09-12
IOS-15408: (part 2) expose query for associateIds of specific backgro…

thuyBox:
…und session id and user id at BOXContentClient interface

### DIFF
--- a/BoxContentSDK/BoxContentSDK/BOXURLSessionManager.m
+++ b/BoxContentSDK/BoxContentSDK/BOXURLSessionManager.m
@@ -613,6 +613,11 @@ static NSString *backgroundSessionIdentifierForMainApp = @"com.box.BOXURLSession
     }
 }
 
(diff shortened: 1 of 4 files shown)


boxcla:

BoxContentSDK/BoxContentSDK/BOXURLSessionManager.m
```diff
@@ -613,6 +613,11 @@ - (void)cancelOnGoingSessionTasksForUserId:(NSString *)userId error:(NSError **)
     }
 }
 
+- (NSArray <NSString *> *)associateIdsOfBackgroundSessionId:(NSString *)backgroundSessionId userId:(NSString *)userId error:(NSError **)error
+{
+    return [self.cacheClient associateIdsOfBackgroundSessionId:backgroundSessionId userId:userId error:error];
```
jlawton, line 618: It's a little odd to me that the cache client is responsible for this. I guess that's always been the case?

jlawton: Ah. This is a special URL session cache.

MashfiqueA (approved):

PR #652, opened 2019-09-23
IOS-17614 - Fix a memory corruption issue in BOXAPIDataOperation

potrebic:
It's likely that this fix, which appears to fix the a big source of
flakiness in running our test suite, will also fix the top crashing
bug in our product (see IOS-17522).

BOXAPIDataOperation instances *were* being "copied", which resulted
in 2 instances retaining a reference to the same underlying NSOutputStream.
This lead to a variety of crashes when one Op closed the Stream
while the other Op was attempting to use it.

This fix is somewhat distasteful as copying an BOXAPIDataOperation instance
now modifies the instance being copying (nil'ing out the output stream
reference).

There is much wrong with BOXAuthenticatedOperation and it's subclasses.
Including fuzzy contracts, lots of duplicated code, contradictory comments & code. 
None of that is addressed here to keep the fix small/focused on an attempt to 
fix the #1 crashed in our product.

### DIFF
--- a/BoxContentSDK/BoxContentSDK/Operations/BOXAPIAuthenticatedOperation.m
+++ b/BoxContentSDK/BoxContentSDK/Operations/BOXAPIAuthenticatedOperation.m
@@ -8,7 +8,6 @@
 
 #import "BOXAPIAuthenticatedOperation.h"
 
-#import "BOXAPIJSONOperation.h"
 #import "BOXLog.h"
 #import "BOXContentSDKErrors.h"
 #import "BOXAbstractSession.h"
@@ -78,7 +77,7 @@
         {
             self.error = [[NSError alloc] initWithDomain:BOXContentSDKErrorDomain code:BOXContentSDKAuthErrorAccessTokenExpiredOperationWillBeClonedAndReenqueued userInfo:nil];
 
-            BOXAPIJSONOperation *operationCopy = [self copy];
+            BOXAPIAuthenticatedOperation *operationCopy = [self copy];
             operationCopy.timesReenqueued = operationCopy.timesReenqueued + 1;
             [self.session.queueManager enqueueOperation:operationCopy];
         }
--- a/BoxContentSDK/BoxContentSDK/Operations/BOXAPIDataOperation.m
+++ b/BoxContentSDK/BoxContentSDK/Operations/BOXAPIDataOperation.m
@@ -362,7 +362,6 @@
 {
     BOXAPIDataOperation *operationCopy = [self copy];
     operationCopy.timesReenqueued++;
-    self.outputStream = nil;
     [self.session.queueManager enqueueOperation:operationCopy];
     [self finish];
 }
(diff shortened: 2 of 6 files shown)


boxcla:

BoxContentSDK/BoxContentSDK/Operations/BOXAPIAuthenticatedOperation.m
```diff
@@ -8,7 +8,6 @@
 
 #import "BOXAPIAuthenticatedOperation.h"
 
-#import "BOXAPIJSONOperation.h"
```
potrebic, line 11: THis was a code smell. An abstract base class including the header of a subclass???

BoxContentSDK/BoxContentSDK/Operations/BOXAPIAuthenticatedOperation.m
```diff
@@ -78,7 +77,7 @@ - (void)processResponse:(NSURLResponse *)response
         {
             self.error = [[NSError alloc] initWithDomain:BOXContentSDKErrorDomain code:BOXContentSDKAuthErrorAccessTokenExpiredOperationWillBeClonedAndReenqueued userInfo:nil];
 
-            BOXAPIJSONOperation *operationCopy = [self copy];
```
potrebic, line 81: ... And then using using that subclass here. `self` isn't always a BOXAPIJSONOperation instance. It can be any subclass that returns `YES` from `canBeReenqueued:`. Yikes.

BoxContentSDK/BoxContentSDK/Operations/BOXAPIDataOperation.m
```diff
@@ -385,8 +384,9 @@ - (id)copyWithZone:(NSZone *)zone
                                                                                           body:bodyCopy
                                                                                    queryParams:queryStringParametersCopy
                                                                                  session:self.session];
+    self.outputStream.delegate = nil;
     operationCopy.outputStream = self.outputStream;
-    operationCopy.outputStream.delegate = nil;
+    self.outputStream = nil;
```
potrebic, line 389: This line is *the* bug fix. Shortly after this code executes `[self finish]` is called. And by nil'ing out the output stream, it's left intact for the copy (thus avoiding a crash). Now, it's still the case that moving the outputStream from one operation to another might be semantically incorrect. If that turns out to be the case then this fix for the flakiness/crashes should be to make BOXDataAPIOperation avoid this `copy` code path in the face of a token-expired error.

BoxContentSDK/BoxContentSDK/Operations/BOXStreamOperation.h
```diff
@@ -14,12 +14,10 @@
 typedef void (^BOXAPIStreamProgressBlock)(NSData *data, long long expectedTotalBytes);
 
 /**
- * BOXAPIDataOperation is a concrete subclass of BOXAPIAuthenticatedOperation.
- * This operation receives binary data from the Box API which may be in the form
- * of downloads or thumbnails.
+ * BOXStreamOperation is a concrete subclass of BOXAPIAuthenticatedOperation.
```
potrebic, line 17: Lots of this header were apparently copy/pasted years ago. Trying to clean it up a bit.

Jeff-Meadows (approved):

jlawton (approved):

MashfiqueA (approved):

PR #654, opened 2019-09-24
IOS-17000 - Refactor to avoid 3 copies of the same 202 handling

potrebic:
3 sibling Operation classes each had it's own chunk of code to deal
with 202 responses. This code was basically identical.

A bit of refactoring in these classes along with
the base classes BOXAPIDataOperation and BOXAPIAuthenticatedOperation
now has this rety logic centralized, right next to the similar logic
for the token expired error.

### DIFF
--- a/BoxContentSDK/BoxContentSDK/Operations/BOXAPIDataOperation.m
+++ b/BoxContentSDK/BoxContentSDK/Operations/BOXAPIDataOperation.m
@@ -12,9 +12,6 @@
 #import "BOXLog.h"
 #import "BOXAbstractSession.h"
 
-#define MAX_REENQUE_DELAY 15
-#define REENQUE_BASE_DELAY 0.2
-
 @interface BOXAPIDataOperation ()
 
 // Buffer data received from the connection in an NSData. Write to the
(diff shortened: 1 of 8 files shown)


boxcla:

Jeff-Meadows (approved):

BoxContentSDK/BoxContentSDK/Operations/BOXAPIDataOperation.m
```diff
@@ -395,9 +367,14 @@ - (id)copyWithZone:(NSZone *)zone
...
+- (BOOL)canBeReenqueuedDueToTokenExpired
 {
     return self.shouldStartImmediately == NO;
 }
 
+- (BOOL)canBeReenqueuedDueTo202NotReady
+{
+    return YES;
+}
+
```
potrebic: Notice the logic/semantics here (pre-existing logic). BOXAPIDataOperation recently started answering the question "Should you be re-enqueued after a token expired error?" with the response "Yes, if shouldStartImmediately ==NO". 

But at the *Same* time BOXAPIDataOperation would always re-enqueue after a 202 response.

The `self.shouldStartImmediately == NO` was recently added (PR 648). I'm concerned that this is an issue. That PR 648 might have missed the logic around 202 handling. This applies to BoxStreamOperation and other 202 handling Operation classes.

jlawton: The reason that re-enqueueing is disabled when `shouldStartImmediately` is that we're asking to skip the queue entirely in that case, so it would be very odd to have the request enqueued anyway.

potrebic: Yes, that's my point. Before Thuy's change this code would re-enqueue logic wasn't disabled at all. After Thuy's change the TokenExpired logic "respected" the `startImmediately` value, but the 202 logic does not.

If we're in agreement over this I can tweak this PR to address that.

jlawton: The ultimate reason to want to avoid the queue is that using the queue introduces a race condition in the operations code. I believe that, in order to avoid effectively the same race in that code, we have to let the operations queue manage retry.

BoxContentSDK/BoxContentSDK/Operations/BOXAPIOperation.m
```diff
@@ -565,46 +565,20 @@ - (void)processResponse:(NSURLResponse *)response
...
-                break;
             case BOXContentSDKAPIErrorTooManyRequests:
-                errorCode = BOXContentSDKAPIErrorTooManyRequests;
-                break;
             case BOXContentSDKAPIErrorInternalServerError:
-                errorCode = BOXContentSDKAPIErrorInternalServerError;
-                break;
             case BOXContentSDKAPIErrorInsufficientStorage:
-                errorCode = BOXContentSDKAPIErrorInsufficientStorage;
+                errorCode = self.HTTPResponse.statusCode;
```
potrebic, line 580: The previous style... its hard to be certain that each case was really doing the right thing... setting the correct value. Easy to intro a copy/paste error with code like that. The new pattern is way better.

I came across this while testing my change (forcing pseudo 202 responses)

jlawton (approved):
LGTM

PR #655, opened 2019-10-03
IOS-17740 - make access/refresh token updates and refreshes atomic

potrebic:
This is to avoid a refresh action updating those values while some
other operation is just reading the values. This is a somewhat
weak/fragile fix, and other access flows could be created in the
future. That's something to be wary of.


boxcla:

Jeff-Meadows (approved):

PR #658, opened 2019-10-18
IOS-15408: (part 6) do not report failures to clean up session task i…

thuyBox:
…f its background session is not yet to be cleaned up


boxcla:

MashfiqueA (approved):

PR #662, opened 2019-10-24
IOS-17850: cache the map of background session ID and its associateID…

thuyBox:
…s under backgroundSessionsIndex

### DIFF
--- a/BoxContentSDK/BoxContentSDK/BOXURLSessionCacheClient.h
+++ b/BoxContentSDK/BoxContentSDK/BOXURLSessionCacheClient.h
@@ -67,6 +67,9 @@
  * 2. To keep track of whose user and its associateId the session task belongs to, we save backgroundSessionId and sessionTaskId
  * as a file name under sub-directory users/$userId/$associateId/info with format $backgroundSessionId-$sessionTaskId
  *
+ * We also record a map of background session ID to its associateIDs at backgroundSessionsIndex/$userID/$backgroundSessionID/$associateID
+ * which provides a quick look up of associateIDs given a background session ID.
+ *
  * 3. Once session tasks complete, their cached info under onGoingSessionTasks/$backgroundSessionId/$sessionTaskId dir
  * will be moved into users/$userId/$associateId/completed dir
  *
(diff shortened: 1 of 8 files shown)


boxcla:

BoxContentSDK/BoxContentSDK/BOXURLSessionCacheClient.m
```diff
@@ -850,30 +890,27 @@ - (NSString *)filePathOfExtensionBackgroundSessionId:(NSString *)backgroundSessi
...
-    for (NSString *associateId in associateIdToBackgroundSessionIdAndSessionTaskIdDict) {
-        BOXURLBackgroundSessionIdAndSessionTaskId *backgroundSessionIdAndSessionTaskId = associateIdToBackgroundSessionIdAndSessionTaskIdDict[associateId];
+    BOOL isDir = NO;
+    NSMutableArray *associateIDs = nil;
 
-        if ([backgroundSessionIdAndSessionTaskId.backgroundSessionId isEqualToString:backgroundSessionId]) {
-            [associateIds addObject:associateId];
+    if ([[NSFileManager defaultManager] fileExistsAtPath:dir isDirectory:&isDir] == YES && isDir == YES) {
+        NSArray *filePaths = [[NSFileManager defaultManager] contentsOfDirectoryAtPath:dir error:error];
+        if (filePaths.count > 0) {
```
jlawton, line 905: I think we should follow the system conventions, and only return `nil` on an error (`filePaths == nil`).

BoxContentSDK/BoxContentSDK/BOXURLSessionCacheClient.h
```diff
@@ -280,6 +283,17 @@
...
 
+/**
+ * Check if a background session is valid, i.e. exists for a userId
+ *
+ * @param userID            Id of user
+ * @param backgroundSessionID       Id of background session
+ *
+ * @return YES if valid, NO if not
+ */
+- (BOOL)isBackgroundSessionValidGivenUserID:(NSString * _Nonnull)userID
```
jlawton, line 294: I think that `Given` is less natural than `For` in these method names.

For this specific name, the idea of `valid` is a little strange. Maybe something like `isBackgroundSessionID:associatedWithUserID:` would be clearer?

jlawton (approved):

BoxContentSDK/BoxContentSDK/BOXURLSessionCacheClient.m
thuyBox: updated in follow-up commit

BoxContentSDK/BoxContentSDK/BOXURLSessionCacheClient.h
thuyBox: updated in follow-up commit

PR #663, opened 2019-10-25
IOS-17850: (part2)

thuyBox:
- Return empty associateIDs/backgroundSessionIDs if their directories are not found, which could be expected after clean up
- Validate backgroundSessionID association with userID only if session has not yet cleaned up


boxcla:

jlawton (approved):

PR #691, opened 2019-12-16
Add ability to view the progress for uploads and downloads

sujaygarlanka:
### Issue Link :link:
- https://jira.inside-box.net/browse/SDK-1181

### Goals :soccer:
- Add ability to view progress for uploads and downloads

- Add observer to progress object on a task

### Testing Details :mag:
- Manual Testing


boxcla:
Verified that @sujaygarlanka has signed the CLA. Thanks for the pull request!

Sources/Network/BoxNetworkAgent.swift
```diff
@@ -143,7 +143,24 @@ public class BoxNetworkAgent: NSObject, NetworkAgentProtocol {
...
+        // Key value observer: Observer attaches to Progress object on task. Every time the Progress object updates,
+        // progressHandler is called with the Progress object passed in. progressHandler calls the request's progress closure and passes in the updated
+        // progress object.
+        var observation: NSKeyValueObservation?
+        let progressHandler: (Progress, NSKeyValueObservedChange<Double>) -> Void = {
+            progress, _ in
+            request.progress(progress)
+            if progress.fractionCompleted >= 1 {
+                observation?.invalidate()
+                observation = nil
```
jlawton: I wouldn't block on this, but I think it'll be worth testing the case that a task is cancelled before completion, to ensure that the observation is cleaned up.

sujaygarlanka: Made a change where I invalidate the observation in callback of the URLSessionTask. So in the case where the task is cancelled, the callback will be called with an error and the observation will be invalidated.

SwiftSDK.playground/contents.xcplayground
```diff
@@ -0,0 +1,4 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
```
PJSimon: Remove playground file via rollback

PJSimon:
Closing PR for now, and we will address issues in comments later.

PR #698, opened 2020-01-14
Change task assignment status from String to Enum

sujaygarlanka:
### Goals :soccer:
- Change `status` field from a String to an Enum in the Task Assignment model

- Used the same enum as the one for the `resolution_state` field because both fields have the same value
- The `status` field is redundant and is not publicly documented, but it is returned by the API, so we must support it

### Testing Details :mag:
- Unit testing

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,16 @@
 Changelog
 =========
 
+## Next Release
+
+__Breaking Changes:__
+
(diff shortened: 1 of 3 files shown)


boxcla:
Verified that @sujaygarlanka has signed the CLA. Thanks for the pull request!

carycheng (approved):

PR #713, opened 2020-04-23
Add ability to cancel uploads and downloads

### LINKED ISSUE #709: How to cancel a file download session
## What did you do?

I use the following code to download a file
client.files.download(fileId: item.id, destinationURL: url, progress: { (progress) in
            progressCallback(Int(progress.completedUnitCount), Int(progress.totalUnitCount))
        }) { (result) in
            switch result {
            case .success(_):
                break
            case .failure(let error):
                break
            }
        }

## What did you expect to happen?

How to cancel this downloading session in the middle, when user pressed a cancel button.

## What happened instead?

if we do not cancel the downloading, the downloading will continue, it is not expected by user.

## BoxSDK Environment

**BoxSDK version: 4.0.0
**Xcode version: 11.4 beta2
**Swift version:5
**Platform(s) running BoxSDK:iOS
**macOS version running Xcode:10.15.3

## Demo Project

n/a

sujaygarlanka:
### Goals :soccer:
- Add ability to cancel uploads and downloads
- Resolves #709

- A `BoxDownloadTask` returned for every download request. This task contains the URLSessionTask that is associated with that request. `BoxDownloadTask` has a cancel method on it that can be called. When called, it cancels the URLSessionTask associated with, resulting in a cancelled request.
- A `BoxUploadTask` returned for every upload request. `BoxUploadTask` has a cancel method on it that can be called.

### Testing Details :mag:
- Added unit tests


Sources/Client/BoxClient.swift
```diff
@@ -437,20 +436,24 @@ extension BoxClient: BoxClientProtocol {
         downloadDestinationURL: URL,
         httpHeaders: BoxHTTPHeaders = [:],
         queryParameters: QueryParameters = [:],
+        task: @escaping (URLSessionTask) -> Void = { _ in },
```
jlawton: It seems like this is not used, and gets shadowed.
```suggestion
```

Sources/Client/BoxClient.swift
```diff
@@ -437,20 +436,24 @@ extension BoxClient: BoxClientProtocol {
...
+    ) -> BoxDownloadTask {
+        let task = BoxDownloadTask()
         send(
             request: BoxRequest(
                 httpMethod: .get,
                 url: url,
                 httpHeaders: httpHeaders,
                 queryParams: queryParameters,
                 downloadDestination: downloadDestinationURL,
+                task: task.getTask(),
```
jlawton:
```suggestion
                task: task.receiveTask(_:),
```

Sources/Modules/FilesModule.swift
```diff
@@ -757,11 +757,11 @@ public class FilesModule {
         fileId: String,
         destinationURL: URL,
         version: String? = nil,
+        //        task: @escaping (URLSessionTask) -> Void = { _ in },
```
jlawton:
```suggestion
```

Sources/Network/BoxDownloadTask.swift
```diff
@@ -0,0 +1,17 @@
...
+//
+
+import Foundation
+
+/// A Box network task returned for a download
+public class BoxDownloadTask: BoxNetworkTask {
+    /// Initializer
+    override init() {
+        super.init()
+    }
```
jlawton:
```suggestion
```

Sources/Network/BoxNetworkTask.swift
```diff
@@ -0,0 +1,47 @@
...
+            if self.cancelBool {
+                sessionTask.cancel()
+            }
+            else {
+                self.task = sessionTask
+            }
+        }
+        // swiftlint:disable:next force_unwrapping
+        return taskMethod!
+    }
```
jlawton:
```suggestion
    func receiveTask(_ sessionTask: URLSessionTask) {
        if self.cancelled {
            sessionTask.cancel()
        }
        else {
            self.task = sessionTask
        }
    }
```

Sources/Network/BoxNetworkTask.swift
```diff
@@ -0,0 +1,47 @@
...
+
+    /// Method to cancel a network task
+    public func cancel() {
+        if let unwrappedTask = task {
+            unwrappedTask.cancel()
+        }
+        else {
+            cancelBool = true
+        }
```
jlawton:
```suggestion
        task?.cancel()
        cancelled = true
```

Sources/Network/BoxNetworkTask.swift
```diff
@@ -0,0 +1,47 @@
...
+public class BoxNetworkTask {
+
+    private var taskMethod: ((URLSessionTask) -> Void)?
+    private var task: URLSessionTask?
+    private var cancelBool: Bool
+    
+    /// Initializer
+    init() {
+        cancelBool = false
+    }
```
jlawton:
```suggestion
    private var cancelled: Bool = false
```

jlawton (dismissed):

Sources/Modules/FilesModule.swift
```diff
@@ -242,26 +242,35 @@ public class FilesModule {
...
+        task.receiveTask(
+            uploadWithPreflightCheck(
+                performCheck: performPreflightCheck,
+                name: name,
+                size: Int64(data.count),
+                parentId: parentId,
+                request: { [weak self] in
+                    guard let self = self else {
+                        return
+                    }
```
jlawton, line 265: Is it expected that uploads don't make progress unless you retain the Files module? (I'm not too bothered by it, it just struck me as potentially unexpected.)

Sources/Network/BoxNetworkTask.swift
```diff
@@ -0,0 +1,32 @@
...
+//  Created by Sujay Garlanka on 3/21/20.
+//  Copyright © 2020 box. All rights reserved.
+//
+
+import Foundation
+
+/// A Box network task request.
+public class BoxNetworkTask {
+
+    var task: URLSessionTask?
```
jlawton: If we only ever use this for cancellation, and we don't expect `task` and `nestedTask` to be alive at the same time (I'm not sure this is true), and it's not part of the public interface, have we considered that we don't need the three task types, and we could get away with:

```swift
// This doesn't need to be public, and probably shouldn't be, as it shares a name with
// a protocol in Combine.
protocol Cancellable {
    func cancel()
}

public class BoxNetworkTask {
    private var task: Cancellable?
    // …
}
```
This would be forwards compatible with deciding to add subclasses in the future. Maybe we already know we'll want them?

sujaygarlanka: We are thinking that BoxDownloadTask can have pause and resuming for downloads. Not sure what we could add to BoxUploadTasks yet, but thought it was better to have a separate class for that in case.

jlawton (dismissed):
This looks like it works to me, and seems like a reasonable evolution of the code.

I think that Combine (which I know we can't use) has a slightly cleaner solution to this sort of thing, but it hinges on the fact that Publishers separate initialization and subscription, and that they are chained, rather than using callbacks. I think moving in that direction would be too large of a change.

This wouldn't be totally trivial to implement. Just to document the idea in case you're interested in an alternative, it would look something like:

```swift
class NetworkTask<T> {
    /// Run a task
    func subscribe(_ completion: @escaping Callback<T>) -> Cancellable
    /// Chain tasks
    func then<U>(_ nextTask: (T) -> NetworkTask<U>) -> NetworkTask<U>
}

func preflightCheck(…) -> NetworkTask<Void>
func upload(…) -> NetworkTask<File>

func uploadWithPreflightCheck(…) -> NetworkTask<File> {
    return preflightCheck(…).then { upload(…) }
}

func uploadWithPreflightCheck(…, _ completion: @escaping Callback<File>) -> Cancellable {
    return uploadWithPreflightCheck(…).subscribe(completion)
}
```

PJSimon (dismissed):

PJSimon (approved):

Issue #697, opened 2020-01-11
Non error logs being flagged as error level


robgtsoftware:
* [ X] I've read, understood, and done my best to follow the [*CONTRIBUTING guidelines](https://github.com/box/box-swift-sdk/blob/master/CONTRIBUTING.md).

## What did you do?

Used the SDK to list and upload file

## What did you expect to happen?

Requests to complete and only log errors if they are actually errors

## What happened instead?

Successful request details are logged to the xcode console and end up in the console.app as error level message. Since these are not error conditions, it would be nice if the logging level was appropriately set so they could be filtered away and not spam the xcode console. Below is an example, but am seeing this for all requests placed.

![image](https://user-images.githubusercontent.com/53913537/72203801-8f154380-3468-11ea-85b4-eda74b24cf74.png)

![image](https://user-images.githubusercontent.com/53913537/72203855-385c3980-3469-11ea-9a1f-a71a83bffa73.png)

## BoxSDK Environment

**BoxSDK version: 3.1
**Xcode version: 11
**Swift version: 5.x
**Platform(s) running BoxSDK: i0S
**macOS version running Xcode:**

## Demo Project

Issue can be seen in OAuth example project contained in this repo

PJSimon:
Thanks for the heads up, @robgtsoftware !  We are tracking this bug (SDK-1320) and will respond back once it's fixed.
Please feel free to submit a PR to resolve it, if you'd like to get it fixed sooner!
Thanks!

PR #714, opened 2020-04-28
Fix logging of API responses

sujaygarlanka:
### Goals :soccer:
- Fix logging of API responses. API responses are being logged as error information rather than debug information

- Responses were being logged as errors when no error occurred and debug information when errors occurred. The logic was flipped to correct this.

### Testing Details :mag:
- Manual testing


sujaygarlanka:
#697

sujaygarlanka:
#697

PJSimon (approved):

PR #716, opened 2020-05-01
Add support for the classification field for Files and Folders

sujaygarlanka:
## Goals :soccer:
- Add support for the classification field for Files and Folders

- Added classification field to `File` and `Folder` models
- Created a new model for this field because this field has fields within it
- Return hex color as UIColor

### Testing Details :mag:
- Added serialization tests

### DIFF
--- a/Sources/Core/BoxJSONDecoder.swift
+++ b/Sources/Core/BoxJSONDecoder.swift
@@ -34,7 +34,6 @@ class BoxJSONDecoder {
         }
 
         guard let object = objectJSON as? T else {
-            // here
             throw BoxCodingError(message: .typeMismatch(key: key))
         }
 
--- a/Sources/Core/BoxJSONDecoder.swift
+++ b/Sources/Core/BoxJSONDecoder.swift
@@ -7,6 +7,7 @@
 //
 
 import Foundation
+import UIKit
 
 // swiftlint:disable:next convenience_type
 class BoxJSONDecoder {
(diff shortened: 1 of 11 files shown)


Sources/Responses/Classification.swift
```diff
@@ -0,0 +1,21 @@
...
+//
+
+import Foundation
+
+/// Details about the classification applied to a Box file or folder
+public class Classification: BoxInnerModel {
+    // MARK: - Properties
+
+    /// The color that is used to display the classification label in a user-interface
+    public let color: String?
```
jlawton: This may be the representation we want in the interface, but I just want to ask if returning a system color object is interesting.

jlawton: Or possibly providing an encoder/decoder. Because it may not be trivial to render this for users.

sujaygarlanka: Thanks for the suggestion. Makes sense to return a UIColor instead. Made a change to do that.

jlawton (dismissed):
LGTM

Sources/Core/BoxJSONDecoder.swift
```diff
@@ -33,6 +34,7 @@ class BoxJSONDecoder {
         }
 
         guard let object = objectJSON as? T else {
+            // here
```
PJSimon:
```suggestion
```

PJSimon (approved):

PR #717, opened 2020-05-12
Add path parameter sanitization

sujaygarlanka:
### Issue Link :link:
- SDK-1363

### Goals :soccer:
- Sanitize path parameters when constructing a URL

- SDK throws an error if path parameters include relative paths such as /../ 

### Testing Details :mag:
- Unit tests

### DIFF
--- a/Sources/Core/Extensions/URLExtension.swift
+++ b/Sources/Core/Extensions/URLExtension.swift
@@ -19,11 +19,8 @@ extension URL {
     }
 
     private static func make(from string: String, relativeTo baseURL: URL) -> URL {
-        let range = NSRange(location: 0, length: string.utf16.count)
-        // Checks that url paths with relative paths like /../ are not sent to the API
-        // swiftlint:disable:next force_try
-        let regex = try! NSRegularExpression(pattern: "\\/\\.+\\/")
-        if regex.firstMatch(in: string, options: [], range: range) != nil {
+        let pattern = "\\/\\.+"
+        if string.range(of: pattern, options: .regularExpression) != nil {
             fatalError("An invalid path parameter exists in \(string). Relative path parameters cannot be passed.")
         }
         guard let url = URL(string: string, relativeTo: baseURL) else {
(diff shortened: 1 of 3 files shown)


carycheng (dismissed):

Sources/Core/Extensions/URLExtension.swift
```diff
@@ -19,6 +19,13 @@ extension URL {
     }
 
     private static func make(from string: String, relativeTo baseURL: URL) -> URL {
+        let range = NSRange(location: 0, length: string.utf16.count)
```
jlawton: Maybe this will always be correct, but I believe the idiomatic way to do this is `NSRange(string.startIndex ..< string.endIndex, in: string)`.

Sources/Core/Extensions/URLExtension.swift
```diff
@@ -19,6 +19,13 @@ extension URL {
...
+        // Checks that url paths with relative paths like /../ are not sent to the API
+        // swiftlint:disable:next force_try
+        let regex = try! NSRegularExpression(pattern: "\\/\\.+\\/")
+        if regex.firstMatch(in: string, options: [], range: range) != nil {
+            fatalError("An invalid path parameter exists in \(string). Relative path parameters cannot be passed.")
+        }
```
jlawton: This might be another formulation to consider.
```suggestion
        for c in string.components(separatedBy: "/") where c == ".." {
            fatalError("An invalid path parameter exists in \(string). Relative path parameters cannot be passed.")
        }
```

jlawton: I think this would catch `..` at the beginning and end of the string, which the current regex wouldn't.

sujaygarlanka: Just trying to filter dots in the path params to avoid people passing in relative paths, so not too worried about the .. in the beginning. Going to update the regex to handle .. at the end. Also, regex makes it possible to search for any number of dots in between slashes.  Also, if we want to make a stricter sanitization in the future we would only have to update the regex.

jlawton: Do we want to search for any number of dots?

PJSimon (approved):

Sources/Core/Extensions/URLExtension.swift
sujaygarlanka: Not necessary to search for any number of dots because it seems unlikely, but if these cases do arise we do throw an error. Also, using regex keeps it consistent across the SDKs.

PR #723, opened 2020-06-03
Fix bug with creating collaboration

sujaygarlanka:
### Goals :soccer:
- Fix bug with creating a collaboration

- `collaborations.create()` and `collaborations.createByUserEmail()` did not add the `accessible_by` parameters to the POST body. I added the parameters to the POST body.

### Testing Details :mag:
- Updated unit tests

### DIFF
--- a/Sources/Modules/CollaborationsModule.swift
+++ b/Sources/Modules/CollaborationsModule.swift
@@ -107,11 +107,11 @@ public class CollaborationsModule {
                 "id": itemId,
                 "type": itemType
             ],
-            "role": role.description,
             "accessible_by": [
                 "id": accessibleBy,
                 "type": accessibleByType.description
-            ]
+            ],
+            "role": role.description
         ]
 
         if let canViewPath = canViewPath {
--- a/Sources/Modules/CollaborationsModule.swift
+++ b/Sources/Modules/CollaborationsModule.swift
@@ -107,15 +107,16 @@ public class CollaborationsModule {
                 "id": itemId,
                 "type": itemType
             ],
-            "role": role.description
+            "role": role.description,
+            "accessible_by": [
+                "id": accessibleBy,
+                "type": accessibleByType.description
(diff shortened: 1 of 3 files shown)


Sources/Modules/CollaborationsModule.swift
```diff
@@ -153,16 +154,17 @@ public class CollaborationsModule {
                 "id": itemId,
                 "type": itemType
             ],
-            "role": role.description
+            "role": role.description,
+            "accessible_by": [
+                "login": login,
+                "type": "user"
+            ]
```
PJSimon:
```suggestion
            "accessible_by": [
                "login": login,
                "type": "user"
            ],
            "role": role.description
```

Sources/Modules/CollaborationsModule.swift
```diff
@@ -107,16 +107,17 @@ public class CollaborationsModule {
                 "id": itemId,
                 "type": itemType
             ],
-            "role": role.description
+            "role": role.description,
+            "accessible_by": [
+                "id": accessibleBy,
+                "type": accessibleByType.description
+            ]
```
PJSimon:
```suggestion
            "accessible_by": [
                "id": accessibleBy,
                "type": accessibleByType.description
            ],
            "role": role.description
```

PJSimon: [nit] What do you think about putting the JSON values in this request in the same order as the response?

(I'm assuming that the test below is saying that `accessible_by` comes before `role`, right?

Sources/Modules/CollaborationsModule.swift
PJSimon: [nit] What do you think about putting the JSON values in this request in the same order as the response?

(I'm assuming that the test below is saying that `accessible_by` comes before `role`, right?

PJSimon (approved):

Issue #727, opened 2020-06-30
SDK Unable to Process Enterprise Events


stevebingsoo:
### Description of the Issue
When using the SDK to get enterprise events of type "SHARE", "UPLOAD", and possibly many other types of events, Box returns a http 200 response but the SDK throws a BoxCodingError

### Steps to Reproduce
Use the SDK to call getEnterpriseEvents and specify event type filters such as

`eventTypes: [.shareEnabled, .upload]`

### Expected Behavior
SDK should process all enterprise events successfully

### Error Message, Including Stack Trace
`
GET https://api.box.com/2.0/events?limit=1&event_type=SHARE&stream_type=admin_logs

Headers: 

    X-Box-UA, Value: agent=box-swift-sdk/1.0; env=iPhone Simulator/10.15.6

    Authorization, Value: Bearer **REDACTED**

Body:

2020-06-30 09:41:19.599226-0500 Box Shared Link Locator[72019:11540840] [NETWORK AGENT] 

Status code: 200: no error

Headers: 

    ◁ Date, Value: Tue, 30 Jun 2020 14:41:19 GMT

    ◁ Connection, Value: keep-alive

    ◁ Content-Encoding, Value: gzip

    ◁ Transfer-Encoding, Value: Identity

    ◁ Strict-Transport-Security, Value: max-age=31536000

    ◁ Content-Type, Value: application/json

    ◁ Cache-Control, Value: no-cache, no-store

    ◁ BOX-REQUEST-ID, Value: **REDACTED**

[...]

PR #730, opened 2020-07-01
Fix bug with getting enterprise events

sujaygarlanka:
### Issue Link :link:
- Deserialization error with enterprise events reported in issue #727.

### Goals :soccer:
- Fix deserialization issue for the `EventItem` model used when getting events.

- The sources of [events](https://developer.box.com/reference/resources/event#param-source) can have keys `item_id`, `item_type` and `item_name`. Our `EventItem` model could not handle these keys.

### Testing Details :mag:
- Added unit test

### DIFF
--- a/Sources/Responses/EventItemType.swift
+++ b/Sources/Responses/EventItemType.swift
@@ -22,6 +22,8 @@ public class EventItem: BoxModel {
         case file(File)
         /// Comment type
         case comment(Comment)
+        /// User type
+        case user(User)
         /// Unknown type
         case unknown
     }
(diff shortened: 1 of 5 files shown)


Sources/Responses/EventItemType.swift
```diff
@@ -35,20 +37,36 @@ public class EventItem: BoxModel {
     /// - Throws: Decoding error.
     public required init(json: [String: Any]) throws {
         rawData = json
+        var updatedJson: [String: Any] = json
 
-        guard let type = json["type"] as? String else {
+        if let entry = updatedJson.removeValue(forKey: "item_id") {
```
carycheng, line 42: Just curious, why are we removing each of the keys and adding them back to updatedJson? Are we changing the keys to map to their respective correct key entries?
```
item_id->id
item_type->type
item_name->name
```

Tests/Modules/EventsModuleSpecs.swift
```diff
@@ -239,11 +239,16 @@ class EventsModuleSpecs: QuickSpec {
...
-                                    expect(event.id).to(equal("f82c3ba03e41f7e8a7608363cc6c0390183c3f83"))
-                                    expect(event.createdBy?.id).to(equal("11111"))
-                                    expect(event.eventType).to(equal(.itemCreated))
-                                    expect(event.sessionId).to(equal("70090280850c8d2a1933c1"))
-
+                                    expect(event.id).to(equal("1a4ade15-b1ff-4cc3-89a8-955e1522557c"))
+                                    expect(event.createdBy?.id).to(equal("55555"))
+                                    expect(event.sessionId).to(beNil())
+                                    switch event.source?.itemValue {
+                                    case let .file(file):
```
carycheng, line 246: item type is probably okay because we used it to deserialize to file but maybe an assertion just to confirm the type?

carycheng (changes requested):
I think it looks good, just a question and nitpick

Sources/Responses/EventItemType.swift
```diff
@@ -22,6 +22,8 @@ public class EventItem: BoxModel {
         case file(File)
         /// Comment type
         case comment(Comment)
+        /// User type
```
carycheng, line 25: I think coveralls is complaining for code coverage, maybe a test case if event type is deserialized into User object?

Sources/Responses/EventItemType.swift
sujaygarlanka: The events endpoint returns objects with `item_id`, `item_type`, and `item_name` as keys. If we use these keys, the deserialization to folders, files, users, etc. fails. https://developer.box.com/reference/resources/event/#param-source

Sources/Responses/EventItemType.swift
sujaygarlanka: added test

PJSimon (approved):

carycheng (approved):

PR #738, opened 2020-10-26
Update tests to run on Swift 5.3 and above

sujaygarlanka:


arjankowski:
same change was already merged #786

PR #740, opened 2020-11-09
Fix clearing of tokenInfo

smowris:
### Issue Link :link:

The revoked token was not properly clearing causing a 401 error. This fix clears the revoked token so the new token can be saved.

### DIFF
--- a/Sources/Sessions/DelegatedAuth/DelegatedAuthSession.swift
+++ b/Sources/Sessions/DelegatedAuth/DelegatedAuthSession.swift
@@ -65,6 +65,7 @@ public class DelegatedAuthSession: SessionProtocol {
             switch result {
             case .success:
                 self.tokenStore.clear { _ in }
+                self.tokenInfo = nil
                 completion(.success(()))
             case let .failure(error):
                 completion(.failure(error))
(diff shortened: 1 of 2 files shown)


jlawton:
The root cause of this bug was the decision to have 2 sources of truth about the token in `DelegatedAuthSession`. There is both a `tokenStore` and a `tokenInfo`. Given that decision, this looks like a good change to me. We don't want revoked tokens hanging around.

Independent of the fix here (which I think we should take), it might be worth re-visiting the initial decision. Are token store reads expected to be too expensive?

PJSimon (dismissed):

PJSimon:
@smowris Looks good!  Can you sign the CLA so I can merge this?  Thanks!

jlawton:
@PJSimon CLA is signed.

PJSimon (dismissed):

PJSimon:
@smowris It looks like this bug exists in other auth classes, too:
- https://github.com/box/box-ios-sdk/blob/master/Sources/Sessions/OAuth2/OAuth2Session.swift#L121
- https://github.com/box/box-ios-sdk/blob/master/Sources/Sessions/SingleToken/SingleTokenSession.swift#L26
Are you in a good position to use the same method you used to find this bug to test that it also exists using the other authentication methods?

smowris:
@PJSimon The other sessions don't support setting tokenInfo to nil. So it isn't necessary to change those other files

PJSimon (approved):

PR #741, opened 2020-11-16
v4.2.0

sujaygarlanka:


docs/docsets/BoxSDK.docset/Contents/Resources/Documents/undocumented.json
```diff
@@ -2,5 +2,5 @@
   "warnings": [
 
   ],
-  "source_directory": "/SDKs/iOS/box-ios-sdk"
+  "source_directory": "/Users/sgarlanka/Documents/SDKs/iOS/box-ios-sdk"
```
PJSimon:
```suggestion
  "source_directory": "/SDKs/iOS/box-ios-sdk"
```

PJSimon: 1) I think it's best to take user-specific info out of this string.
2) Without this change, we don't need to change the file at all.
3) Does it break anything if we manually revert this via this suggested change?

PJSimon (approved):

Issue #689, opened 2019-12-11
Redirect URI between box ios sdk and box Rest Apis


nfz312:
Hi,
I need to way to set OAuth 2.0 Redirect URI valid for "box ios sdk" and ("box android sdk" or "box Rest Apis" )as my app work on multiple platforms,  that new redirect URL changed to:

> boxsdk-<< YOUR CLIENT ID >>://boxsdkoauth2redirect

As you know that Box not allow to set multiple redirect URIs.

PJSimon:
Hi @nfz312,

Sorry for the delay.  My team is still catching up following the holidays!  We'll respond back to you soon!

Thanks!

Tracking: SDK-1252

PR #746, opened 2020-12-08
Add support for OAuth 2 custom callback url

swfree:
### Issue Link :link:

- https://github.com/box/box-ios-sdk/issues/689

### Goals :soccer:

- Add support for OAuth 2 custom callback url

- Users now have the option to pass in a custom `callbackURL` when initializing the SDK: `BoxSDK(clientId: "YOUR CLIENT ID HERE", clientSecret: "YOUR CLIENT SECRET HERE", callbackURL: "YOUR CALLBACK URL HERE")`

### Testing Details :mag:

- Added unit tests & ran TestApp

### DIFF
--- a/Sources/BoxSDK.swift
+++ b/Sources/BoxSDK.swift
@@ -53,7 +53,7 @@ public class BoxSDK {
     ///     The text of that item is your application's client secret.
     ///   - callbackURL: An optional custom callback URL string. The URL to which Box redirects the browser when authentication completes.
     ///     The user's actual interaction with your application begins when Box redirects to this URL.
-    ///     If not specified, default URL is used in a format of `boxsdk-clientId://boxsdkoauth2redirect` with the real value of `clientId`.
+    ///     If not specified, the default URL is used, in the format of `boxsdk-CLIENTID://boxsdkoauth2redirect`, where CLIENTID is replaced with the value of the `clientId` parameter.
     public init(clientId: String, clientSecret: String, callbackURL: String? = nil) {
         // swiftlint:disable:next force_try
         configuration = try! BoxSDKConfiguration(clientId: clientId, clientSecret: clientSecret, callbackURL: callbackURL)
--- a/Sources/BoxSDK.swift
+++ b/Sources/BoxSDK.swift
@@ -51,10 +51,9 @@ public class BoxSDK {
     ///     log in to your Box developer console and click the Edit Application link for the application you're working with.
     ///     In the OAuth 2 Parameters section of the configuration page, find the item labeled "client_secret".
     ///     The text of that item is your application's client secret.
-    ///   - callbackURL: An optional string containing the URI to load after the user completes the OAuth 2 webview authentication flow.
-    ///     This URI allows Box to redirect back to your app with an authentication code.
-    ///     This should be a custom url scheme registered to your app. Do not set this to a custom value,
-    ///     unless you are also setting a custom redirect URI in your app's developer settings.
+    ///   - callbackURL: An optional custom callback URL string. The URL to which Box redirects the browser when authentication completes.
+    ///     The user's actual interaction with your application begins when Box redirects to this URL.
+    ///     If not specified, default URL is used in a format of `boxsdk-clientId://boxsdkoauth2redirect` with the real value of `clientId`.
     public init(clientId: String, clientSecret: String, callbackURL: String? = nil) {
         // swiftlint:disable:next force_try
         configuration = try! BoxSDKConfiguration(clientId: clientId, clientSecret: clientSecret, callbackURL: callbackURL)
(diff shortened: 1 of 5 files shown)


Sources/BoxSDK.swift
```diff
@@ -181,10 +182,12 @@ public class BoxSDK {
         public func getOAuth2Client(
             tokenInfo: TokenInfo? = nil,
             tokenStore: TokenStore? = nil,
+            callbackURL: String? = nil,
```
swfree: I initially picked this approach of adding the callbackURL to `getOAuth2Client`, because the callbackURL only applies to the OAuth 2 auth type. But I'm thinking it may be cleaner to instead add an optional callbackURL parameter to the initializer, so users could instantiate the sdk as `BoxSDK(clientId: "", clientSecret: "", callbackURL: "")`. I'd then store it in BoxSDKConfiguration so it would be available as `configuration.customCallbackURL`.

Thoughts & suggestions much appreciated!

swfree: I ended up pushing a new commit with the implementation I mentioned above. Take a look at the first commit if you'd like to compare the approaches.

sujaygarlanka: I like the approach of the second commit because callback url is an app setting like client id and secret, so it makes sense to set all of them together.

Sources/BoxSDK.swift
```diff
@@ -469,7 +473,7 @@ extension BoxSDK {
     /// - Returns: Standard URL object to be used for authorization in external browser.
     public func makeAuthorizeURL(callbackURL: String? = nil, state: String? = nil) -> URL {
```
sujaygarlanka: Is the callbackURL parameter still required here now? I don't see how it is used in the SDK.

sujaygarlanka:
Do you mind updating the docs here https://github.com/box/box-ios-sdk/blob/master/docs/usage/authentication.md#traditional-3-legged-oauth2? Also, were you able to test the change using the sample apps?

Sources/BoxSDK.swift
swfree: Good point, I'll remove this callbackURL too. I couldn't find where it was used either, my only guess was that it might've been something that partially carried over from the objective-c sdk.

swfree:
@sujaygarlanka will do, I'll update the docs now & add unit tests. And yes I was able to test with the TestApp - it works with & without passing in the callbackURL on initialization.

Sources/Core/BoxSDKConfiguration.swift
```diff
@@ -84,14 +90,16 @@ public struct BoxSDKConfiguration {
...
+        clientAnalyticsInfo: ClientAnalyticsInfo? = nil,
+        callbackURL: String? = nil
     ) throws {
         self.clientId = clientId
         self.clientSecret = clientSecret
 
         self.apiBaseURL = apiBaseURL ?? defaultAPIBaseURL
         self.uploadApiBaseURL = uploadApiBaseURL ?? defaultUploadAPIBaseURL
         self.oauth2AuthorizeURL = oauth2AuthorizeURL ?? defaultOAuth2AuthorizeURL
+        self.callbackURL = callbackURL
```
sujaygarlanka: What do you think about setting the callbackURL here where if nothing is passed, the defaultCallbackURL is used like it is for the urls above.

swfree: I like that idea, it'll make BoxSDK.swift a lot cleaner because then we can just reference `configuration.callbackURL` everywhere instead of having conditionals. I'll try out making that change 👍

Sources/BoxSDK.swift
```diff
@@ -51,9 +51,12 @@ public class BoxSDK {
     ///     log in to your Box developer console and click the Edit Application link for the application you're working with.
     ///     In the OAuth 2 Parameters section of the configuration page, find the item labeled "client_secret".
     ///     The text of that item is your application's client secret.
-    public init(clientId: String, clientSecret: String) {
+    ///   - callbackURL: An optional custom callback URL string. The URL to which Box redirects the browser when authentication completes.
+    ///     The user's actual interaction with your application begins when Box redirects to this URL.
+    ///     If not specified, default URL is used in a format of `boxsdk-clientId://boxsdkoauth2redirect` with the real value of `clientId`.
```
PJSimon:
```suggestion
    ///     If not specified, the default URL is used, in the format of `boxsdk-CLIENTID://boxsdkoauth2redirect`, where CLIENTID is replaced with the value of the `clientId` parameter.
```

sujaygarlanka (dismissed):

PJSimon (changes requested):
Can you also remove Step 5 from the documentation [here](https://github.com/box/box-ios-sdk/blob/master/README.md#oauth2-sample-app) since it's not needed now?

Also, can you also update the JWT Sample App and OAuth Sample App by removing the entry in the Info.plist file referenced above?  (I believe it's in both apps.)  Also check for any references to the callback URL value from the info.plist file in the Sample Apps and remove it.  The idea is to simplify the Sample Apps to not pass an overriding callback URL, since that is more uncommon.

PJSimon:
In the box-ios-preview-sdk, we should also update the sample apps' Info.plist file.

swfree:
@PJSimon as far as I can tell, [Step 5 here](https://github.com/box/box-ios-sdk/blob/master/README.md#oauth2-sample-app) is still a required step. The developer still needs to register a URL scheme in Info.plist for linking back to the app ([see Apple docs here](https://developer.apple.com/documentation/xcode/allowing_apps_and_websites_to_link_to_your_content/defining_a_custom_url_scheme_for_your_app)). This is needed for the OAuth 2 authentication flow for both the default and custom callback URL use cases.

Perhaps to clarify that, I need to add more details to the `authentication.md` update I've made, something like `Open the Info.plist file in the sample app and find the key under URL Types --> Item 0 --> URL Schemes --> Item 0. Set the value for Item 0 to your custom callback URL.`

PJSimon (approved):

sujaygarlanka (approved):

PR #749, opened 2021-01-12
Add zip download support

swfree:
### Goals :soccer:

- Add support for zip download functionality

- Users can now create and download zip files by calling `client.files.downloadZip(name:items:destinationURL:completion:)`.

### Testing Details :mag:

- Added unit tests & tested manually

### DIFF
--- a/Sources/Core/BoxJSONDecoder.swift
+++ b/Sources/Core/BoxJSONDecoder.swift
@@ -164,6 +164,21 @@ class BoxJSONDecoder {
         return color
     }
 
+    static func optionalDecodeZip<T>(json: [[String: Any]]) throws -> T? where T: BoxModel {
+        var modelJSON: [String: Any] = [:]
+        modelJSON["conflict"] = json
+
+        return try decode(json: modelJSON)
+    }
+
+    static func optionalDecodeZipCollection<T>(json: [String: Any], forKey key: String) throws -> [T]? where T: BoxModel {
+        guard let modelCollectionJSON: [[[String: Any]]] = try optionalExtractJSON(json: json, key: key) else {
+            return nil
+        }
+
+        return try modelCollectionJSON.compactMap { try optionalDecodeZip(json: $0) }
+    }
+
     // MARK: - Decodes
 
     static func decode<T>(json: [String: Any]) throws -> T where T: BoxModel {
--- a/Sources/Modules/FilesModule.swift
+++ b/Sources/Modules/FilesModule.swift
@@ -1196,12 +1196,12 @@ public class FilesModule {
     ///   - completion: Returns a standard ZipDownload object or an error
     public func createZip(
         name: String,
-        items: [[String: String]],
+        items: [ZipDownloadItem],
         completion: @escaping Callback<ZipDownload>
     ) {
         var body: [String: Any] = [:]
         body["download_file_name"] = name
-        body["items"] = items
+        body["items"] = items.map { $0.bodyDictWithDefaultKeys }
 
         boxClient.post(
             url: URL.boxAPIEndpoint("/2.0/zip_downloads", configuration: boxClient.configuration),
@@ -1217,10 +1217,10 @@ public class FilesModule {
     ///   - items: Array of files or folders to be part of the created zip
     ///   - destinationURL: A URL for the location on device that we want to store the file once it has been downloaded
     ///   - completion: Returns an empty response or an error
-    public func downloadZip(name: String, items: [[String: String]], destinationURL: URL, completion: @escaping Callback<Void>) {
+    public func downloadZip(name: String, items: [ZipDownloadItem], destinationURL: URL, completion: @escaping Callback<ZipDownloadStatus>) {
         createZip(name: name, items: items) { [weak self] result in
             guard let self = self else {
-                completion(.failure(BoxSDKError(message: .instanceDeallocated("Unable to download Zip - FilesModule deallocated"))))
+                completion(.failure(BoxSDKError(message: .instanceDeallocated("Unable to create Zip - FilesModule deallocated"))))
                 return
             }
 
@@ -1228,13 +1228,36 @@ public class FilesModule {
             case let .success(zip):
                 self.boxClient.download(
                     url: URL.boxAPIEndpoint(zip.downloadUrl.absoluteString, configuration: self.boxClient.configuration),
-                    downloadDestinationURL: destinationURL,
-                    completion: ResponseHandler.default(wrapping: completion)
-                )
+                    downloadDestinationURL: destinationURL
+                ) { [weak self] zipDownloadResult in
+                    guard let self = self else {
+                        completion(.failure(BoxSDKError(message: .instanceDeallocated("Unable to download Zip - FilesModule deallocated"))))
+                        return
+                    }
+
+                    switch zipDownloadResult {
+                    case .success:
+                        self.getZipDownloadStatus(statusURL: zip.statusUrl, completion: completion)
+                    case let .failure(error):
+                        completion(.failure(error))
+                    }
+                }
             case let .failure(error):
                 completion(.failure(error))
             }
         }
+    }
+
+    /// Gets zip download status
+    ///
+    /// - Parameters:
+    ///   - statusURL: The URL to monitor the status of the zip download
+    ///   - completion: Returns a standard ZipDownloadStatus object or an error
+    private func getZipDownloadStatus(statusURL: URL, completion: @escaping Callback<ZipDownloadStatus>) {
+        boxClient.get(
+            url: statusURL,
+            completion: ResponseHandler.default(wrapping: completion)
+        )
     }
 
     // swiftlint:disable:next file_length
--- a/Sources/Modules/FilesModule.swift
+++ b/Sources/Modules/FilesModule.swift
@@ -1216,7 +1216,7 @@ public class FilesModule {
     ///   - name: The name of the zip file to be created
     ///   - items: Array of files or folders to be part of the created zip
     ///   - destinationURL: A URL for the location on device that we want to store the file once it has been downloaded
-    ///   - completion: Returns an empty response or an error
+    ///   - completion: Returns a standard ZipDownloadStatus object or an error
     public func downloadZip(name: String, items: [ZipDownloadItem], destinationURL: URL, completion: @escaping Callback<ZipDownloadStatus>) {
         createZip(name: name, items: items) { [weak self] result in
             guard let self = self else {
@@ -1237,7 +1237,7 @@ public class FilesModule {
 
                     switch zipDownloadResult {
                     case .success:
-                        self.getZipDownloadStatus(statusURL: zip.statusUrl, completion: completion)
+                        self.getZipDownloadStatus(zip: zip, completion: completion)
                     case let .failure(error):
                         completion(.failure(error))
                     }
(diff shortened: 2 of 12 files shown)


Sources/Modules/FilesModule.swift
```diff
@@ -1188,5 +1188,54 @@ public class FilesModule {
...
 
+    /// Creates a zip of multiple files and folders.
+    ///
+    /// - Parameters:
+    ///   - name: The name of the zip file to be created
+    ///   - items: Array of files or folders to be part of the created zip
+    ///   - completion: Returns a standard ZipDownload object or an error
+    public func createZip(
+        name: String,
+        items: [[String: String]],
```
box-internship-take-home, line 1199: May be worth trying to pass in Items with a custom object like shown [here](https://github.com/box/box-ios-sdk/blob/master/docs/usage/metadata.md#create-metadata-template) for metadata.

sujaygarlanka, line 1199: May be worth trying to pass in items with a custom object like shown [here](https://github.com/box/box-ios-sdk/blob/master/docs/usage/metadata.md#create-metadata-template) for metadata. What do you think?

Sources/Modules/FilesModule.swift
```diff
@@ -1188,5 +1188,54 @@ public class FilesModule {
...
+    ///   - completion: Returns an empty response or an error
+    public func downloadZip(name: String, items: [[String: String]], destinationURL: URL, completion: @escaping Callback<Void>) {
+        createZip(name: name, items: items) { [weak self] result in
+            guard let self = self else {
+                completion(.failure(BoxSDKError(message: .instanceDeallocated("Unable to download Zip - FilesModule deallocated"))))
+                return
+            }
+
+            switch result {
+            case let .success(zip):
```
sujaygarlanka, line 1228: I believe there is one more endpoint(`/zip_downloads/{id}/status`) after downloading the zip that gets the status of the download.

Sources/Modules/FilesModule.swift
swfree: Great idea, I'll add that in

Sources/Modules/FilesModule.swift
swfree: Ah that makes sense, so then the final response would be that status object instead of void?

Sources/Responses/ZipDownload.swift
```diff
@@ -0,0 +1,38 @@
...
+    public private(set) var rawData: [String: Any]
+
+    /// The URL to download the Zip file. If entered in a browser, this URL will trigger a download of the Zip file.
+    public let downloadUrl: URL
+    /// The URL to monitor the status of the download.
+    public let statusUrl: URL
+    /// Expiration date of the Zip file download.
+    public let expiresAt: Date?
+    /// Conflicts that occur between items that have the same name.
+    public let nameConflicts: [[String: String]]?
```
sujaygarlanka, line 24: I think nameConflicts should be `[[[String: String]]]`.

sujaygarlanka: I think nameConflicts should be `[[[String: String]]]`.

Sources/Modules/FilesModule.swift
sujaygarlanka: May be worth trying to pass in items with a custom object like shown [here](https://github.com/box/box-ios-sdk/blob/master/docs/usage/metadata.md#create-metadata-template) for metadata. What do you think?

Sources/Modules/FilesModule.swift
sujaygarlanka, line 1228: I believe there is one more endpoint(`/zip_downloads/{id}/status`) after downloading the zip that gets the status of the download.

sujaygarlanka (changes requested):

Sources/Modules/FilesModule.swift
swfree: Good idea, I'll add that in

Sources/Modules/FilesModule.swift
swfree: That makes sense, so the final response will be that status object instead of void?

sujaygarlanka: yup

Sources/Modules/FilesModule.swift
```diff
@@ -1188,5 +1188,77 @@ public class FilesModule {
...
+        createZip(name: name, items: items) { [weak self] result in
+            guard let self = self else {
+                completion(.failure(BoxSDKError(message: .instanceDeallocated("Unable to create Zip - FilesModule deallocated"))))
+                return
+            }
+
+            switch result {
+            case let .success(zip):
+                self.boxClient.download(
+                    url: URL.boxAPIEndpoint(zip.downloadUrl.absoluteString, configuration: self.boxClient.configuration),
```
sujaygarlanka: Is zip.downloadUrl a full endpoint like `https://dl.box.com/....`? If it is, will URL.boxAPIEndpoint append `api.box.com` to the front?

Sources/Responses/ZipDownload.swift
```diff
@@ -0,0 +1,38 @@
...
+    public private(set) var rawData: [String: Any]
+
+    /// The URL to download the Zip file. If entered in a browser, this URL will trigger a download of the Zip file.
+    public let downloadUrl: URL
+    /// The URL to monitor the status of the download.
+    public let statusUrl: URL
+    /// Expiration date of the Zip file download.
+    public let expiresAt: Date?
+    /// Conflicts that occur between items that have the same name.
+    public let nameConflicts: [[[String: String]]]?
```
sujaygarlanka: Sorry, should have mentioned this in the previous review, but forgot. What do you think about making an object for a single name conflict, so name conflicts is a list of name conflict objects? It is something we did in the java SDK [here](https://github.com/box/box-java-sdk/pull/825/files#diff-5ab88aec08abb5a7ef144a623ccfd0d98a63360e3d290fe2a54e791971adfd79). Do you think it could be useful to make nameConflicts more clear or not really?

sujaygarlanka (changes requested):

Sources/Responses/ZipDownload.swift
swfree: Ah good point, I think that'll make nameConflicts clearer too. I'll add that in!

Sources/Modules/FilesModule.swift
swfree: It is a full endpoint, so I thought the same would happen as well. But it downloads correctly with this setup, and instead when I try passing in `url: zip.downloadUrl` I get a 500 🤔  I'll take another look at it though, I might've missed something.

Sources/Modules/FilesModule.swift
```diff
@@ -1188,5 +1188,77 @@ public class FilesModule {
         }
     }
 
+    /// Creates a zip of multiple files and folders.
+    ///
+    /// - Parameters:
+    ///   - name: The name of the zip file to be created
+    ///   - items: Array of files or folders to be part of the created zip
+    ///   - completion: Returns a standard ZipDownload object or an error
+    public func createZip(
```
PJSimon: Can you make this private to match the implementation in [Python](https://github.com/box/box-python-sdk/pull/539/files#diff-82cbbfb4d3d90a1e83e55682688a71badd004edacf3754c4110c83c95e0bc38cR1676) and [.Net](https://github.com/box/box-windows-sdk-v2/pull/700/files#diff-7fb373d03b85bf95bfac5c4d09c2ea5816fc6a69ef62bb97191733815a70881eR1406) SDKs?  The idea is to offer just one method that simply does the download and sends back a "[combined response](https://github.com/box/box-windows-sdk-v2/pull/700/files#diff-7fb373d03b85bf95bfac5c4d09c2ea5816fc6a69ef62bb97191733815a70881eR1360)."

docs/usage/files.md
```diff
@@ -520,3 +520,59 @@ client.files.listRepresentations(
 ```
 
 [get-representations]: https://opensource.box.com/box-ios-sdk/Classes/FilesModule.html#/s:6BoxSDK11FilesModuleC19listRepresentations6fileId18representationHint10completionySS_AA018FileRepresentationJ0OSgys6ResultOySayAA0lM0VGAA0A8SDKErrorCGctF
+
+Create Zip
```
PJSimon: Per comment above, let's make Create Zip private, so this section can be removed.

Tests/Modules/FilesModuleSpecs.swift
```diff
@@ -1875,6 +1875,122 @@ class FilesModuleSpecs: QuickSpec {
             })
         }
 
+        describe("createZip()") {
```
PJSimon: Per comment above, let's make Create Zip private, so this test can be removed.

sujaygarlanka (dismissed):

Sources/Modules/FilesModule.swift
swfree: Will do!

Sources/Core/BoxJSONDecoder.swift
```diff
@@ -164,6 +164,21 @@ class BoxJSONDecoder {
         return color
     }
 
+    static func optionalDecodeZip<T>(json: [[String: Any]]) throws -> T? where T: BoxModel {
```
PJSimon, line 167: This is cool, BTW!  Good job!

docs/usage/files.md
```diff
@@ -28,6 +28,7 @@ file's contents, upload new versions, and perform other common file operations
 - [Set Shared Link](#set-shared-link)
 - [Remove Shared Link](#remove-shared-link)
 - [Get Representations](#get-representations)
+- [Download Zip](#download-zip)
```
PJSimon, line 31: You can roll back this commit.  Running Jazzy Docs during the release will auto generate this.

docs/usage/files.md
```diff
@@ -553,3 +554,5 @@ client.files.downloadZip(name: name, items: items, destinationURL: destinationUR
     print("Zip download status: \(zipDownloadStatus.state)")
 }
 ```
+
+[download-zip]: https://opensource.box.com/box-ios-sdk/Classes/FilesModule.html
```
PJSimon: You can roll back this commit.  Running Jazzy Docs during the release will auto generate this.

Sources/Modules/FilesModule.swift
```diff
@@ -1188,5 +1188,77 @@ public class FilesModule {
...
+                    downloadDestinationURL: destinationURL
+                ) { [weak self] zipDownloadResult in
+                    guard let self = self else {
+                        completion(.failure(BoxSDKError(message: .instanceDeallocated("Unable to download Zip - FilesModule deallocated"))))
+                        return
+                    }
+
+                    switch zipDownloadResult {
+                    case .success:
+                        self.getZipDownloadStatus(statusURL: zip.statusUrl, completion: completion)
```
PJSimon: We need to combine this with zip.nameConflicts, returned from the call to createZip, so that callers to downloadZip get back both the download status AND the name conflicts (since we made createZip private).  You can follow [the example in .Net](https://github.com/box/box-windows-sdk-v2/pull/700/files#diff-7fb373d03b85bf95bfac5c4d09c2ea5816fc6a69ef62bb97191733815a70881eR1360).

docs/usage/files.md
swfree: I left this one in since it doesn't seem like jazzy updates this, but if jazzy makes a duplicate during the release process I can delete it then

Sources/Responses/ZipDownloadStatus.swift
```diff
@@ -24,6 +24,8 @@ public class ZipDownloadStatus: BoxModel {
     public let skippedFolderCount: Int
     /// State of the download for the zip file
     public let state: String
+    /// Conflicts that occur between items that have the same name. This is manually added in the FilesModule.getZipDownloadStatus() method.
```
PJSimon:
```suggestion
    /// Conflicts that occur between items that have the same name. This is always initially nil and updated manually later, via the FilesModule.getZipDownloadStatus() method.
```

PJSimon (dismissed):
I think that if you choose "commit suggestion" then I don't have to re-approve, but feel free to change it if you otherwise want to.

PJSimon (approved):

PR #757, opened 2021-02-18
Add support for filtering groups by name 

sujaygarlanka:
### Goals :soccer:
- Update `listEnterpriseGroups()` to use documented parameter for filtering by name

- Change undocumented `name` key to documented `filter_term` because they both do the same thing

### Testing Details :mag:
- None

### DIFF
--- a/Sources/Modules/GroupsModule.swift
+++ b/Sources/Modules/GroupsModule.swift
@@ -164,12 +164,14 @@ public class GroupsModule {
     ///
     /// - Parameters:
     ///   - name: Only return groups whose name contains the given string (case insensitive).
+    ///   - filterTerm: Limits the results to only groups whose name starts with the search term
     ///   - offset: The offset of the item at which to begin the response.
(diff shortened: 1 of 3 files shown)


swfree (dismissed):
LGTM. Since test coverage fell, maybe add a unit test for `listForEnterprise` to `GroupsModuleSpecs.swift`?

Sources/Modules/GroupsModule.swift
```diff
@@ -164,12 +164,14 @@ public class GroupsModule {
     ///
     /// - Parameters:
     ///   - name: Only return groups whose name contains the given string (case insensitive).
```
PJSimon: If `filterTerm` filters on name, then what does `name` do?  I'm thinking it's actually ignored by the API?  It's not listed in the API docs: https://developer.box.com/reference/get-groups/

PJSimon (approved):

swfree (approved):

PR #759, opened 2021-03-03
Add support for folder lock functionality

swfree:
### Goals :soccer:

- Add support for folder lock functionality

- Users can now create, delete, and retrieve folder locks.

### Testing Details :mag:

- Added unit tests


Sources/Modules/FoldersModule.swift
```diff
@@ -598,4 +598,71 @@ public class FoldersModule {
...
+    ///
+    /// - Parameters:
+    ///   - folderId: The ID of the folder on which to retrieve folder locks.
+    ///   - marker: The position marker at which to begin the response. See [marker-based paging]
+    ///     (https://developer.box.com/guides/api-calls/pagination/marker-based/) for details.
+    ///   - limit: The maximum number of items to return. The default is 100 and the maximum is 1,000.
+    ///   - completion: Returns all of the locks on the folder, or an error if the request is unsuccessful.
+    public func listLocks(
+        folderId: String,
+        marker: String? = nil,
```
sujaygarlanka: Does this endpoint have a marker and limit parameter? It seems that for now it does not have these and only returns a single item.

swfree: That's true, looks like it's only possible to create one folder lock per folder for now (and therefore only return one per folder). I'll take these out!

sujaygarlanka (approved):

Issue #761, opened 2021-03-08
Callback is not called if token has been revoked


hironytic:
- [x] I have checked that the [SDK documentation][sdk-docs] doesn't solve my issue.
- [x] I have checked that the [API documentation][api-docs] doesn't solve my issue.
- [x] I have searched the [Box Developer Forums][dev-forums] and my issue isn't already reported (or if it has been reported, I have attached a link to it, for reference).
- [x] I have searched [Issues in this repo][github-repo] and my issue isn't already reported.

### Description of the Issue

If current token of OAuth2 session has been revoked, a completion callback is not called.

I checked the code in the sdk, and found that it returns before calling the closure here:
https://github.com/box/box-ios-sdk/blob/0cce407733d5c2efac555f4697de2c22c30e3a13/Sources/Client/BoxClient.swift#L177-L185

### Steps to Reproduce
1. On iOS app, authenticate using `BoxSDK.getOAuth2Client(tokenStore:context:completion:)`.
2. Open web browser and visit https://app.box.com/services/browse/my (My Applications)
3. Remove the app from there, this revokes the token as my understanding.
4. Back to iOS app, call any API, for example: `UsersModule.getCurrent(fields:completion:)`.
5. The completion closure never called.

### Expected Behavior

The closure is called with failure result.

### Versions Used
Box SDK: 4.3.0
iOS SDK: 14.4
iOS: 14.4
Xcode: 12.4

[sdk-docs]: ./doc
[api-docs]: https://developer.box.com/docs
[dev-forums]: https://community.box.com/t5/Platform-and-Development-Forum/bd-p/DeveloperForum
[github-repo]: https://github.com/box/box-java-sdk/search?type=Issues

PR #762, opened 2021-03-10
Fix bug for OAuth where the callback is not called if token has been revoked

hironytic:
### Issue Link :link:
#761 

### Goals :soccer:
I believe in that it should always call completion handler even if token has been revoked.

There may be two ways to fix this problem.

1. Remove `return` after calling `handleExpiredToken(completion:)`
2. Call `completion` in a closure that is passed to `handleExpiredToken(completion:)`

I think 2 is preferable. Because `handleExpiredToken(completion:)` is an asynchronous method and the whole process is completed at its completion handler.

### Testing Details :mag:
Added test describes that issue. It occurs only when using `OAuth2Session` as a session (to be exact, only when the session conforms to `ExpiredTokenHandling` protocol.) In the test case, `BoxClient` is created by `getOAuth2Client` to use `OAuth2Session`.

Without fix, the test is timed out because completion handler is not called.
With fix, it passed.


jlawton (dismissed):
LGTM

sujaygarlanka (dismissed):

sujaygarlanka:
Thank you @hironytic for the PR! We really appreciate the contribution.

PR #763, opened 2021-03-16
Add support for copyInstanceOnItemCopy field for metadata templates

swfree:
### Goals :soccer:

- Add support for copyInstanceOnItemCopy field for metadata templates

- Users can now optionally set the `copyInstanceOnItemCopy` field when creating a metadata templae by calling `client.metadata.createTemplate(scope:templateKey:displayName:hidden:copyInstanceOnItemCopy:fields:completion:)`

### Testing Details :mag:

- Added unit test

### DIFF
--- a/Tests/Modules/MetadataModuleSpecs.swift
+++ b/Tests/Modules/MetadataModuleSpecs.swift
@@ -114,7 +114,28 @@ class MetadataModuleSpecs: QuickSpec {
 
             describe("createTemplate()") {
                 it("should make API call to create metadata template and produce file model when API call succeeds") {
-                    stub(condition: isHost("api.box.com") && isPath("/2.0/metadata_templates/schema") && isMethodPOST()) { _ in
+                    stub(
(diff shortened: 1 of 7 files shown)


sujaygarlanka (dismissed):

Tests/Modules/MetadataModuleSpecs.swift
```diff
@@ -187,6 +208,67 @@ class MetadataModuleSpecs: QuickSpec {
                         }
                     }
                 }
+
+                it("should make API call to create metadata template with copyInstanceOnItemCopy and produce file model when API call succeeds") {
```
sujaygarlanka, line 212: Do we need another test or can we just merge this field into the test above?

swfree: I'm thinking of keeping both tests to validate that the behavior works as expected both when copyInstanceOnItemCopy is passed in and when its not, since its just an optional field

sujaygarlanka (approved):

PR #766, opened 2021-03-30
Support stream upload of new file versions

swfree:
### Goals :soccer:

- Add support for stream upload of new file versions. Stream upload of new files already exists. 

### Testing Details :mag:

- Added unit tests

### DIFF
--- a/Sources/Modules/FilesModule.swift
+++ b/Sources/Modules/FilesModule.swift
@@ -474,6 +474,93 @@ public class FilesModule {
         )
     }
 
+
+    /// Upload a new version of an existing file.
+    ///
+    /// - Parameters:
+    ///   - stream: An InputStream of data for the file to be uploaded.
+    ///   - fileSize: The length of the InputStream
+    ///   - forFile: The ID of the file
+    ///   - name: The name of the file. Box supports file names of 255 characters or
+    ///     less. Names containing non-printable ASCII characters, "/" or "\", names with trailing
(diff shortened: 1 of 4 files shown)


jlawton (dismissed):
This looks good to me. I'd love to see us also support the `if-match` and the very badly named `content-md5` headers. Particularly the former I think is important for correctness of some applications.

swfree:
Thanks @jlawton! I added support for `If-Match` to both `streamUploadVersion` & `uploadVersion`. I'll file a ticket separately for adding `content-md5` headers throughout since that'll include some extra changes to support it

Sources/Modules/FilesModule.swift
```diff
@@ -314,6 +314,7 @@ public class FilesModule {
     ///   - name: The name of the file. Box supports file names of 255 characters or
     ///     less. Names containing non-printable ASCII characters, "/" or "\", names with trailing
     ///     spaces, and the special names “.” and “..” are also not allowed.
+    ///   - ifMatch: The `etag` of the file version. Ensures this item hasn't recently changed before making changes.
```
jlawton: Just riffing here, to make the usage clear. Not insisting on a change.
```suggestion
    ///   - ifMatch: The `etag` of the old file version. Ensures that the file hasn't been updated since getting the etag, which could indicate conflicting edits.
```

jlawton (dismissed):

jlawton (approved):

sujaygarlanka (approved):

PR #772, opened 2021-04-21
Update Travis CI config to shorten build time

swfree:


arjankowski:
This change is no longer needed because in #793  we migrated to GitHub Actions

Issue #775, opened 2021-05-17
Warning: Enum case 'success' has one associated value that is a tuple of 2 elements


rayfix:
This line generates a warning

https://github.com/box/box-ios-sdk/blob/3dc34a7d0ad2741b3792d0e81db29527eb5489e3/Sources/Sessions/DelegatedAuth/DelegatedAuthSession.swift#L106

PR #783, opened 2021-07-08
Fix Swift warning for deprecated tuple destructuring.

jlawton:
### Issue Link #775
Swift compiler warning for deprecated syntax, which allowed for destructuring tuple elements in an enum associated value without writing the tuple parentheses.

### Testing Details :mag:
This change is purely syntactic.


mhagmajer (approved):
Thanks!

PR #785, opened 2021-08-23
Replace insensitive event types

mhagmajer:


Sources/Responses/EventType.swift
```diff
@@ -68,9 +68,9 @@ public enum EventType: BoxEnum {
     /// 2 factor authentication enabled by user.
     case twoFactorEnabled
     /// Free user accepts invitation to become a managed user.
-    case masterInviteAccepted
```
mwwoda, line 71: Is it not a breaking change if we are removing this case?

Sources/Responses/EventType.swift
```diff
@@ -68,9 +68,9 @@ public enum EventType: BoxEnum {
...
+    case adminInviteAccepted
     /// Free user rejects invitation to become a managed user.
-    case masterInviteRejected
```
mwwoda, line 73: Is it not a breaking change if we are removing this case?

Sources/Responses/EventType.swift
jlawton: It does seem breaking, but it's a rename, not a deletion, so the change is simple to accommodate. What's the implication of it being a breaking change? That we shouldn't make it? That we need to bump a major version?

mhagmajer:
Thanks for the review @mwwoda. I've restored the removed cases.

mwwoda (dismissed):
I'm approving it, but please check if it is possible to deprecate those old master* cases. I think we should discourage clients from using the old values.

PR #786, opened 2021-09-14
Update tests to be able to run on XCode 12.5

arjankowski:
According to XCode 12.5 release notes a library libswiftXCTest.dylib (used by Nimble v8.0.0) is no longer included
and was replaced by libXCTestSwiftSupport.dylib. 

To make tests running on XCode 12.5 Nimble was updated to v9.0.1 which resulted in updating the API used in tests.


BoxSDK.xcodeproj/xcshareddata/xcschemes/TestApp.xcscheme
```diff
@@ -0,0 +1,78 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
PJSimon: Just making sure: you want to commit this file?  Is it for the Test App you created recently or for the existing Test App in the project?

.idea/box-ios-sdk.iml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
PJSimon: Are these .idea files from IntelliJ?  Do we want to commit them for others to use?

PJSimon (dismissed):
Left some comments (questions), but I'll go ahead and approve now, so as not to block you from merging, if all is good.

.idea/box-ios-sdk.iml
arjankowski: I don't. I've already removed them in the next commit.

BoxSDK.xcodeproj/xcshareddata/xcschemes/TestApp.xcscheme
arjankowski: This file is redundant. I've already removed it in the next commit.

PJSimon (approved):

PR #787, opened 2021-09-24
Increase test coverage

arjankowski:


Sources/Core/ThreadSafeQueue.swift
```diff
@@ -30,13 +30,13 @@ class ThreadSafeQueue<T> {
...
+            }
+
             self.completionQueue.async {
-                if !self.list.isEmpty {
-                    completion(self.list.removeFirst())
-                }
-                else {
-                    completion(nil)
-                }
+                completion(firstItem)
```
PJSimon, line 39: Can you explain this change, please?

Sources/Modules/SearchModule.swift
```diff
@@ -23,9 +23,9 @@ public enum SearchScope: BoxEnum {
     /// - Parameter value: String representing search scope
     public init(_ value: String) {
         switch value {
-        case "user":
+        case "user_content":
             self = .user
-        case "enterprise":
+        case "enterprise_content":
```
PJSimon, line 28: Nice catch!  :^)

Sources/Responses/EventContainer.swift
```diff
@@ -9,6 +9,7 @@
 import Foundation
 
 /// Container for Events
+@available(*, deprecated, message: "This class is obsoleted and will be removed in v5.0. Please use EntryContainer<Element> instead.")
```
PJSimon: Can you briefly explain this one, please?

Sources/Responses/EventType.swift
```diff
@@ -452,6 +452,36 @@ public enum EventType: BoxEnum {
...
+        case "DEVICE_TRUST_CHECK_FAILED":
+            self = .deviceTrustCheckFailed
+        case "TERMS_OF_SERVICE_ACCEPT":
+            self = .termsOfServiceAccepted
+        case "TERMS_OF_SERVICE_REJECT":
+            self = .termsOfServiceRejected
+        case "DOWNLOAD":
+            self = .downloaded
+        case "DELETE_USER":
+            self = .deletedUser
```
PJSimon, line 484: Is this related to fixing existing tests?  If not, maybe we should break this out into another PR for adding missing EventTypes?

Sources/Responses/Token.swift
```diff
@@ -35,6 +35,6 @@ public class Token: BoxModel {
         accessToken = try BoxJSONDecoder.optionalDecode(json: json, forKey: "access_token")
         expiresIn = try BoxJSONDecoder.optionalDecode(json: json, forKey: "expires_in")
         tokenType = try BoxJSONDecoder.optionalDecode(json: json, forKey: "token_type")
-        restrictedTo = try BoxJSONDecoder.optionalDecodeCollection(json: json, forKey: "scope")
+        restrictedTo = try BoxJSONDecoder.optionalDecodeCollection(json: json, forKey: "restricted_to")
```
PJSimon, line 38: Great catch!

Sources/Responses/Webhook.swift
```diff
@@ -202,9 +202,9 @@ public class Webhook: BoxModel {
             case .taskAssignmentUpdated:
                 return "TASK_ASSIGNMENT.UPDATED"
             case .metadataInstanceCreated:
-                return "METEDATA_INSTANCE.CREATED"
+                return "METADATA_INSTANCE.CREATED"
             case .metadataInstanceUpdated:
-                return "METEDATA_INSTANCE.UPDATED"
+                return "METADATA_INSTANCE.UPDATED"
```
PJSimon, line 207: Just double-checking: not a breaking change?

PJSimon (dismissed):

Sources/Core/ThreadSafeQueue.swift
arjankowski: To ensure that ThreadSafeQueue is really safe, all operations like `enqueue`, `dequeue`, `isEmpty` and `peek` should be executed on the same serial queue which is in our case `utilityQueue`. 
Otherwise a race condition can occurs which happened in our case. 

In `ThreadSafeQueue` class `enqueue` and `isEmpty` methods are accessing our resource which is a `private var list = [T]()` property in `utilityQueue`, but `dequeue` and `peek`
are accessing to `list` property on different queue `completionQueue`. So it is possible that two separeted threads accessing our `list` property at the same time hence race condition.

To fix that I changed that all operations on `list` property are executed on `utilityQueue` and `completionQueue` only gets the result.

Sources/Modules/SearchModule.swift
arjankowski: thx!

Sources/Responses/EventContainer.swift
arjankowski: `EventContainer` and `TrackingCode` classes are not used in our code and should be removed from repository.
The @available attribute used here will inform the consumer of this class by displaying a warning message in his code with the message placed in `message` field.

Sources/Responses/EventType.swift
arjankowski: I think that it should be placed it this commit. 
These values that we see here acutally were added a long time ago, but thanks to tests I found out that they are actually not working.
This is because there were misssing a mapping json(string) values into enum values, so I've just added it here.

Sources/Responses/Token.swift
arjankowski: thx!

Sources/Responses/Webhook.swift
arjankowski: Unfortunately, but because of this typo, the backend was receiving invalid values that it can not recognize

It's not a breaking change because it doesn't affect the existing API in any way.

Sources/Core/ThreadSafeQueue.swift
PJSimon: Cool - I see that now, on line 28!  Thanks!

PJSimon (approved):

PR #790, opened 2021-10-04
Remove deprecated insensitive language field in BoxClient

arjankowski:


Sources/Client/BoxClient.swift
```diff
@@ -23,8 +23,6 @@ public class BoxClient {
     public private(set) lazy var collections = CollectionsModule(boxClient: self)
     /// Provides collaborations functionality.
     public private(set) lazy var collaborations = CollaborationsModule(boxClient: self)
-    /// Provides collaborations whitelist functionality (deprecated)
-    public private(set) lazy var collaborationWhiteList = CollaborationAllowlistModule(boxClient: self)
```
mhagmajer, line 27: Won't this be a breaking change if we remove that? We originally kept to make sure that clients still work

arjankowski: Yes, this is a breaking change. And I want to merge this before our release to don't have to do another release with breaking changes soon after this. 

Main reason for this is that I planned to change swiftlint's warnings to errors. If we don't remove this field, then this will not be possible because `collaborationWhiteList` field will generate an error.

What this more, currently users  will not be informed that this field `collaborationWhiteList` is deprecated so they will still using this filed.

mhagmajer:
I guess this is okay if we want to go ahead with the braking change. Please update the Changelog

mhagmajer (dismissed):

mhagmajer (approved):

Sources/Client/BoxClient.swift
PJSimon: The change to deprecate this was merged in July, but there hasn't been a release of the SDK since April.  That means we haven't released a version of the SDK with the deprecation.  I think we need to release the deprecation in the next release and let that sit with the community to give them time to adjust their code, before we remove it in a later release.

PJSimon (approved):

PR #791, opened 2021-10-08
Make swiftlint produce errors instead of warnings

arjankowski:
Till now swiftilnt was generating warnings that someone might have overlooked.

To make lint messages more visible, which will also improve code quality, now they will be shown as errors.
As a result of this, each PR pushed to repository will have to follow the rules set in .swiftlint.yml.

### DIFF
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -42,7 +42,7 @@ opt_in_rules:
   # - function_default_parameter_at_end
   - identical_operands
   - implicitly_unwrapped_optional
-  - implicit_return
+  #- implicit_return
   - last_where
   - legacy_multiple
   - legacy_random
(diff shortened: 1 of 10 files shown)


.swiftlint.yml
```diff
@@ -42,7 +42,7 @@ opt_in_rules:
   # - function_default_parameter_at_end
   - identical_operands
   - implicitly_unwrapped_optional
-  - implicit_return
+  #- implicit_return
```
antusus, line 45: Why this is disabled?

arjankowski: The implicit return reduces the readability of the code, in particular when people who are just learning swift work with the SDK.

Additionally we currently have 34 places in the code where this rule is not followed, where I believe it is the right choice.

antusus (approved):

PR #793, opened 2021-10-15
ci: integrate build process with GitHub Actions

arjankowski:
Replace Travis CI with GitHub Actions.


.github/workflows/pull-request-to-main.yml
```diff
@@ -0,0 +1,37 @@
...
+      - name: Extract branch name
+        shell: bash
+        run: echo "##[set-output name=branch;]$(echo ${GITHUB_REF#refs/heads/})"
+        id: get_branch
+      - run: bundle exec slather
+        env:
+          GIT_BRANCH: ${{ steps.get_branch.outputs.branch }}
+          CI_PULL_REQUEST: ${{ github.event.number }}
+          COVERAGE_ACCESS_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      - name: Lint
```
antusus: Maybe we should run linter before running unit tests. Compile and linter should be quick and it should fail immediately when files do not follow our coding rules. Is it possible to separate that?

arjankowski: In this configuration you see steps almost the same as it was in travis config. Previously lint was executed after unit test so i put here same. My idea was to first match configuration and later improve them to our needs. 

But as you have suggested this already, I will add this change to his PR. 
Thanks!

antusus (approved):

mhagmajer (approved):
👍 👍 👍

PR #803, opened 2021-12-06
chore: Add integration tests for folder module

arjankowski:


IntegrationTests/Configuration.swift
```diff
@@ -0,0 +1,13 @@
...
+//  BoxSDKIntegrationTests-iOS
+//
+//  Created by Artur Jankowski on 01/12/2021.
+//  Copyright © 2021 box. All rights reserved.
+//
+
+import Foundation
+
+enum Configuration {
+    static var developerToken: String = "" // Please put your DEVELOPER TOKEN here
```
mwwoda, line 12: So the developer token must be placed here every time the tests are ran? Dev tokens are valid only for 60 minutes. Are we considering support for obtaining access tokens based on client credentials?

IntegrationTests/BaseIntegrationSpecs.swift
```diff
@@ -0,0 +1,54 @@
...
+            }
+        }
+    }
+
+    func deleteFolder(_ folder: Folder?, recursive: Bool = false) {
+        guard let folder = folder else {
+            return
+        }
+
+        waitUntil(timeout: .seconds(10)) { done in
```
mwwoda: It should use `Constants.defaultTimeout` like createFolder I guess

Sources/Responses/EntryContainer.swift
```diff
@@ -65,22 +65,23 @@ public class EntryContainer<T: BoxModel>: BoxModel {
...
+        if let limitValue = json["limit"] {
+            limit = Int(String(describing: limitValue))
+        }
+        else {
+            limit = nil
+        }
+
         // The `nextStreamPosition` field is inconsistent in the API
         // Sometimes it could be an Int and another time it could be a String.
         // So as a work around we should check both of these types and try to decode original value.
```
mwwoda, line 82: So this comment is no longer valid?

IntegrationTests/FolderModuleIntegrationSpecs.swift
```diff
@@ -0,0 +1,559 @@
...
+                            done()
+                        }
+                    }
+                }
+            }
+
+            context("list") {
+                var folder: Folder?
+
+                beforeEach {
```
mwwoda, line 92: Are `beforeEach` and `afterEach` called before every `it` statement in this context? (I'm not that familiar with Quick) Was that your intention to recreate folder structure after line 147? Other tests are contained in a single `it`.

IntegrationTests/Configuration.swift
arjankowski: Yes currently is should be placed before every tests run. 
But this solution will be evolving w to OAuth access_token by using secret and client_id.

IntegrationTests/BaseIntegrationSpecs.swift
arjankowski: Done

Sources/Responses/EntryContainer.swift
arjankowski: It's still valid but I replaced code with simple one

IntegrationTests/FolderModuleIntegrationSpecs.swift
arjankowski: `beforeEach` and `afterEach` are called before every it.
But these tests are focused on different approaches to fetching items so I put them in separate `it` blocks.

mwwoda (approved):

PR #820, opened 2022-03-11
docs: Information about supported versions

mgrytsai:
Added details about version support policy.


README.md
```diff
@@ -222,69 +215,24 @@ func obtainJWTTokenFromExternalSources() -> DelegatedAuthClosure {
...
+### Version schedule
 
-> #### Support for RC releases
-> A RC release
-> - is Considered [Active](#active) when released
-> - transitions to [End-of-life](#end-of-life) when the next release becomes Active
+| Version | Current Patch/Minor | Supported Environments                                  | State     | First Release | EOL/Terminated |
+|---------|---------------------|---------------------------------------------------------|-----------|---------------|----------------|
+| 5       | 5.1.0               | iOS 11.0+ / Mac OS X 10.13+ / tvOS 11.0+ / watchOS 4.0+ | Supported | 28 Oct 2021   | TBD            |
+| 4       | 4.4.0               |                                                         | EOL       | 13 Feb 2020   | 30 July 2022   |
```
arjankowski: Is EOL/Terminated "30 July 2022" for version 4.4.0 is OK? Because now we have here ambiguity with it's state which is EOL.

mgrytsai: I am thinking about stating official EOL date in 3 months as we previously haven't communicated clearly EOL and version support. So saying that 4..4.0 is EOL on July 30th 2022 and until then it is not under active development/support.

Do you think we can communicate date when this policy is effective differently?

arjankowski (approved):

PR #841, opened 2022-04-27
fix: Respect Content-Type coming from custom headers

potrebic:
For json bodies, if a Content-Type is set in the custom headers then
don't overwrite that with `application/json`.


lukaszsocha2 (approved):
Looks good

PR #872, opened 2022-08-30
feat: Added support for Modifiable Retention Policies

antusus:
Added support of Modifiable Retention Policies: https://developer.box.com/changelog/#2022-07-23-retention-api-updates


arjankowski (approved):
👍

Issue #892, opened 2023-03-31
Empty string "next_marker" should be treated as the end of paging


potrebic:
- [ x ] I have checked that the [SDK documentation][sdk-docs] doesn't solve my issue.
- [ x ] I have checked that the [API documentation][api-docs] doesn't solve my issue.
- [ x] I have searched the [Box Developer Forums][dev-forums] and my issue isn't already reported (or if it has been reported, I have attached a link to it, for reference).
- [ x ] I have searched [Issues in this repo][github-repo] and my issue isn't already reported.

### Description of the Issue
Its appears that the "recents" Box V2 API has started returning "" for next_marker to indicate the last page. I suspect
this to be a recent change, our code using the Box IOS SDK and the recents API hasn't changed, just the PagingIterator
just started misbehaving, basically going into an infinite loop of paging due to the empty-string.

The Box paging documentation does state that an empty-string is a valid "last page" indicator.
ref: https://developer.box.com/guides/api-calls/pagination/marker-based/#collections

### Expected Behavior
When the PagingIterator hits an "" next_marker paging should terminate.

### Versions Used
iOS SDK: 5.4.0

PR #893, opened 2023-03-31
fix: An empty string `nextMarker` should be treated as end-of-paging.

potrebic:
The Box V2 API docs are inconsistent on this. However it is stated on the paging docs(1):
> If this value is null or an empty string there are no more results to fetch.

So with this change an empty string next_marker will end paging.

(1): https://developer.box.com/guides/api-calls/pagination/marker-based/#collections

### Issue Link :link:
https://github.com/box/box-ios-sdk/issues/892


jlawton (approved):

arjankowski (approved):
👍

PR #899, opened 2023-06-19
chore: Make the contract of AuthModuleDispatcher clearer.

jlawton:
### Goals :soccer:

This shouldn't change any public contracts of the `AuthModule`.

Change `AuthModuleDispatcher`, to remove an unused counter which is subject to a race. Guarantee that, as long as auth actions all call their completion handler, all actions will be processed. Neither of these are likely issues in practice, but I believe this contract would be expected by consumers.

Fix `OAuth2Session` to ensure that it calls the auth action completion block, even in the case it is deallocated. I'm not sure this can happen or cause issues in practice, but this is unclear from purely local reasoning.

Removing `ThreadSafeQueue`, as it is only used in this one case, it introduces unnecessary thread hops, and has features which are unused and unlikely to be efficiently usable.


arjankowski (approved):
thanks for these fixes!

PR #900, opened 2023-06-19
fix: Make code compatible with macOS.

jlawton:
### Goals :soccer:

Allow for the SDK code to be compiled targeting macOS.

This is mostly avoiding the dependency on UIKit. In order to avoid a change to the existing interface on iOS, this conditionally selects which color implementation to use, depending on the target platform.

This is only the code change, and doesn't include the build setting changes needed to target macOS.


arjankowski (approved):
👍

PR #901, opened 2023-06-19
fix: Remove use of deprecated string scanner API.

jlawton:
### Goals :soccer:

Remove use of deprecated `scanHexInt32`.

While making the change to `dataFromHexString` it became clear that this method wasn't actually needed. It's there in order to take a hexadecimal string encoding of a hash and re-encode it to a base64, but simply skipping the production of the hex from the data in the first place makes the method unnecessary.

### Testing Details :mag:

Tested `sha1()` method manually.


arjankowski (approved):
👍

PR #902, opened 2023-06-20
fix: Identify iPhone Simulator on Apple Silicon

jlawton:
### Goals :soccer:

Fix UIDevice.modelName test, when run on iOS Simulator on an Apple Silicon Mac.


arjankowski (approved):
👍

PR #919, opened 2024-03-07
feat: Add support for additional fields in `SignRequest`

arjankowski:


IntegrationTests/SignRequestsModuleIntegrationSpecs.swift
```diff
@@ -73,10 +91,27 @@ class SignRequestsModuleIntegrationSpecs: QuickSpec {
...
                             case let .success(signRequestResult):
                                 signRequest = signRequestResult
+                                expect(signRequest?.name).to(equal(signParameters.name))
+                                expect(signRequest?.isPhoneVerificationRequiredToView).to(equal(signParameters.isPhoneVerificationRequiredToView))
+                                expect(signRequest?.signatureColor).to(equal(signParameters.signatureColor))
                                 expect(signRequest?.redirectUrl).to(equal(signParameters.redirectUrl))
                                 expect(signRequest?.declinedRedirectUrl).to(equal(signParameters.declinedRedirectUrl))
                                 expect(signRequest?.parentFolder.id).to(equal(rootFolder.id))
                                 expect(signRequest?.signFiles?.files?.count).to(equal(2))
+                                // first signer is the sender with role final_copy_reader, second and third is the recipient with role signer
```
mwwoda, line 101: Is it also possible to assert roles in this test?

Sources/Responses/SignRequestSigner.swift
```diff
@@ -77,6 +77,22 @@ public class SignRequestSigner: BoxModel {
     public let redirectUrl: String?
     /// The URL that a signer will be redirect to after declining to sign a document.
     public let declinedRedirectUrl: String?
+    /// This URL is specifically designed for  signing documents within an HTML `iframe` tag.
```
mwwoda: Extra space between "for" and "signing". I saw it's like this in the spec so maybe we shouldn't care about it.

mwwoda (dismissed):
Approved! Please also check comments

mwwoda:
Also consider another name for this PR as it's adding much more than a single field

mwwoda (approved):
🚀

PR #936, opened 2025-06-11
feat!: Bump minimum supported platform versions

arjankowski:


mwwoda (approved):
Do we consider it as a breaking change? Should it be `feat!` instead?

PR #1055, opened 2025-09-10
chore: Make SDK compiles on Linux

arjankowski:
Adjusted the code so it compiles and runs on Linux. The changes included:
* Adding compiler directives to conditionally exclude or replace platform-specific code.
* Conditionally importing FoundationNetworking in files that use networking APIs.
* Providing a custom implementation of PlatformColor when it’s not available in the library.
* Creating a SHA1Calculator protocol with two implementations:
    * One using the CommonCrypto library (for iOS).
    * One custom implementation for Linux (sourced from the [box-swift-sdk-gen repo](https://github.com/box/box-swift-sdk-gen/blob/main/Sources/Internal/SHA1.swift)). This class is then used in DataExtension.swift to calculate SHA1 hashes.
* Updating ConsoleLogDestination and FileLogDestination to work on platforms where os.log is unavailable.

This PR is going to be merged into feature branch, so GHA didn't run.
But to ensure that the changes here works, I've created another PR to main -> [here](https://github.com/box/box-ios-sdk/pull/1054) <- just to proof that the it works

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -7,14 +7,14 @@ on:
     branches:
       - main
 jobs:
-  build-and-test:
+  ios-build-and-test:
     runs-on: macos-13
     strategy:
       matrix:
         destination: ['platform=iOS Simulator,OS=16.2,name=iPhone 14']
     steps:
       - name: Git - Checkout
-        uses: actions/checkout@v2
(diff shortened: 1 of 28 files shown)


Pask423 (approved):

Pask423:
Reviewed online looks good despite it complexity

PR #1062, opened 2025-09-12
chore: Add cron job to run tests

arjankowski:
Added cron job that runs daily on weekdays at around 00:20 CET (full hours could cause delays). 
The job runs all the integration tests on sdk-gen branch

### DIFF
--- a/.github/workflows/build_and_test_daily.yml
+++ b/.github/workflows/build_and_test_daily.yml
@@ -1,0 +1,149 @@
+name: build-and-test
+on:
+  schedule:
+    - cron: "20 22 * * 1-5"
+jobs:
+  ios:
+    name: ${{ matrix.name }}
+    runs-on: ${{ matrix.runsOn }}
+    strategy:
+      matrix:
+        include:
+          - destination: 'platform=iOS Simulator,OS=18.5,name=iPhone 16 Pro'
+            name: 'xcodebuild (iOS 18.5)'
+            xcode: 'Xcode_16.4'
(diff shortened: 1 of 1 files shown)


.github/workflows/build_and_test_daily.yml
```diff
@@ -0,0 +1,149 @@
+name: build-and-test
+on:
+  schedule:
+    - cron: "20 22 * * 1-5"
```
mwwoda, line 4: I'm still thinking what time we should give between the cron runs between the sdks to avoid rate limit issues. I see that current runs can take up to 15 minutes - https://github.com/box/box-windows-sdk-v2/actions/runs/17662802724. I think this core currently assumes 15 minutes break between the other one (00:05 run, 00:20 run etc). Do you think that's enough, or should we push it further to like 20-30 minutes?. Or maybe we shouldn't even care and just migrate to separate accounts per SDK at some point. Love to hear your thoughts on this

arjankowski: I believe that we can start just with an one hour shift between running tests in each SDK. 
For instance, here I'm starting them  at 0:20, then other, 1:20, another one 2:20 and so on.
That should be enough, but if not we can adjust it based on the results

mwwoda: Oh okay, that makes sense I guess. For node maybe we could do 30 minutes slots between browser - server

mwwoda (approved):

PR #1561, opened 2026-06-12
test: Fix flaky tests

arjankowski:
The tests now use a fake `NetworkAgentProtocol`, so they still verify the generated `BoxRequest` and response parsing, but no longer depend on async HTTP stubbing behavior that was causing timeout failures on CI


Pask423 (approved):
👍

keanuharrell (approved):

